PR #147, opened 2023-11-22
Fix font reference, fix breadcrumbs, minor layout tweak

eckmo:
- Fix reference to font loaded via npm
- Fix breadcrumbs and how they render
- Override the status message template to add status messages to the grid

To test:
- pull the branch locally, run a site-install
- in the browser, open the console and verify that no errors are present (font reference errors were present before)
- Now find a page that has more than one breadcrumb (somewhere like this /communities/confederated-tribes-colville-reservation) and verify that the breadcrumbs look good and function well
- open the inspect window and pull the window to tablet and mobile size to verify that the breadcrumbs still look good at different breakpoints


eckmo:
Merged this with PR 168, so I'll close this out.

PR #186, opened 2023-12-13
Begin theming Browse DH page

eckmo:
Theming the Browse Digital Heritage page
- Tweak template in `mukurtu_browse` to include an aside and have data in semantic order
- Apply the grid and add results and sidebar to the grid
- Fix a typo in a few Digital Heritage templates
- Tweak sass-lint file to allow for IDs
- Add a page template to add additional classes for styling
- Add a note in the readme about the fact that `gulp sass` needs to be run when `gulp watch` isn't picking up changes

*Moved the Map button to the sidebar but opted to leave it over the content area as it makes more sense from a usability/accessibility approach

Separated the search/sort from its view to display on the sidebar - tried having the sort submit on change but the search/sort are part of the same exposed form, so left them tied together for the time being, and moved the Apply button below both for accessibility/clarity's sake

Digital Heritage Browse Map page is now themed

Digital Heritage Browse buttons now correctly switch the view between List/Grid/Map accordingly

Individual Digital Heritage pages are theme, including those with and without the tabs at the top

Additional theme clean up, tying loose ends, etc


eckmo:
@nick-deer I added you as a reviewer on here, but I think other than the two Controller files and some config, this should mostly all live in the theme and should be safe to merge without disrupting any "non-theme" work :)

One "action item" I'll need your help with is getting the [Twig Tweak](https://www.drupal.org/project/twig_tweak) module in the project.

Thanks for being so wonderful to work with, and I hope you have a great journey forward!

nick-deer (approved):
Hi Morgan! Thank you so much for your work on this! Nearly all of it looks great and works as expected. There were only a few things that weren't working for me. 

There were some unmet dependencies and a duplicate config file, but those were easy to fix and I don't mind making those changes myself, so I don't think it's necessary to go over them :)

The only substantial thing is about DH browse. It looks awesome and nearly all of it works great, but the "Sort by" system wasn't responding to me changing a filter and clicking Apply. It's possible I don't have something enabled that is necessary here. Everything else about the sidebar looks great! Michael and Alex will be reviewing these changes too once I approve your PR.

And I can add the Twig Tweak module to our composer.json no problem :)

Thank you again! It was great to work with you and I hope you had some nice time off during the holidays!

PR #300, opened 2024-04-08
add block config to install profile

pjudge:
this should remove the error on gitpod for the main branch and make the site look correct for local users.


nick-deer:
@pjudge looks great!

PR #310, opened 2024-04-17
278: Digital heritage continued

pjudge:
This PR:

- adds styling and template logic to knowledge keepers field in DH sidebar
- updates the knowledge keepers teaching field to be short, unformatted (max char length 255)
- ensures claro and mukurtu_v4 are default themes in install code
- styling for CC images -- I removed some inline styles from the code that seemed to be stripped anyway, and replaced with empty alt text, which is required for decorative images for a11y.
- a single field template for "Original date" on DH items -- I realized I could do this through twig, though I still recommend the date formats show in the UI as options for date fields, so non-coding users can update if they'd like. :) 

The carousel for DH is a work in progress still -- I've been playing around with it locally. It seems the Slick module can integrate with Accessible Slick code, but it requires a module update (Slick ^3.0) which was giving me composer dependency issues. If @nick-deer has any recommendations for that, I'd love to hear them! :)


nick-deer:
Hi @pjudge I spun up your PR in a gitpod instance and got this error: `'block' entity with ID 'stable9_hero_image_and_site_description' already exists.` The site was able to build, but I could not login as admin due to the error

pjudge:
Thanks for letting me know, @nick-deer . There's no config for a stable9 blocks... But I did update the default theme to mukurtu_v4, which looks like it's based on stable9. I'm trying to see if that change in the install code got some things mixed up.

pjudge:
@nick-deer would you mind doing a code review here when you get a chance? i removed config for the two faulty blocks and it seems to have solved the issue. I did also however update the blocks that are created in the install --i just updated their IDs to avoid possible collisions with other themes. I also commented out what looked like duplicate code in the same file. https://github.com/MukurtuCMS/Mukurtu-CMS/pull/310/files#diff-1b376fe2f515a8bbb884dc5e8a686edd685575d80b4595083db0812f97b5ad94R118

Updating the IDs and commenting out the duplicate code did _not_ solve the issue but if you don't mind reviewing those changes, I wonder if we should just commit them anyway. Thanks!

nick-deer:
@pjudge Looks good! I found that your changes actually did make the error go away!

michael-wynne-wsu:
@pjudge I added one new related issue, but otherwise everything included in this PR is looking great to me

Issue #311, opened 2024-04-18
Indigenous knowledge keepers partial dates


michael-wynne-wsu:
Citing Indigenous knowledge keepers renders complete YYYY-MM-DD entries fine, but seems to misinterpret partial (YYYY, YYYY-MM) dates. 3 pairs of screenshots attached.

![Screenshot 2024-04-18 at 10 33 03 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/4144d456-d170-4751-b601-9e21f9d2648d)
![Screenshot 2024-04-18 at 10 33 08 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/640e3c47-1b97-4ea2-bdab-b0aca9f47a64)
![Screenshot 2024-04-18 at 10 33 25 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/fc9b7d46-2c37-4349-8483-32d346f5805b)
![Screenshot 2024-04-18 at 10 33 34 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/e510d46f-0e80-4a89-ba56-c2435d81d724)
![Screenshot 2024-04-18 at 10 35 04 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/4c1c25dc-70b7-42bf-b883-dadd7c67fac9)
![Screenshot 2024-04-18 at 10 35 13 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/3cddd45c-ec08-453f-a910-37d19343195b)

michael-wynne-wsu:
@pjudge this is resolved, thanks!

Issue #280, opened 2024-03-13
Multi-page item display changes needed


michael-wynne-wsu:
1. Page preview/carousel not styled 
2. Page jump doesn’t work

pjudge:
Hi @michael-wynne-wsu Can you expand a bit on #2 above? (Page jump doesn't work) Thanks!

michael-wynne-wsu:
@pjudge This issue is out of date (before you started theming) and is no longer a problem. The "jump to page" dropdown does properly bring the selected node into focus now. I know the styling in the attached screenshot is out of date from your current version, that's a gitpod build from this morning, but it was working in your live demo earlier this week as well.

Unless you have a reason to keep this issue open I think you can close it.

![Screenshot 2024-04-26 at 8 54 01 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/c49af4f6-8691-4721-b996-2e422b075bfa)

michael-wynne-wsu:
@pjudge I think this can be closed - the one outstanding MPI carousel issue is reflected here: https://github.com/MukurtuCMS/Mukurtu-CMS/issues/138

Issue #278, opened 2024-03-13
Digital heritage display changes needed


michael-wynne-wsu:
1. Not all field names are styled as labels. 
2. Would prefer reduced spacing on labels. 
3. In v3 original date rendered as January 5, 2024. Currently showing data as entered YYYY-MM-DD 
4. Creative commons licenses should be smaller/not stacked 
5. All term and entity references are displaying as list – at least SOME should be comma separated instead 
6. Media embed in HTML fields not complete 
7. Related content (bottom of page) not styled/defined 
8. Related content teaser (sidebar) does not exist yet 
9. LC Labels and Notices too large/how to handle text 
10. Rights Statements would be better without the line break after the icon 
11. Do we want to keep community/protocol separate or render that data together? 
12. Citing Indigenous Elders and Knowledge Keepers field not included 
13. Citation field not included 
14. Location (taxonomy) field not included 
15. Parent collection displays in Related Content field, that’s not right, needs to be a distinct field. 
16. No clear contextual edit/add CR, add book page etc menu. 
17. No media carousel for multiple media assets is present

michael-wynne-wsu:
- [ ] Citing Indigenous Elders and Knowledge Keepers seems to be displaying even when the field/paragraph is empty.
- [ ] Might be nice to add sub-divisions to the sidebar. Even just a more prominent horizontal divide. Eg: core mukurtu fields || additional metadata || rights. If that seems like a good idea I can define the groupings.
- [ ] "related content display" > "Display value of the item's related content field, but also include content that includes this item as related content." at /admin/config/mukurtu/settings doesn't seem to be working. This may be a backend issue, @nick-deer ?

michael-wynne-wsu:
- [ ] If possible, only display node title in sidebar when there is a community record/multi-page item. Not necessary for a standalone DH item. LOW priority at this time.

pjudge:
Hi @michael-wynne-wsu !

I'm looking at the link you sent me for Citing Indigenous Elders and Knowledge Keepers and I think the format you're referring to using is this?

```
How to reference Indigenous Elders or Knowledge Keepers
 
Format:

Last name, First initial., Nation/Community. Treaty Territory if applicable. Where they live if applicable. Topic/subject of communication if applicable. personal communication. Month Date, Year.
```
[...]

Issue #138, opened 2023-11-15
Multi-page item thumbnail aspect ratios


michael-wynne-wsu:
In v3, multi-page document (multi-page items in v4) thumbnails above the media asset are a fixed portrait aspect ratio.

Ideally this thumbnail would be responsive to the ratio of each media asset, since some items will be landscape, and others may have mixed media formats and dimensions.

![Screenshot 2023-11-15 at 2 06 24 PM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/c2a1f865-0e24-4a6f-95c4-05fb65efc2b7)
![Screenshot 2023-11-15 at 2 07 20 PM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/7b1f6476-a29a-4bba-b8e6-a9b36bf7e5e8)

michael-wynne-wsu:
@pjudge MPI thumbnails/carousel is displaying as expected with image and audio but the video thumbnails aren't showing for me (gitpod build of https://github.com/MukurtuCMS/Mukurtu-CMS/pull/312)

![Screenshot 2024-04-26 at 12 26 27 PM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/18133920-3b35-4cae-9450-005accd1d6e0)

michael-wynne-wsu:
@pjudge vimeo and youtube carousel preview still not displaying for me after the latest PR.
However, we discussed changing those to use a static thumbnail instead of a video player so that is probably a moot point now. 

If that's logged elsewhere, I think this issue (thumbnail aspect ratios) can be closed.

<img width="1426" alt="Screenshot 2024-04-29 at 1 26 39 PM" src="https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/51e3ca96-9d69-43c0-bf74-1a51eaa4e3a0">

michael-wynne-wsu:
This is great, thank you!

PR #312, opened 2024-04-18
Remove slick and its dependencies; styles carousels

pjudge:
This PR encompasses a few things:

- it removes the slick module and related libraries, in favor of Splide (which was already installed)
- it creates a Splide library inside the theme so it can be easily accessed by theme or modules
- adds a Media Assets display mode for all media types, to display the media assets field correctly on DH nodes
- adds styles and functionality for Multi-page carousel
- adds styles and functionality for media assets carousel (though a bug remains on pages that have both the multipage carousel and the media assets carousel)
- config and styles for embedded media including user friendly names for image styles and minor styling for captions
- initial style idea for local contexts info in the DH sidebar
- fix ajax bug for anons

I think with this PR, the following tickets can be closed:

- #311 
- #280 
- #278 
- #138 
- #85


nick-deer:
@pjudge I spun up a Gitpod instance of this to investigate why Blazy is not being enabled, and the build failed with this message: `Configuration objects provided by <em class="placeholder">mukurtu_dictionary</em> have unmet dependencies: <em class="placeholder">core.entity_view_display.node.dictionary_word.content_browser (blazy)</em>`. I see that `mukurtu_dictionary.info.yml` doesn't have Blazy listed in the dependencies, maybe that's why Blazy isn't showing as enabled? I'm not sure though, until I can get a working build I can't verify if Blazy is enabled. Are you getting this error on your end too? I'm baffled why it's here to be honest--many of our view displays had Blazy as a dependency prior to the changes in your PR here, and they never complained about it, despite Blazy not being officially enabled in the install profile...

pjudge:
@nick-deer Yeah, that's the error I'm trying to debug, and my guess is that it's happening because Blazy isn't being enabled, after removing Slick. 

If you start a gitpod instance from this PR, or my branch, you can `cd` into `web/profiles/mukurtu-cms`.

From there you can `ddev drush cex` to look at `core.extension.yml` and see that Blazy isn't installed. 
Or you can run `ddev drush uli` to log in. (The themes aren't loading but the site is functional.)

pjudge:
Screenshot 2024-04-19 at 3 50 23 PM

nick-deer:
@pjudge I figured out how to resolve the error! Just add blazy as a dependency to `mukurtu_dictionary.info.yml`!

pjudge:
@nick-deer ah, of course! thanks, i'll try that.

nick-deer:
@pjudge It looks awesome, thank you so much! There was only one thing I noticed: I made a DH item with one external embed media item and made it a page in a multipage item. For some reason, the external embed was showing up twice: 
image
This may be related to the bug you mentioned about when an item has both a media assets carousel and a multipage item carousel, but I'm not exactly sure. Otherwise, looks great! I'll wait for everyone else to comment stuff in case I missed something.

michael-wynne-wsu:
@nick-deer It looks like multiple media files are showing in your MPI carousel, that is not what I'm getting. Did you build that from this pull request or main?

What I'm seeing/attaching is what I would expect. In the MPI carousel (the first one) there is a thumbnail of the first/only media asset within the selected page. Then in the media assets field you see that thumbnail again. 

Screenshot 2024-04-26 at 2 50 12 PM

One thing I just noticed is that when I have the same setup as above (a MPI with a DH item page within that has multiple media assets), it loads fine initially, then I move to another page, fine, then I move BACK to the first page and my multiple media asset carousel is no longer displaying.

Screenshot 2024-04-26 at 2 52 57 PM

Screenshot 2024-04-26 at 2 53 04 PM

Screenshot 2024-04-26 at 2 53 09 PM

nick-deer:
@michael-wynne-wsu Oh I see, those are two carousels on top of each other, never mind then! Works as expected.

pjudge:
@michael-wynne-wsu @nick-deer did anyone want to do more testing before i merge?

michael-wynne-wsu:
@pjudge I'm happy for you to merge it!

Issue #308, opened 2024-04-16
Original date display formats


michael-wynne-wsu:
On DH items, Original Date only supports YYYY-MM-DD display format.

Please add additional/standard date display formats (eg: January 5, 2024)

Note that this is a custom piece of code since the v3 partial date module was not available. This is a backend issue for @nick-deer and NOT a theming issue for @pjudge

nick-deer:
@michael-wynne-wsu added another display format! Let me know if you have any suggestions, especially with the naming convention of the display formats.

michael-wynne-wsu:
@nick-deer 

For naming, what about "Original Date display - YYYY-MM-DD" and "Original Date display: Month Day, Year". That's a bit easier for me to read. But if there is an established naming convention for these kind of settings, let me know.

Right now they are both working in Bartik, but Mukurtu theme is only displaying "Month Day, Year" regardless of selection. I'm assuming that'll change once approved and you notify Pauline of the update. But yes, I really like having both options available and it's working as expected in bartik!

michael-wynne-wsu:
@nick-deer One tiny note on the date, with "Month Day, Year" format, partial dates include spaces and comma as if the date was complete. But using both formats worked perfectly otherwise.

![Screenshot 2024-05-02 at 1 19 06 PM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/39b673c2-3945-4986-8762-8fb70e704110)
![Screenshot 2024-05-02 at 1 19 52 PM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/a31354cc-9107-4395-8a19-8cddc2100d70)

michael-wynne-wsu:
Confirmed complete! @nick-deer

Issue #305, opened 2024-04-12
DH full content needs media carousel


michael-wynne-wsu:
Currently media display inline - needs a medial carousel like in V3.

michael-wynne-wsu:
Example: https://plateauportal.libraries.wsu.edu/digital-heritage/k%E2%80%99%C3%BAusey%E2%80%99ne-sapo%CB%80saap%C3%B3%CB%80s-womans-saddle-nez-perce

1. When items have enough media assets that the thumbnail carousel becomes scrollable, users can get lost and experience "infinite scroll". Conversely, having to scroll back to the beginning may be annoying.
3. Consider adding a counter (eg: 1/4) and/or displaying filenames/alt text so users can orient themselves (I believe this happens in the v3 zoom image tool).
5. Carousel thumbnails are a fixed size/ratio. They should reflect the proportions of the thumbnail - no cropping.
7. A small number of users prefer NOT to use a carousel. Supporting an inline or alternate display where they can access multiple media assets simultaneously would be a nice option.

michael-wynne-wsu:
@pjudge This was probably a misunderstanding on my part, but I was expecting the carousel would be to select the media and have a more full-size display of the media below. Can you confirm that the screenshot below is showing the whole media assets field as you expected?

The carousel itself is working great though!

![Screenshot 2024-04-26 at 12 30 55 PM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/9af5ccd2-f7cd-41f3-a749-679a978655d9)

pjudge:
Thanks for reminding me @michael-wynne-wsu -- this was something I wanted to bring up in our next sync. I think that can be done, however I'm wondering about having too much happening on the page at once. Each media asset has a link button that takes the user to the full media for them to experience. That's a click away, but it does improve site performance to not have to have a linked slider. The page gets visually confusing when there's a multipage carousel at the top, and then a page with multiple media assets below. 

I would recommend that if we do keep the larger version of the media asset showing below the carousel, the carousel itself should have much smaller thumbnails. 

I made them larger here as a way to provide users with a decent-sized visual, large enough to allow them to quickly decide if they'd like to click to access the full size version. Just my thoughts -- I can do whichever!

michael-wynne-wsu:
Definitely worth discussing next week!

- Yes, there *can* be way too much going on with the multi-page item carousel, community record tabs, AND multiple media assets. 
[...]

PR #326, opened 2024-05-02
305: carousels final form

pjudge:
This PR provides a more finalized round of styles and function for synced carousels that appear when the media assets field on DH nodes has more than one item. This PR also adds some updated templates and config for display modes, specifically for when content appears on browse pages, below main content under related content, on sidebar under related content (though that is still WIP).

I believe this PR closes the following tickets:

- https://github.com/MukurtuCMS/Mukurtu-CMS/issues/305
- https://github.com/MukurtuCMS/Mukurtu-CMS/issues/308


michael-wynne-wsu:
@pjudge One tiny note on the date, with "Month Day, Year" format, partial dates include spaces and comma as if the date was complete. But using both formats worked perfectly otherwise.

Screenshot 2024-05-02 at 1 19 06 PM

Screenshot 2024-05-02 at 1 19 52 PM

nick-deer:
@michael-wynne-wsu Regarding the partial date display, I think that's an issue with the formatters I worked on not accounting for partial dates. I can change that!

michael-wynne-wsu:
I made some kind of cursed carousels and am going to break out independent issues for each different component. Overall this is working really well, just have some questions and a couple notes. Potentially we could move some of these to lower priority if you feel the need to move on to something else, because it's very functional, mostly looking at minor adjustments (I think).

https://github.com/MukurtuCMS/Mukurtu-CMS/issues/308 (this isn't really the focus of this PR, just flagging it since I noticed while testing CR tabs + media assets carousel, but it's the same regardless of how many media assets)

https://github.com/MukurtuCMS/Mukurtu-CMS/issues/328 (the default graphics from v3 are ugly)

https://github.com/MukurtuCMS/Mukurtu-CMS/issues/329 (applies to carousel as well as single media)

https://github.com/MukurtuCMS/Mukurtu-CMS/issues/330 and https://github.com/MukurtuCMS/Mukurtu-CMS/issues/331 (one is about media and one is about multipage item but the points are basically the same).

michael-wynne-wsu:
> @michael-wynne-wsu Regarding the partial date display, I think that's an issue with the formatters I worked on not accounting for partial dates. I can change that!

@nick-deer great I'll move my comments into the other date format issue.

nick-deer:
@pjudge I think this looks great regarding our requirements! I do agree with Michael's comments, but I will merge this PR as the base functionality is there.

michael-wynne-wsu:
Yes, thank you Nick!

Issue #273, opened 2024-03-13
Navigation menu changes


michael-wynne-wsu:
1. Needs to support dropdown/submenus (theming) 
2. Need to be able to rename stock links (“This link is provided by the Mukurtu Browse module. The title and path cannot be edited.”) 
3. Have run into issues in the past with people having a lot of links (or a few long links), and the display doesn’t handle that well. 
4. Active links need to be indicated. 

Controlled at /admin/structure/menu/manage/main

alexmerrill:
To add another item here -- the /dashboard can be removed from the Main Navigation Menu as it appears contextually in the User Account menu -- having it both places seems redundant.

michael-wynne-wsu:
[like]  Wynne, Michael reacted to your message:
________________________________

michael-wynne-wsu:
Both - theming is theming, changing links is back end.

PR #568, opened 2024-08-15
Adding helper text to Location Description DH field and removing second Dashboard link in Menu (Part Deux)

alexmerrill:
Title says it all, I quite literally only did those two things here. The menu link removal was added to issue #273 as a comment so this would not close the entire issue, just my own additional comment. I commented out the menu link only because I noticed other folks commenting items (instead of deleting) in similar spots.

### DIFF
--- a/modules/mukurtu_core/mukurtu_core.links.menu.yml
+++ b/modules/mukurtu_core/mukurtu_core.links.menu.yml
@@ -5,13 +5,13 @@ mukurtu_core.dashboard:
   menu_name: mukurtu
   route_name: mukurtu_core.dashboard
   weight: -1000
-mukurtu_core.dashboard_main:
-  title: 'Dashboard'
-  description: 'Dashboard'
-  parent: main
-  menu_name: main
-  route_name: mukurtu_core.dashboard
-  weight: 1000
+#mukurtu_core.dashboard_main:
+#  title: 'Dashboard'
+#  description: 'Dashboard'
+#  parent: main
+#  menu_name: main
+#  route_name: mukurtu_core.dashboard
+#  weight: 1000
 mukurtu_core.dashboard_admin:
   title: 'Mukurtu Dashboard'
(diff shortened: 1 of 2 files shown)


modules/mukurtu_core/mukurtu_core.links.menu.yml
```diff
@@ -5,13 +5,13 @@ mukurtu_core.dashboard:
...
   route_name: mukurtu_core.dashboard
   weight: -1000
-mukurtu_core.dashboard_main:
-  title: 'Dashboard'
-  description: 'Dashboard'
-  parent: main
-  menu_name: main
-  route_name: mukurtu_core.dashboard
-  weight: 1000
+#mukurtu_core.dashboard_main:
```
jonscottwsu, line 8: I am of an opinion that all 'non-context-documenting' comments in yml files can safely be removed since we can capture before and after (and diff side-by-side) any changes but defer to you both if you like having historical context captured going forward

michael-wynne-wsu: In this case specifically, it's removing something that wasn't intended to be there in the first place, so I don't think it's a change that needs to be retained like that.

alexmerrill: Sounds good to me I am going to pull this back to a draft and look at the other two items that are commented out and see if it makes sense just to remove those as well...

alexmerrill:
I went back and removed, instead of commented out the Dashboard link,  checked in on the other two links that were commented out and removed the Create Person link as it was also redundant on the dashboard. I left in the comments around the "Manage All Site users" link -- it looks like that it should be on the todo list as the target of the link is a view that has not yet been created.  I couldn't find an issue for the link -- but will create one soon.

jonscottwsu (approved):
looking good to me - thank you for cutting out commented yml

PR #576, opened 2024-09-27
Update README.md

michael-wynne-wsu:
Testing local editing.

Made a minor text edit to a readme file to test my setup. @nick-deer can you make sure that this seems like the proper workflow, valid PR etc? We can discuss before you accept it.


nick-deer (approved):
Looks good!

michael-wynne-wsu:
@nick-deer looks like this build is failing checks and failing to build- not sure if it's related to what I pushed or not?

michael-wynne-wsu:
Sounds good, maybe @jonscottwsu has an idea about the test suite?

There is one thing I want to confirm on this build before doing more - I haven't checked a ddev build yet though. Will do that now.

michael-wynne-wsu:
@nick-deer @jonscottwsu It built fine in ddev. I was testing a minor edit to a field description in the hopes that I can start implementing things like that without troubling you and it seems to have worked, so that's excellent!

PR #604, opened 2024-10-30
599: the locally hosted patch needs to be added to the profile first

pjudge:
Adding the patches directory to the profile so that the patch will exist when the site installs.


pjudge:
@nick-deer Hi Nick, I think this PR needs to be merged before the other PR can work without breaking gitpod. 😄

PR #605, opened 2024-10-30
use remote URL for patch

pjudge:


pjudge:
Well, the patch applied without error but it isn't fixing the empty div issue. Sigh. I'll keep working on this.

nick-deer:
I'll merge this anyway since it fixes the gitpod errors!

Issue #611, opened 2024-11-05
Run tests on Pull Requests


quicksketch:
Right now tests are run on the `main` and `develop` branches:

```
on:
  push:
    branches:
      - main
      - develop
```

But they're not run on pull requests. Ideally we'd run the tests before we merge things and not afterward.

PR #614, opened 2024-11-06
611: Run tests on pull requests.

quicksketch:
Fixes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/611

This adds running tests when pull requests are filed.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - main
       - develop
+  pull_request:
+    types: [opened, synchronize, reopened]
 jobs:
   testing:
     name: Mukurtu v4 - PHP ${{ matrix.php-versions }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -4,6 +4,8 @@ on:
     branches:
       - main
       - develop
+  pull_request:
+    types: [opened, synchronize, reopened]
```
quicksketch, line 8: These are actually [the default types](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request), but I specified them for clarity since there are a lot of potential event types that may fire related to pull requests.

michael-wynne-wsu:
@nick-deer @jonscottwsu Any reason not to merge this?

nick-deer (commented):
Looks great thank you!

PR #624, opened 2024-11-11
Uncheck the checkbox to ensure empty paragraph doesn't render

pjudge:
This took entirely too long, but I believe this piece of config is all that's needed to ensure the empty paragraph doesn't render. 

The patch is applying, and it provides a boolean on the edit paragraph screen. By default the box is checked, I think likely to not interfere with existing paragraphs on a production site. But in our case, we want the checkbox UNchecked.


michael-wynne-wsu:
This looks perfect! @nick-deer please merge when you have a moment.

Issue #683, opened 2025-01-24
D10 compatible modules


michael-wynne-wsu:
Make sure all modules are up to date and complaint with D10. Remove unneeded modules.

quicksketch:
Initial PR to at least specify all of our modules as being D10 compatible here: https://github.com/MukurtuCMS/Mukurtu-CMS/pull/697

quicksketch:
All contrib modules are already D10 compatible. :tada:

michael-wynne-wsu:
Hey that's fantastic!

quicksketch:
Completed in https://github.com/MukurtuCMS/Mukurtu-CMS/pull/697.

PR #698, opened 2025-02-05
683: Set accessCheck() on all entity queries as required for Drupal 10+.

quicksketch:
Part of https://github.com/MukurtuCMS/Mukurtu-CMS/issues/683

This updates our custom code to handle one of the biggest API changes in Drupal 10, which requires setting `accessCheck()` specifically on all `entityQuery()` usages. The default behavior was to set `accessCheck(TRUE)` previously, so anywhere we add that line results in effectively no change. There were a few places where `accessCheck(FALSE)` seemed like the more correct solution.

See https://www.drupal.org/node/3201242

This change should be compatible with Drupal 9 and should be able to be merged separate from the D10 upgrade.


michael-wynne-wsu:

Link:
- https://pr698-s4uxgfnnpmx6meyngxnxplt4wxfkk3lh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/67a3b668f5294b5a5d10c4e4

michael-wynne-wsu:
@quicksketch Do you think a functional review of this is needed?

quicksketch:
> Do you think a functional review of this is needed?

The places where I changed the access check to FALSE are the only functional differences. It's easy to check the UID `#1` warning and that is working correctly:

image

However the change I don't know how to verify is the validation code in mukurtu_community_records. It looks like the code is trying to check when an "Original Record" is created/saved, it cannot also be a "Community Record". I don't really know what that means or how to test it.

Because that code looks like it's only validation code, and that conditional checking seems like it wouldn't need to take into consideration the current user's access, I'm pretty sure it's a safe change, but I would love if we had a way to confirm that.

michael-wynne-wsu:
An item not being able to be both an original record and a community record is correct. After an item is created, additional items can be created and attached to it as "community records" - they are effectively independent entities but are bundled together in display. 

^ Best effort at a high level description, let me know if you want more info.

Screenshot 2025-02-05 at 11 52 20 AM

Screenshot 2025-02-05 at 11 53 04 AM

Screenshot 2025-02-05 at 11 53 08 AM

michael-wynne-wsu:
I'm happy for you to merge this unless anyone else is doing more testing.

michael-wynne-wsu:

Link:
- https://pr698-s4uxgfnnpmx6meyngxnxplt4wxfkk3lh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/67a3b668f5294b5a5d10c4e4

quicksketch:
I added in other D10 compatibility fixes as reported by the https://www.drupal.org/project/upgrade_status module.

- Support from all Views contextual filter settings for the default_argument_skip_url setting is removed from drupal:11.0.0. No replacement is provided. See https://www.drupal.org/node/3382316.
- render() replaced with Renderer service - https://www.drupal.org/node/2939099
- Call to deprecated function `drupal_get_path()`. Deprecated in drupal:9.3.0 and is removed from drupal:10.0.0. Use Drupal\Core\Extension\ExtensionPathResolver::getPath() instead.
- Call to deprecated method `assertEqual()` of class `Drupal\KernelTests\KernelTestBase`. Deprecated in drupal:8.0.0 and is removed from drupal:10.0.0. Use $this->assertEquals() instead.
- Some permissions had invalid names, seemingly due to the source module having updates. Corrected these names.

Now Upgrade Status has a clean report for deprecated code segments.

michael-wynne-wsu:
Fantastic news!

Issue #678, opened 2025-01-22
Upgrade to D11


michael-wynne-wsu:

quicksketch:
Upgrade Status flags the following modules as not yet being D11 ready:

- Blazy (blazy)
- Colorbox (colorbox)
- Content Browser (content_browser)
- Facets Reset Button (facets_reset_button)
- Features (features)
- Leaflet (leaflet)
- Message Subscribe (message_subscribe) 

Most or all of these probably have a patch ready in their issue queues.

The Mukurtu installation profile has 61 problems flagged, most of which seem pretty easy to resolve.

And we should stop installing the Tour module by default, or switch to the contrib module. Mukurtu could actually benefit from the gentle on-boarding provided by Tour module, but that should probably be a separate task from Drupal 11 upgrade work.

Overall, Drupal 11 doesn't seem far out of reach.

quicksketch:
The following modules need to be updated or we need to use a custom upstream. I'll reach out to the maintainers to see if I can accelerate the fixes being applied:

Content Browser: https://drupal.org/node/3503979 (created new MR)
Features: https://drupal.org/node/3447460 (existing MR)
Facets Reset Button: https://drupal.org/node/3430417 (existing MR)

The other modules all have D11 versions that seem like they should be fine.

PR filed at https://github.com/MukurtuCMS/Mukurtu-CMS/pull/707 that makes all the necessary code changes according to Upgrade Status.

michael-wynne-wsu:
Thanks Nate, this looks great.

quicksketch:
I filed a PR to just prepare for Drupal 11 while staying on Drupal 10 here: https://github.com/MukurtuCMS/Mukurtu-CMS/pull/712

In particular it updates Blazy, Colorbox, Facets, Leaflet, and Message Subscribe to new major versions. I checked the release notes for each of them though and none seemed to indicate major changes. There weren't even any configuration changes needed. Blazy has a few update hooks. Overall, pretty small changes it looks like.

Nonetheless, it would be great to do a thorough test run-through on https://github.com/MukurtuCMS/Mukurtu-CMS/pull/712 to verify the updated didn't cause any other regressions.

michael-wynne-wsu:
I'll try to run through that today.

quicksketch:
[...]

PR #706, opened 2025-02-24
678: Disable Tour module, which is removed from Drupal 11.

quicksketch:
Part of #678: We should disable Tour module, as it is removed from Drupal 11. Tour module provides the ability to show popup hints, walking through steps to introduce software. This may actually be useful for Mukurtu, but we would pivot to the contrib module: https://www.drupal.org/project/tour

For the sake of upgrading to Drupal 11, we just need to turn this module off for the time being. There does not seem to be any existing integrations in Mukurtu.

image

### DIFF
--- a/modules/mukurtu_core/mukurtu_core.features.yml
+++ b/modules/mukurtu_core/mukurtu_core.features.yml
@@ -16,7 +16,6 @@ excluded:
   - core.entity_view_mode.node.token
   - core.entity_view_mode.path_alias.token
   - core.entity_view_mode.shortcut.token
-  - core.entity_view_mode.tour.token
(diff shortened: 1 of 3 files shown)


michael-wynne-wsu:

Link:
- https://pr706-nznefpqoen5biqt1covwtbgxujqhlv1p.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/67bceb97b029347f4b232e33

michael-wynne-wsu:
Thanks! I've logged this for future action here as well: https://github.com/MukurtuCMS/Mukurtu-CMS/issues/708

Issue #343, opened 2024-05-17
Map view changes (browse pages, home page, etc)


michael-wynne-wsu:
Map views (on all browse pages, front page, probably everywhere they appear by default) should use a "sidebar" style content preview (or a recommended alternative) instead of an on-map popup.
- We used on-map popup in v3, but that has accessibility concerns.

Notes:
- Leaflet is still not fully ported over with all of the features we need, but I haven't checked on that for a while.
- V3 did not have a good points clustering system. We were waiting for leaflet to add that. Good example of clustering: https://ewatlas.net/
- Map view of all mapped content (subject to protocols, search, facets, etc).
- Some way to preview the content from the map (title, thumbnail/media player, select metadata fields).
- From the preview be able to click through to the content.
- Be aware that we support multiple locations for each piece of content, and that can include any combination of points, lines, rectangles, and polygons. (Not circles, but I think that was also a waiting for leaflet thing).

![Screenshot 2024-05-16 at 8 09 38 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/6135364c-f0aa-47a1-8749-26fb10514dc7)

![Screenshot 2024-05-16 at 8 09 59 AM](https://github.com/MukurtuCMS/Mukurtu-CMS/assets/88502274/500a8ba7-09dd-46e7-9368-602cdf0069fe)

michael-wynne-wsu:
Clean up issue, YES explore moving to side panel preview instead of popup (accessibility, etc)

michael-wynne-wsu:
@pjudge A few initial discoveries...
1) Thinking back on what I'd seen, and reading a bit more, I suspect you're right and that the popup layer was just not displaying properly on the map and was rendering out of bounds.
2) There IS a leaflet sidebar plugin, but I haven't dug into it's version requirements, etc: https://github.com/turbo87/leaflet-sidebar/
3) Leaflet does make an effort to have good accessibility (https://leafletjs.com/examples/accessibility/), so maybe my concerns on that front are overblown and a sidebar view isn't as beneficial for that.

Would love your thoughts. The major issues we had in v3 was the content preview not reliably fitting in the popup, and the map not adjusting and popups displaying half on/half off the map - those things are what I wanted to avoid more than anything.

pjudge:
[...]

PR #725, opened 2025-03-17
Style Leaflet previews

bronzehedwick:
Initial pass at styling the map previews. I made some light design decisions, and generally tried to keep everything inside the preview and at a reasonably small size. I added a new Leaflet Overrides CSS file to make the tweaks, that is only loaded when the Leaflet map is rendered.

There's more work left to do in the task that will be coming soon, but wanted to slice this chunk out for review now.

Open to any feedback. Thanks!

Re #343


michael-wynne-wsu:
A great start! Looks like you focused on the DH item previews, right? (I didn't check PDFs, video upload, or embeds this time)

Thoughts/notes/questions - I'm always up for discussion on these of course. 
1) I don't think we need the extra white space and bottom grey line below the title.
Screenshot 2025-03-17 at 11 32 17 AM

2) Overall the previews feel a bit small - any opinion on bumping the title text size up a little? I'm thinking we may want to set minimum preview dimensions? 

3) Related, when the title is short it really shrinks down the size of the preview (am I correct that the width is based off title length?), and with that, how is the height calculated? Examples are especially notable with the audio player that can be so small it completely collapses.
Screenshot 2025-03-17 at 10 12 53 AM

4) Vimeo seems to love including excess top and bottom padding. This is most obvious with a smaller preview, but it's something we see across our sites and elsewhere as well.
Screenshot 2025-03-17 at 10 13 53 AM
Screenshot 2025-03-17 at 10 14 46 AM

5) The layer button in the top right is exposing I think a per-item filter? Not sure if we want to use that for anything in particular or just get rid of it. I can think of some other tools that would be neat in the map but they're probably future tickets.
Screenshot 2025-03-17 at 10 14 07 AM

bronzehedwick:
Thanks @michael-wynne-wsu! Responses below.

> A great start! Looks like you focused on the DH item previews, right? (I didn't check PDFs, video upload, or embeds this time)

Right, this PR doesn't address PDF preview generation, although it should work for direct video uploads or generic embeds.

> I don't think we need the extra white space and bottom grey line below the title.

Sure thing, I'll remove that in this context.

> Overall the previews feel a bit small - any opinion on bumping the title text size up a little? I'm thinking we may want to set minimum preview dimensions?

Sure, no issue bumping size. I left it as the default vertical card styles specifically to gauge feedback. I'll play with this a bit, but I'm thinking getting closer to the horizontal card font size.

> Related, when the title is short it really shrinks down the size of the preview (am I correct that the width is based off title length?), and with that, how is the height calculated? Examples are especially notable with the audio player that can be so small it completely collapses.

Yes, the Leaflet library inserts inline styles to make the preview roughly the width of the content. I can play with adding some overrides for this, although for instances similar in the past, that can sometimes mess up the libraries geometry calculations, so I may report something along those lines.

> Vimeo seems to love including excess top and bottom padding. This is most obvious with a smaller preview, but it's something we see across our sites and elsewhere as well.

Gotcha. I'll test Vimeo specifically and see if I can tame that padding.

> The layer button in the top right is exposing I think a per-item filter? Not sure if we want to use that for anything in particular or just get rid of it. I can think of some other tools that would be neat in the map but they're probably future tickets.

I can look into disabling this for now, and then we can re-evaluate if/when a use is found.

michael-wynne-wsu:
@bronzehedwick 

> Sure, no issue bumping size. I left it as the default vertical card styles specifically to gauge feedback. I'll play with this a bit, but I'm thinking getting closer to the horizontal card font size.
Great! I'm not sure what the horizontal card size it, but we're definitely not restricted to the vertical style, if you think the horizontal layout (or anything else) is worth using/exploring too.

> Yes, the Leaflet library inserts inline styles to make the preview roughly the width of the content. I can play with adding some overrides for this, although for instances similar in the past, that can sometimes mess up the libraries geometry calculations, so I may report something along those lines.
Increasing the font size as above might be enough to address this as well. I do think keeping the audio player from shrinking so far it collapses would be good though.

> I can look into disabling this for now, and then we can re-evaluate if/when a use is found.
Perfect, I suspect we won't need it.

bronzehedwick:
@michael-wynne-wsu Looks like Leaflet has min/max preview widths, which can be configured per instance. The `/digital-heritage` one has the max at 300, and the min at 50. I'll play with increasing those.

bronzehedwick:
Pushed fixes for what you brought up above, @michael-wynne-wsu. LMK if I should adjust!

michael-wynne-wsu:
@bronzehedwick this is looking great.

Agreed that the black backgrounds for both remote video sources is an improvement for consistency.

Increased font size, removing the vertical borders, removing the extraneous layer control all looks good.

The only things that still feel off right now are:
1) Audio player not working, but you noted that already.
2) For the video (and to a lesser extent audio), the previews are still just too small (at least at the size on my displays) with a short title to be really accessible. Any chance we could bump minimum width to ~300px? See below for unaltered (~130px) vs 300px wide.

Screenshot 2025-03-17 at 2 38 54 PM

Screenshot 2025-03-17 at 2 38 41 PM

michael-wynne-wsu:

Link:
- https://pr725-zxqtwpxozvcfq7zslibrm9euqat7iito.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/67d8545ee6cc8e4e840f338b

bronzehedwick:
Hey @michael-wynne-wsu, thanks for the feedback!

> Audio player not working, but you noted that already.

Yes, still working on that one. I'd like to slice that work out to a separate PR, as it is unrelated to these style changes, and is taking longer than I'd like. I can speak more to it in today's meeting.

> Any chance we could bump minimum width to ~300px? See below for unaltered (~130px) vs 300px wide.

Done! I've changed the minimum preview width to be `300px`, and increased the maximum to `400px`. Let me know how that feels.

michael-wynne-wsu:
@bronzehedwick agreed re: audio player, and I stubbed out #727 for that.

I'm not seeing the updated minimum card width reflected in the tugboat build yet

michael-wynne-wsu:
@bronzehedwick Ahhh I was looking at /browse, NOT /digital-heritage, so YES it's looking good at /digital-heritage!

bronzehedwick:
@michael-wynne-wsu Thanks! Should have specified, and I'll go ahead and find and update the other views/content types with the new min/max.

michael-wynne-wsu:
@bronzehedwick Do you want to merge this PR as is or are you planning to include more in it?

bronzehedwick:
@michael-wynne-wsu I'd like to hold off merging for a moment while I migrate everything over to use the Map Display Mode instead of the List Display Mode; turns out keeping them the same was causing the List view to regress!

michael-wynne-wsu:
@bronzehedwick makes sense!

bronzehedwick:
Okay @michael-wynne-wsu, pushed the fixes for the Map Display Mode! Should probably be looked at one more time, and then we can merge.

Note that I'll handle the `/browse` and other pages once Nate follows up on that PHP error, as well as the other unhandled tasks in the main task list, in follow up PRs.

michael-wynne-wsu:
@bronzehedwick the previews are looking great - I think we can merge this one.

michael-wynne-wsu (approved):

PR #756, opened 2025-03-25
655 Fix multiple collections links for a11y

### LINKED ISSUE #655: Confirm if multiple text links to collections on /collections fails accessibility checks
If it's a warning we can leave as it, otherwise we need to remove one (probably the title).

bronzehedwick:
Confirmed the `/collections` page duplicate item links pass accessibility checks (ran against axe).

CleanShot 2025-03-25 at 10 04 11@2x

I did notice two other smaller issues, however.

## Duplicate links are discouraged

The first, larger issue, is that while it is not a WCAG requirement, it is a best practice to not have duplicate links to the same URL for accessibility. See [Can I have multiple links that go to the same page?](https://accessibleweb.com/question-answer/what-should-i-do-if-i-have-multiple-links-that-go-to-the-same-page/)

> When there are text links in addition to image links to the same destination, it is best practice to hide one of these duplicate links from users of assistive technology (AT). Although not technically required by WCAG, it would be a better user experience for AT users as this would prevent unnecessary tab stops and repetitive announcements.

The suggested solution is to add `aria-hidden="true"` and `tabindex="-1"` to the second link to remove it from the accessibility tree and keyboard navigation. It's a simple fix, and I've added that to this PR.

## Icons should be hidden for a11y

While investigating the issue above, I ran the regular browser tool accessibility checker, and found an issue with SVGs. SVGs, being markup, should have an accessible label or be hidden from the accessibility tree if they are purely presentational. All the icons seem presentation-only, which is a common pattern. I added `aria-hidden="true"` to all the SVG icons.

The Logo is *not* purely presentational, so I gave it the proper `role="img"` so screen readers will treat it as a graphic (SVGs default to a document role), and gave it an accessible label.

Finally, I minified all the SVGs, to save a few bytes.

None of these changes should have any impact on the display.

Fixes #655


michael-wynne-wsu:

Link:
- https://pr756-qvflccdkdlxu35xzuos2xwbwu2uc9jjr.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/67e2c0b506460d78817e9bd2

michael-wynne-wsu (approved):
Accessibility upgrades are much appreciated! And confirming that I see no functional/display changes.

Issue #724, opened 2025-03-14
Map can do weird things to point placement at extreme zoom levels


michael-wynne-wsu:
I don't know that this is a big issue or something we need to pursue (and is likely with Leaflet, not us), but when I was fully zoomed out while placing map points, the worldmap will let you "repeat" continents...

![Image](https://github.com/user-attachments/assets/5e9419ff-e7e6-4e86-857c-dd598dfb6cf1)

![Image](https://github.com/user-attachments/assets/dcfa6fc4-02b1-444b-a23f-30067fb98ad0)

![Image](https://github.com/user-attachments/assets/5ca4dfdc-cc86-4e33-aa47-06757d6d8eaf)

michael-wynne-wsu:
Additionally, when moving the map (on browse pages), sometimes the points will disappear and reappear.

quicksketch:
I've been seeing this issue too. We could attempt to mitigate it by setting a minimum zoom level maybe?

PR #762, opened 2025-03-26
724: Set a minimum zoom level on Leaflet to prevent placing invalid markers.

quicksketch:
Fixes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/724 by setting a minimum zoom on location input.

To test:

- Create a word, collection, DH, or person.
- Under "Additional Fields", scroll down to the map field.
- Try to zoom out the map in a way that the entire world is visible. Should not be possible.
- Try to place a marker on the same content on both the left and right side of the map. Should not be possible.
- View the saved content. Note that the map can be zoomed out to the global level on view.


michael-wynne-wsu:
@quicksketch YES that seems to resolve it for me!

quicksketch:
Perfect, thanks seems like a safe and easy compromise to me too.

PR #866, opened 2025-05-28
Update mukurtu_media.module

michael-wynne-wsu:


michael-wynne-wsu:
Removing extraneous "map points" field display from DH full template.

Issue #631, opened 2024-11-19
Remove custom repositories for colorbox and masonry


quicksketch:
Right now [our template repository](https://github.com/MukurtuCMS/mukurtu-template/blob/main/composer.json) includes two manual repositories for `colorbox` and `masonry`:

```
    "repositories": {
        "drupal": {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        },
        "asset-packagist": {
            "type": "composer",
            "url": "https://asset-packagist.org"
        },
        "colorbox": {
            "type": "package",
            "package": {
                "name": "mukurtu/colorbox",
                "version": "1.6.4",
                "dist": {
                    "url": "https://github.com/jackmoore/colorbox/archive/master.zip",
                    "type": "zip"
                },
                "type": "drupal-library"
            }
        },
        "masonry": {
            "type": "package",
            "package": {
                "name": "mukurtu/masonry",
                "version": "4.2.2",
                "dist": {
                    "url": "https://github.com/desandro/masonry/archive/v4.2.2.zip",
                    "type": "zip"
                },
                "type": "drupal-library"
            }
        }
    },
```

To make our dependencies more maintainable, we should investigate if it's possible to replace these custom repositories with an official repository. This should be possible through https://asset-packagist.org/

quicksketch:
Just a note here that although these packages exist with the [same names on bower](https://asset-packagist.org/package/search?query=masonry&platform=bower%2Cnpm):

![image](https://github.com/user-attachments/assets/91150719-b4b2-4222-b7a8-d9b7830ca61c)

We shouldn't use bower at this point [because it's deprecated](https://bower.io/blog/2017/how-to-migrate-away-from-bower/).

So we'll need to convert the paths to the new names, which will be `masonry-layout` and `jquery-colorbox`.

quicksketch:
I filed an initial PR for the template repository here: https://github.com/MukurtuCMS/mukurtu-template/pull/1

Next we'll need to update this repository as well to use the new npm-provided packages.

michael-wynne-wsu:
Well that's annoying! 

Clearing out Masonry, and moving this to a "future work" issue (to be addressed once Drupal Core knows what they're doing) seems like a reasonable solution to me.

quicksketch:
PR at https://github.com/MukurtuCMS/Mukurtu-CMS/pull/999 for this repository. Once removed, we can remove it from the template repository as well.

[...]

PR #999, opened 2025-07-22
631: Remove custom Masonry dependency.

quicksketch:
Part of https://github.com/MukurtuCMS/Mukurtu-CMS/issues/631

While trying to consolidate all our dependencies into the main repository, I found that `masonry` may not be needed because it's now a dependency by the `mukurtu_v4` theme. See [package.json#L32](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/themes/mukurtu_v4/package.json#L32).

I think that means we don't need to install it through composer at all, because the copy in `libraries/masonry` is not actually used.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,6 @@
         "drupal/views_bulk_operations": "*",
         "drush/drush": "^13",
         "mukurtu/colorbox": "*",
-        "mukurtu/masonry": "*",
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu (approved):
Looks good to me. So far I haven't spotted any unexpected front-end changes as a result.

PR #1023, opened 2025-07-29
1004 - Displaying protocol and community main region if empty

### LINKED ISSUE #1004: Community and protocol pages alignment
On both community and protocol pages, if there is no description added, the page alignment (right sidebar) is off. once a description is added, the alignment is correct. Page should always have that same alignment, whether the description is empty or not.

Image

Image

Image

Image

Javi-er:
Issue #1004 

### Description
This PR adjust protocol and community so it shows the main region even if empty.

### How To Test

1. Login to Tugboat
2. Create a community and a protocol with empty content, validate that the sidebar is in the correct location.


michael-wynne-wsu (approved):
@Javi-er works perfectly, please merge!

Issue #1131, opened 2025-10-06
DH item transcription field


michael-wynne-wsu:
Currently set to string, should be text_long (same as v3)

PR #1133, opened 2025-10-06
DH item transcription field

nick-deer:
Addresses https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1131


michael-wynne-wsu (approved):
Yep that's perfect, thanks!

Issue #923, opened 2025-06-10
Dictionary language taxonomy - decision needed


michael-wynne-wsu:
Would like more controls over which languages are available for the dictionary.

In v3 it was tied to the "language community" which is not in v4.

Exposing all languages when creating a word isn't a great user experience.

Options: 
- Use a separate “dictionary language” vocabulary. 
- Find a way to (site-level?) select which language terms are available for dictionary words.

PR #1139, opened 2025-10-13
Add sitewide dictionary language selection config

nick-deer:
https://github.com/MukurtuCMS/Mukurtu-CMS/issues/923


michael-wynne-wsu (changes requested):
@nick-deer works perfectly, thanks!

One question: if possible, when there is only one language available, default to that language (since there is no choice to be made). I suspect MOST sites only have one language. But let me know if that seems messy.

One note: disabling a language after words already use it doesn't remove it from or disable those words. And I think that's completely okay, just noting that I AM aware of it.

Screenshot 2025-10-14 at 8 52 55 AM

Screenshot 2025-10-14 at 8 53 33 AM

michael-wynne-wsu (approved):
@nick-deer perfect, thank you! I added in the updated language field description as well. Merge whenever you're ready.

Issue #1152, opened 2025-10-22
Cannot create CR on site-initial DH item


michael-wynne-wsu:
Seems like the first created DH item cannot be given a community record. All subsequent items work fine. There just is no button to do so.

<img width="1425" height="758" alt="Image" src="https://github.com/user-attachments/assets/6cbfe212-5050-47f0-967c-d7f21641e154" />

PR #1154, opened 2025-10-22
1152: Cannot create CR on site-initial DH item

nick-deer:
Addresses https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1152

The problem was that the community record field was only being loaded and attached to digital heritage upon saving the Community Records Content Types settings form. However, since digital heritage is enabled by default, no user would think to save the form again. To solve this, I created the community record field and attached it to digital heritage on site install.


michael-wynne-wsu (approved):
Fantastic, thank you!

christa-kavanagh-wsu (commented):
Hey Nick! I checked it out this morning and that seems to have solved the problem, thank you!

PR #1191, opened 2025-11-18
1174: Add an option to re-create the default landing page after the migration finishes

### LINKED ISSUE #1174: Front page needs removal for migration
Migration requires the site has no content. Front/landing page is considered content and must be manually deleted before you can run migration.

Can we/should we...

1. Have an easy "build landing page" page tool?
2. Find a way to avoid having to delete the landing page?
3. Other?

m4olivei:
Resolves #1174

This adds an option to the Migration credentials form for "Create default landing page". Off by default, if checked, at the end of a successful migration it will create a new default landing page and set it as the homepage. We also refactor the code that creates the landing page out of `mukurtu.install` and into a service in `mukurtu_landing_page` module, so that it can be used consistently both when installing Mukurtu from fresh, and when finished the migration.

In addition, if the "Create default landing page" option is not checked, we at least set the front page to `/node`, which is the Drupal stadard install profiled default, and will at least prevent the error referenced in https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1174#issuecomment-3543379333. The error stems from the front page configuration remaining set at `/node/1`, which is what the Mukurtu install profile sets, which at the time of install is the created default landing page. However, immediatly post-migration, `/node/1` could be anything that the migration created, which can cause errors, or would otherwise make no sense as a homepage. If the migration does set `/node` as the front page, a message is shown to the user to encourage them to set it to something more appropriate.

## Steps to Test

- [ ] `ddev drush si mukurtu --site-name=Mukurtu --yes`
- [ ] Verify that the default landing page is created and set to the home page of the site, as per the usual.
- [ ] Login and delete any existing content
- [ ] Go to `/admin/migrate`
- [ ] When at the credentials’ screen, set the "Create default landing page" toggle.
- [ ] Run the migration
- [ ] At the completion of the migration, you should recieve a message saying the landing page was created successfully
- [ ] Verify that the default landing page was created and set to the home page of the site
- [ ] Re-run the migration without the toggle set
- [ ] Verify that no content was created
- [ ] Verify that on visiting the home page, you don't get a PHP error as described in https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1174#issuecomment-3543379333


michael-wynne-wsu (approved):
Looks perfect to me, please merge as you see fit!

nick-deer (approved):
Looks good to me too!

PR #1196, opened 2025-11-21
899: Add a primary key for the mukurtu_protocol_access table

### LINKED ISSUE #899: Error: Transaction isolation level
When I load a new local iteration, I get this error message. I got it periodically last week and with each local site I built today. 

Image

Below are my notes from last week:
    _Error message: Transaction isolation level READ COMMITTED For this to work correctly, all tables must have a primary key. The following table(s) do not have a primary key: mukurtu_protocol_access. See the setting MySQL transaction isolation level page for more information 

    Worked with Brian on this; apparently it is a much more involved development issue.  

    This error happened 4Jun2025 in both iterations I built: mukurtujun and mukurtuinstalltest. It did NOT happen on 5Jun2025 in the mukurtu5jun iteration. Brian said it might be an intermittent error and to keep banging away at it to see if it happens again. Keeping the two broken iterations to compare if the error happens again. 

    This popped up again on 9JUN. Brought it up to Michael._

m4olivei:
Resolves #899

Adds a primary key to the `mukurtu_protocol_access` table in `hook_schema`. Models the key after the key used in the `node_access` table.

### DIFF
--- a/modules/mukurtu_protocol/mukurtu_protocol.install
+++ b/modules/mukurtu_protocol/mukurtu_protocol.install
@@ -90,6 +90,7 @@ function mukurtu_protocol_schema() {
         'unsigned' => TRUE,
       ],
     ],
+    'primary key' => ['id', 'entity_type_id', 'protocol_set_id', 'langcode'],
   ];
 
   return $schema;


michael-wynne-wsu (approved):
Looks good to me, thank you!

nick-deer (approved):
The error is gone for me!

michael-wynne-wsu:
I think you can merge whenever you're back at it, no need to wait for all three reviewers.

PR #1212, opened 2025-12-02
922: Allow Community Record Stewards to manage Community Records

### LINKED ISSUE #922: Community Record permissions
From Nick:

To create/edit/delete a CR, one must first have permission to edit the content type that the CR is. Right now the community record steward should be able to create/edit/delete community records, but they don’t have permission to edit ANY content types, so they can’t make CRs. What do we do? 

Since CR can be of any content type, we can’t just grant the community record steward edit rights to a particular type like DH since it wouldn’t reflect if the user chose to enable CRs on more content types.  

Could grant permissions to edit all content types, but that doesn’t sound in the spirit of our permissions and roles.

m4olivei:
Resolves #922 

### Description

Makes use of a handy OG content entity permissions event to incorporate logic allowing Community Record Stewards to manage Community Records.

### Testing Instructions

**Basic case**

- [ ] Create two Community, Protocol pairs
- [ ] Create a new user, and give them the "Community Record Steward" role on one of the protocols
- [ ] Create a new Media Asset for the first Community, Protocol pair
- [ ] Create a new Media Asset for the second Community, Protocol pair
- [ ] Create a new DH Item for the first Community, Protocol pair, adding the Media Asset from the last step
- [ ] Create a new DH Item for the second Community, Protocol pair, adding the Media Asset from the last step
- [ ] Create a new CR for the second DH Item
- [ ] Login is the "Community Record Steward" created earlier
- [ ] View the DH Item created by the admin in the previous step for the first Community, Protocol pair
- [ ] Create a Community Record for this item
- [ ] After creation, you should be able to Edit and Delete the DH Community record as well as the Community Record Steward user
- [ ] View the CR for the second DH Item created by the admin in the previous step
- [ ] The users should not be able to edit or delete it

**Other scenarios**

- [ ] Feel free to stress test other scenarios around this use case to try and break things

### DIFF
--- a/modules/mukurtu_community_records/src/EventSubscriber/MukurtuCommunityRecordEventSubscriber.php
+++ b/modules/mukurtu_community_records/src/EventSubscriber/MukurtuCommunityRecordEventSubscriber.php
@@ -34,8 +34,12 @@ class MukurtuCommunityRecordEventSubscriber implements EventSubscriberInterface 
    * @param \Drupal\og\OgAccessInterface $ogAccess
    *   The OG access service.
    */
-  public function __construct(protected PermissionManagerInterface $permissionManager, protected EntityTypeManagerInterface $entityTypeManager, protected EntityTypeBundleInfoInterface $entityTypeBundleInfo, protected OgAccessInterface $ogAccess) {
-  }
+  public function __construct(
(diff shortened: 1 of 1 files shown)


nick-deer (approved):
Everything looks good to me!

Issue #1088, opened 2025-09-12
Gin theme not used in group bulk management forms


michael-wynne-wsu:
In a community or protocol, if you manage members, select one or more members, and apply a bulk action, the following forms are not using Gin theme. Unsure where this setting is.

eg: /group/roles/add?destination=/admin/communities/2/members&group=2&og_membership_type=default&entity_type_id=community

<img width="1844" height="579" alt="Image" src="https://github.com/user-attachments/assets/ef920fce-6ede-49a3-af9b-1fa54bbef3a3" />

<img width="1847" height="909" alt="Image" src="https://github.com/user-attachments/assets/751d7e9d-5191-48c4-89ac-914fd68fd729" />

nick-deer:
@michael-wynne-wsu yeah that route and that form are supplied by OG. We could try overriding it with a hook, but I've never tried changing a form's theme this way so I'm not sure if it's possible. If you wanted to try it, the hook you'd need to override would be `mukurtu_protocol_form_og_membership_add_multiple_roles_action_alter(&$form, FormStateInterface $form_state)`

michael-wynne-wsu:
@nick-deer We could also use this module...

https://www.drupal.org/project/admin_theme

It feels a little hacky, but in the interest of get it done, what do you think? Would need to figure out how to pre-populate relevant paths.

<img width="1851" height="609" alt="Image" src="https://github.com/user-attachments/assets/90e7af03-bc98-4cc2-97be-2b5d11119166" />

<img width="1812" height="1015" alt="Image" src="https://github.com/user-attachments/assets/beffe655-7268-473f-89b1-c938c7c88781" />

nick-deer:
> [@nick-deer](https://github.com/nick-deer) We could also use this module...
> 
> https://www.drupal.org/project/admin_theme
> 
> It feels a little hacky, but in the interest of get it done, what do you think? Would need to figure out how to pre-populate relevant paths.

It looks like it works fine. I think it would be less hacky than trying to force a change of theme in a form hook.

nick-deer:
@michael-wynne-wsu Let's make a new issue for it

PR #1220, opened 2025-12-04
Adding admin_theme module

michael-wynne-wsu:
Addresses #1088 

Only added it for the one form, so we may need to add more paths in the future.

To test:

- [ ] Manage a community or protocol and use the "add roles" bulk operation on members.


christa-kavanagh-wsu (commented):
This appears with the proper theme throughout the process.

nick-deer (approved):
Works for me!

Issue #1221, opened 2025-12-04
Change role select form in bulk operations


michael-wynne-wsu:
At `/group/roles/add`, change from a multi select field to checkboxes.
This will affect community and protocol bulk membership management.

<img width="1851" height="609" alt="Image" src="https://github.com/user-attachments/assets/30555fea-085d-40a6-8d51-5d2df8168456" />

PR #1222, opened 2025-12-04
1221: Change role select form in bulk operations

nick-deer:
Addresses https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1221

Steps to test:

- [ ] Create a community
- [ ] Go to manage members in that community
- [ ] Select a community member and apply the 'Add roles to selected membership(s)' bulk action
- [ ] Verify that the roles options display as checkboxes instead of multiselect.

### DIFF
--- a/modules/mukurtu_protocol/mukurtu_protocol.module
+++ b/modules/mukurtu_protocol/mukurtu_protocol.module
@@ -735,3 +735,17 @@ function add_protocol_member_form_validate(&$form, FormStateInterface $form_stat
     }
   }
 }
+
+/**
+ * Implements hook_form_form_id_alter().
+ *
+ * Change the roles select from multiselect to checkboxes.
+ */
+function mukurtu_protocol_form_og_membership_add_multiple_roles_action_form_alter(&$form, FormStateInterface $form_state) {
+  $rolesElement = $form['roles'];
+
+  $form['roles'] = [
+    '#type' => 'checkboxes',
+    '#options' => $rolesElement['#options']
+  ];
+}


michael-wynne-wsu (approved):
That is perfect, thank you!

joelle-birano-wsu:
The checkboxes look great! I did notice something else though - after applying one role and saving, the other two roles are applied. This happens no matter which role I apply. 

Screenshot 2025-12-04 at 4 23 00 PM

Screenshot 2025-12-04 at 4 20 05 PM
Screenshot 2025-12-04 at 4 21 06 PM

nick-deer:
@michael-wynne-wsu @joelle-birano-wsu upon further inspection, this looks more complicated than I thought it would be initially. Yes I can change the multiselect widget to a checkbox widget, however, that doesn't change how the selected roles are submitted and handled by OG, which expects multiselect values coming in. To do this properly, it looks like I'd have to extend OG quite a bit by making a custom form with custom submit handling and a custom action just to bypass OG. I can look into it more if you want, but I just wanted to let you know it will be an undertaking.

michael-wynne-wsu:
@nick-deer thank you so much for trying to track that down. No, I don't think it's worth your time right now to continue pursuing this, but we can keep the issue for future work.

Annoying because I think in v3 is IS a checkbox on the equivalent form!

michael-wynne-wsu:

Link:
- https://pr1222-u59mvyyvhgmwt0ff7lzwcsxtt8zlvq2y.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/693201f06b076f468b690370

Issue #873, opened 2025-05-29
Community record and multi-page item editing confusion


christa-kavanagh-wsu:
From Michael:
> Local tasks (view, edit, etc) reflect the main node, not the active record. Because what’s happening with the tabs is that they don’t actually take you to a new page, they just show/hide content on the main node. So the local tasks buttons aren’t updating.
>
>Effectively the only way to edit a CR or MPI is to access that node directly from search or an admin view. Need to add some kind of in-page way to edit a node "contained" in a DH item.
>
> Additionally, on save, would be great to be directed back to the edited node/tab, not the base item.

This happens when editing community records:
When I go into an item, select a tab, and select Edit it only lets me edit the digital heritage item page I started from. In v3 you could edit the record associated with whichever tab you selected.

So if I choose Pangur Ban as the DH item, it only lets me edit that record. I choose The scholar and his cat as the DH item, it only lets me edit that record. This happens no matter which tab I select to edit. 

![Image](https://github.com/user-attachments/assets/25632969-2554-4689-b12f-08db51946ec7)

![Image](https://github.com/user-attachments/assets/d40ffc03-5994-40ff-9427-d823c0b34eb7)

 (Correct)

![Image](https://github.com/user-attachments/assets/f5dca06c-465c-43c9-8802-eb48c5a8912f)

![Image](https://github.com/user-attachments/assets/05ad7cf9-edba-434c-abf0-9415f1fe2d90)

(Incorrect, should be The scholar and his cat)

PR #1249, opened 2025-12-10
873: Update the local tasks as you switch between CR

m4olivei:
Handles the CR side of #873 

This PR makes changes such that when logged in, and you have one or more CR's of content, as you switch between the CR tabs, the local task links will update to reflect the active tab.

## Testing instructions

- [ ] Create content and one or more CR's of that content
- [ ] As an admin, when viewing either the original item, or the CR, click between them using the CR tabs
- [ ] Verify that the local task links (eg. Edit, Delete, etc.) update to reflect the currently active tab

NOTE: Multi-page Items will be handled in a separate pull request.

### DIFF
--- a/modules/mukurtu_community_records/js/mukurtu-community-records.js
+++ b/modules/mukurtu_community_records/js/mukurtu-community-records.js
@@ -87,7 +87,7 @@
     toggleCrTask: function(isCr) {
       const localTasksElements = document.querySelectorAll('.local-tasks');
       localTasksElements.forEach((localTasks) => {
-        localTasks.querySelectorAll('li:has([href*=community-record\\/add])')
+        localTasks.querySelectorAll('li:has([href*=community-record\\/add]), li:has([href*=new-multipage])')
           .forEach((element) => {
             if (isCr) {
               element.setAttribute('hidden', '');


michael-wynne-wsu:

Link:
- https://pr1249-hpiyxzofydho3t3ynjxmdpsldn5tq4ps.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6939ea385408d4dcd0c13520

m4olivei:
> A little vertical space between the local tasks and the community record tabs would be nice (very low priority).

Fixed in recent commit.

> In previous versions a CR could only be created from a base item, not from a CR. I don't actually think it's a permissions issue here, just that seeing the "add community record" button on an existing CR may lead users to think they're doing that, and not adding another to the larger/original item.

Right, so when you click "Add Community Record" while viewing a CR, just like the "+ Community Record" local action before, the node form you're taken to is set to create a new CR with reference to the original content, not the CR. That prevents a user from creating a CR of a CR. I've also taken steps to dynamically show/hide the local task based on if the item in view is a CR or not.

Fixed in recent commit.

> "Create multi-page item" shouldn't show on CRs. If you try to do it, it does give an error message (which is correct), but not displaying it would be better.

Fixed in recent commit.

michael-wynne-wsu:

Link:
- https://pr1249-hpiyxzofydho3t3ynjxmdpsldn5tq4ps.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6939ea385408d4dcd0c13520

michael-wynne-wsu:

Link:
- https://pr1249-hpiyxzofydho3t3ynjxmdpsldn5tq4ps.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6939ea385408d4dcd0c13520

PR #1277, opened 2025-12-16
873: Multi-page item local task adjustment

### LINKED ISSUE #873: Community record and multi-page item editing confusion

m4olivei:
Resolves #873 

### Description

* Adjusts the label for "Edit MPI"
* Adds a task for "Edit Page"
* Updates the links dynamically when switching between each item in a multipage items entity

### Testing instructions

* Create some content
* Create a Multipage items entity from that content
* View the items of the Multipage items
* Verify the local tasks are working as expected on initial page load, and as you switch between items

### DIFF
--- a/modules/mukurtu_multipage_items/src/Hook/HideExtraLocalTasksFromNodeEditRoute.php
+++ b/modules/mukurtu_multipage_items/src/Hook/HideExtraLocalTasksFromNodeEditRoute.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\mukurtu_multipage_items\Hook;
+
+use Drupal\Core\Hook\Attribute\Hook;
+
+/**
+ * Hide extra local tasks from node edit route.
+ */
+class HideExtraLocalTasksFromNodeEditRoute {
+
+  /**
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu (approved):
Yep that sorted it out, thanks!

PR #1289, opened 2025-12-17
1256: Mult-page grab bag of issues

### LINKED ISSUE #1256: Multi-page item updates
- [x] Once an MPI is created, even if a user can only view one page, they still see the whole carousel (eg: one public page and then other private pages). Likely unavoidable, and LOW priority.
   - [x] Currently not showing this behaviour (dec 16)
- [x] Hide the carousel and jump menu before additional pages are added.
- [x] I have tested mixed public/private pages, and all works as expected so far.
- [x] Person records do not display Name field anywhere. 
- [x] Dictionary word thumbnail missing in splide.
   - [x] If there is a populated thumbnail field, use that image. If not, use the default thumbnail as with DH audio.
- [x] Can't remove pages from edit form.
- [ ] PDF not using generated thumbnails in splide, but that's blocked by #936 
- [ ] Uploaded video not using generated thumbnail in splide.
- [x] Soundcloud player embed doesn't seem to be working, this seems new.
- [x] You can add one DH item to more than one MPI. This should NOT be allowed.
   - [x] This doesn't become a problem until you try to edit and save either MPI that the item is in; you are unable to edit and save. You can't delete the item from the MPI when you are in the MPI's edit tab.
- [x] There was an error in navigating the Select Content modal when trying to select content for MPIs: you can't navigate to the second page if the modal has more that one page. It returns an undesignated browser error.

m4olivei:
Resolves #1256 

### Description

This is a potpourii of fixes for multi-page items.

### Testing instructions

- [ ] Reference the issue description for the list of issues that are aiming to be fixed here.
- [ ] Verify that all issues are fixed.

### DIFF
--- a/themes/mukurtu_v4/templates/content/node--person--full.html.twig
+++ b/themes/mukurtu_v4/templates/content/node--person--full.html.twig
@@ -120,7 +120,7 @@
               {{ content.community_protocol_list }}
               {{ content.group_metadata|without('flag_export_content') }}
               <div class="field">
-                <div class="field__label">{{ "Title"|t }}</div>
+                <div class="field__label">{{ "Name"|t }}</div>
                 <div class="field__item">{{ node.label() }}</div>
               </div>
             </div>


modules/mukurtu_multipage_items/src/Plugin/Validation/Constraint/MultipageValidNodeConstraintValidator.php
```diff
@@ -1,125 +1,190 @@
...
+      if (!in_array($target_id, $unique_items)) {
+        $unique_items[] = $target_id;
       }
       // Item is a duplicate.
-      elseif (in_array($id, $uniqueItems)) {
-        $this->context->addViolation($constraint->isDuplicate, ['%value' => $this->getTitle($id)]);
+      else {
+        $this->context->buildViolation($constraint->isDuplicate)
+          ->setParameter('%value', $this->getTitle($target_id))
+          ->atPath((string) $delta . '.target_id')
```
m4olivei: @nick-deer I tagged you on this PR, b/c I nearly completely re-wrote your validator logic 😬. This line of code was one of the main culprits. Drupal wants you to be as specific as possible in targeting the violation. The constraint is set at the field level, and this now sets the violation at the field item property level. Prior the violation was set at the field level, which for some reason causes issues. I can't exactly say why it does, but I've been bitten a couple times by this, and I now just follow this principal. `\Drupal\Core\Entity\Plugin\Validation\Constraint\ValidReferenceConstraintValidator` is a great example that I'm following here.

nick-deer: That makes sense, thanks!

michael-wynne-wsu (changes requested):
I believe fairly minor adjustments (and I'm happy for those to be changed without review):
1. Person record: "Title" should be labelled "Name" instead
2. Video media asset should use the actual video thumbnail, not the generic one where available.

Image

And then some things that I think we can move from this issue to a different milestone:
- PDF thumbnails, as noted.
- Not allowing users to select invalid content types as pages (this is a broader issue, already logged)
- Not allowing users to select invalid content as pages (eg: not listing a piece of content that is already a page of another MPI, instead of allowing selection but throwing an error).

michael-wynne-wsu:

Link:
- https://pr1289-rybv9ls0cqah0x8o0unzrvlmbbsuw0dw.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6942dd21e77ea5fafb46cb7e

michael-wynne-wsu (approved):

PR #1303, opened 2025-12-30
228: Implement ability to manually sort the glossary facet for the dictionary

### LINKED ISSUE #228: Control dictionary glossary order
At `/dictionary` there is a "glossary" facet allowing users to filter search results by the first letter of the word. 

- [ ] The basic display of the facet is currently active.
- [ ] It references the "glossary entry" field in dictionary words.
   - [ ] If a user does not populate that field, on save the first letter of the "term" field is saved.
   - [ ] Users can also provide values in this field. Important to note that it might be multiple characters (eg: ch), or accented characters, or something else).

We need to add a site-wide control (Mukurtu Manager role) for setting the glossary order.

- [ ] It should have a few sort options.
   - [ ] A "default" sort being unicode order (basically your "regular" alphabet order).
   - [ ] Possibly a "transliterated" sort (per @m4olivei suggestion).
   - [ ] A user-defined sort where they can control the entire order. This could be based on an above sort or they could be required to enter all entries, that's fine.
- [ ] There should be an option to display/hide unused letters. Eg: if my alphabet is the same as English but X isn't used, since there are no words starting with x, it doesn't need to be shown in the facet.

Screenshots from v3:

Image

Image

m4olivei:
Resolves #228 

### Description

Implements the ability to control the dictionary glossary order.

### Testing Instructions

- [ ] Add some Dictionary Word nodes with and without customized Glossary Entry values
- [ ] Go to `/dictionary`. Should see default unicode ordering for the glossary entries
- [ ] Go to `/admin/config/mukurtu/dictionary-glossary-order`
- [ ] Set the sort to "User-defined sort"
- [ ] Manually set weights for known glossary entry values
- [ ] Go to `/dictionary`. Should see your configured weights reflected in the display of the glossary entries

NOTE: The default behavior of facets is to only show values that exist in the system. If we want to give admins the option to flesh out entries to fill the alphabet with empty links, let’s do that as a follow-up. It feels like that would just present noise in the UX, and we're better off not having that option.

### DIFF
--- a/.tugboat/config.yml
+++ b/.tugboat/config.yml
@@ -8,13 +8,13 @@ services:
       init:
         # Install dependencies for PHP extensions.
         # Add poppler-utils for the pdftotext executable.
-        - apt update && apt install -y libzip-dev libpng-dev libjpeg-dev libwebp-dev libfreetype6-dev poppler-utils
+        - apt update && apt install -y libzip-dev libpng-dev libjpeg-dev libwebp-dev libfreetype6-dev libicu-dev poppler-utils
 
         # Add JPEG and FreeType support to GD.
         - docker-php-ext-configure gd --with-jpeg --with-webp --with-freetype
 
         # Install the PHP extensions all at once to minimize rebuilding.
-        - docker-php-ext-install zip opcache gd
+        - docker-php-ext-install zip opcache gd intl
 
         # Enable Apache 'rewrite' module for clean URLs.
         - a2enmod rewrite
--- a/modules/mukurtu_dictionary/src/Form/MukurtuDictionaryGlossaryOrderForm.php
+++ b/modules/mukurtu_dictionary/src/Form/MukurtuDictionaryGlossaryOrderForm.php
@@ -4,7 +4,6 @@ declare(strict_types = 1);
 
 namespace Drupal\mukurtu_dictionary\Form;
 
-use Collator;
 use Drupal\Core\Form\ConfigFormBase;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
@@ -111,10 +110,19 @@ class MukurtuDictionaryGlossaryOrderForm extends ConfigFormBase {
    * Collator accepts ISO 639 language codes and automatically uses appropriate
    * locale-specific sorting rules.
    *
+   * Falls back to asort() if the intl extension is not available.
+   *
    * @param array &$array
    *   The array to sort (passed by reference).
    */
   protected function sortByLocale(array &$array): void {
+    // Check if the intl extension is available.
+    if (!class_exists('\Collator')) {
+      // Fallback to simple alphabetical sort if intl is not available.
+      asort($array);
+      return;
+    }
+
     // Get the current language code from Drupal.
     $current_language = $this->languageManager->getCurrentLanguage();
     $langcode = $current_language->getId();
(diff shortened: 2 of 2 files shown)


michael-wynne-wsu:

Link:
- https://pr1303-iq6mw1mq1ncsi9oxnt1nmo6zk9rcm1yl.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6953f393e84840977f54bc54

michael-wynne-wsu:

Link:
- https://pr1303-iq6mw1mq1ncsi9oxnt1nmo6zk9rcm1yl.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6953f393e84840977f54bc54

michael-wynne-wsu:

Link:
- https://pr1303-iq6mw1mq1ncsi9oxnt1nmo6zk9rcm1yl.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6953f393e84840977f54bc54

michael-wynne-wsu:

Link:
- https://pr1303-iq6mw1mq1ncsi9oxnt1nmo6zk9rcm1yl.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6953f393e84840977f54bc54

michael-wynne-wsu:

Link:
- https://pr1303-iq6mw1mq1ncsi9oxnt1nmo6zk9rcm1yl.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6953f393e84840977f54bc54

m4olivei:
Gah, hang on, not there just yet. While the facets are showing distinctly now, the filtering is not correct yet :/

m4olivei:
@nick-deer @michael-wynne-wsu This is ready for re-review. As I noted in my last comment, we're somewhat restrained here in that the facets will now be very verbose by default, and will expose all of the following if there are `field_glossary_entry` values for each one:

* e
* E
* é
* É

This can be mitigated if desired by the user, by editors adopting a convention of using either upper or lower case consistently if they want exposed facets to collapse down, for instance if they only want uppercase to show in the facets, they would set `field_glossary_entry` to either "E" or "É", where the term itself can be lowercase. Although it is perhpas desireable to have that level of verbosity, if for instance an editorial team adopts a style guide that reserves uppercase for proper nouns, and they actually want all 4 facets. Let me know what you think.

m4olivei:
> @m4olivei That's fantastic. I see the potential issues with that, but I think we can work with it just as you described. I suspect it will be rare for it to cause any difficulty.
> 
> And I see the facet filtering is also working, great!
> 
> Hopefully one last note... Regardless of user/default glossary sort, the results are in a different order from that setting, and I would expect that to match. I actually can't really tell what the order of results are keyed to.
> 
> Screenshot 2026-01-08 at 1 25 21 PM

Thanks @michael-wynne-wsu. I've updated sorting to match the manually set Dictionary Glossary Order when applicable. That was tricky to do given the configuration is removed from the actual data and we need it to be part of the views sort, which is a Search API based view. AI was a huge help though, and I think we came up with a good solution for that here. 

I also added a "Reset Order to Alphabetical" action on the "Dictionary Glossary Order" form (`/admin/config/mukurtu/dictionary-glossary-order`). This can be useful if a user has set a manual order for things, but then later added a number of words which end up floating to the end of the previously manually sorted glossary entries. This lets a user start from a basis of alphabetical again, and re-arrange from there. This might be only useful for testing scenarios like we find ourselves in now, but maybe it will be more broadly useful:

image

PR #1304, opened 2025-12-30
g911: Visual tweaks to dictionary word front-end

### LINKED ISSUE #911: Dictionary word display
- [ ] The "term" field within word entries is not displayed. They should use the same "call out" styling as the main "term" field.
- [ ] When a word includes word entries, especially if there are several and/or any of them are quite long, the page gets very long.
   - [ ] One idea is to apply a tabbed view between words entries, similar to community records, but we are open to other ideas.
- [ ] In general, the page gets pretty long, and identifying the fields can be a bit hard. Maybe adjust spacing or field labels? Could do two-column content? Also see v3 screenshot for reference.
- [ ] Map field seems to seem some top margin/padding.
- [ ] Need a clearer separator between "Other Entries" (word entries) and "Additional Information".
   - [ ] Could consider moving additional information fields to sidebar?
- [x] Vertical spacing between field titles and text looks good.  
- [x] All paragraphs display, even when empty. See https://github.com/MukurtuCMS/Mukurtu-CMS/issues/572
- [x] Map displays even when empty. See https://github.com/MukurtuCMS/Mukurtu-CMS/issues/889

Image

Image

And including a v3 screenshot for reference:

Image

m4olivei:
Resolves #911 

## Description

I took a first pass at some fixes for the styling here, drawing on existing styles, to add separation and space where called out in the issue. The issue description wasn't clear on definitive implementation details, so I'm starting simple here, seeing whether this is enough to move us forward. If we want to go further and wade into some suggestions made in the issue description (eg. moving fields to the sidebar, tabbed interface, etc.), I'd like to suggest that we discuss those, decide on a direction and then implement.

Here I've done the following:

* Show the "term" field for additional word entries in the "call out" style
* Add clearer separation for "Other Entries" and "Additional Information"
* Fixe styles for fields under "Additional Information" to make the spacing and styles consistent

## Testing instructions

- [ ] Create a fully fleshed out Dictionary Word, with one or more word entries, additional fields, and related content
- [ ] View the Dictionary Word
- [ ] Verify that we're at least in a better place, addressing the items in the issue description, without exploring the as yet undecided suggestions

### DIFF
--- a/themes/mukurtu_v4/templates/content/node--person--browse.html.twig
+++ b/themes/mukurtu_v4/templates/content/node--person--browse.html.twig
@@ -86,7 +86,7 @@
     <div class="horizontal-card__content-header">
       {{ title_prefix }}
       <h2{{ title_attributes }}>
-        <a href="{{ url }}" rel="bookmark">{{ label }}</a>
+        <a href="{{ url }}" rel="bookmark">{{ label.0 }}</a>
       </h2>
       {{ title_suffix }}
     </div>


michael-wynne-wsu:

Link:
- https://pr1304-lthzdmfxbgpkjircfrrcq2fvdzhyefiv.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69540c80bd62bc3438c3b42a

michael-wynne-wsu (approved):
Thank you @m4olivei this absolutely addresses the well defined issues and I'm very happy to have it merged in. 

At this point we could certainly call it stable beta complete. Or if you have ideas (I'll mull on this too), the broad issue of layout/good use of space remains (since a lot of these fields contain pretty short text strings in reality). Which might just benefit from a conversation as well.

PR #1312, opened 2026-01-09
656 : Remove sort widget for collections view

### LINKED ISSUE #656: Remove sort widget on collections browse page
~~It's a bit weird where it is. Pauline also mentioned that she had issues making things look/not look like buttons.~~

/collections

For now I think we should just remove it. This display is odd and the available options aren't that useful.

m4olivei:
Resolves #656 

## Steps to test

- [ ] Create a Community, Protocol, DH Items and Collections of those DH Items
- [ ] Browse to `/collections`
- [ ] Verify the sort widget is gone and that sorting is based on the collection title in ascending order


michael-wynne-wsu (approved):
Yep that's spot on, thank you!

PR #1313, opened 2026-01-09
259: Collection management adjustments 

### LINKED ISSUE #259: Manage Collection Organization
Needs some work.

Adding new subcollection from view is good. Would like to add from "Manage Collection Organization" tab as well.

Shouldn't be able to select an already-added collection (it doesn't let you add it, but it's select-able)

Can the selection be a dropdown instead of autocomplete only?

Needs more direction about placement. Not clear that you remove a collection by moving it all the way to the left.

m4olivei:
Resolves #259 

Fixes various administrative issues with Collections.

### Testing instructions

- [x] Create a Community / Protocol pair
- [x] Create a handful of DH Items (or other content)
- [x] Create Collections from the DH Items
- [x] View a Collection and verify the "New sub-collection" action has moved to local tasks
- [x] Edit a Collection, and verify the "New sub-collecction" local task is present and functioning from this context
- [x] Navigate to Manage Collection Organization. There should be the following improvements:

* If you drag a sub-collection all the way left at the same level as the root collection, and try to Save, you'll get an error, and need to correct this.
* If you drag the root collection below some other item, you'll get an error
* You can remove parts of the hierarchy. You need to Save to persist this.
* The auto-complete will not return items that are already present in the hierarchy
* "Add to Organization" button is updated to "Add to Collection"

### DIFF
--- a/modules/mukurtu_collection/src/Controller/CollectionOrganizationController.php
+++ b/modules/mukurtu_collection/src/Controller/CollectionOrganizationController.php
@@ -74,13 +74,20 @@ class CollectionOrganizationController extends ControllerBase {
    */
   public function handleAutocomplete(NodeInterface $node, Request $request) {
     $results = [];
+    if (!$node instanceof CollectionInterface) {
+      return new JsonResponse($results);
+    }
     $input = $request->query->get('q');
     if (strlen($input) > 0) {
       $query = \Drupal::entityQuery('node')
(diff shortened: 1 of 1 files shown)


m4olivei:
The only item remaining on this is to attempt to change the widget. That's non-trivial given this is a custom form. Will see whats possible.

PR #710, opened 2025-02-25
691: Fixing configuration of the 'search_api_db_mukurtu_browse_auto_index' Search API index to use fewer full text fields.

quicksketch:
Fixes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/691

This seems to fix installation issues with the `search_api_db_mukurtu_browse_auto_index` Search API index.

This index seemed to be configured almost exactly opposite the way it should be, with 16 fields that are UUIDs being "Fulltext" instead of "String", and the Taxonomy label being "String" instead of "Fulltext". This PR brings this index in line with the other indexes, using strings for UUIDs. By reducing the amount of data in a table, this seems to resolve the installation issues.

PR #1320, opened 2026-01-21
1308 : SAPI adjust taxonomy reference field type

### LINKED ISSUE #1308: Custom taxonomy migration
In trying to plan out custom migration documentation, we have started with one site that has multiple custom taxonomies, a fairly simple case.

I have run into the following (documented in more detail in follow-up comments):

- One through three taxonomies, works as expected!
- Four taxonomies, install errors related to search API, but the site builds.
- Certain taxonomies are not migrating properly.
   - This SEEMS to happen when the taxonomy name and machine name in the v3 site don't match. Aligning them seems to resolve it.

This is the migration process I am following: https://github.com/MukurtuCMS/Mukurtu-CMS/wiki/DDEV-for-local-migration-and-development-testing

These are the components of custom taxonomy migration: https://github.com/MukurtuCMS/Mukurtu-CMS/wiki/Migrating-custom-taxonomies

This is the complete migration as planned, with smaller incremental branches available as well: https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1311

m4olivei:
Resolves #1308 

Uses the `string` data type for all taxonomy reference fields in the `mukurtu_browse_auto_index` index. While this was the intent of https://github.com/MukurtuCMS/Mukurtu-CMS/pull/710 , that PR made the change to the installed config. There remain scenarios where the automatic creation / edit of fields in the `mukurtu_browse_auto_index` index can essentially revert that initial config state.

Also taking the opportunity here to clean up the `TaxonomyRecordViewController` class from a code style perspective as well as to consolidate logic that was spread out into `mukurtu_taxonomy_views_pre_view`.

## Steps to test

- [ ] Run through the test in the issue summary
- [ ] Verify that we don't get those same errors anymore
- [ ] Navigate to the SAPI `mukurtu_browse_auto_index` index, eg. `/admin/config/search/search-api/index/mukurtu_browse_auto_index`
- [ ] Verify that the index is indexed, if not, click Index Now towards the bottom of the form.
- [ ] Visit a taxonomy page, either by clicking a referenced taxonomy term in content, eg. DH item, or by navigating to Structure > Taxonomy, then clicking "List Terms" next to a vocabulary with terms, and content tagged with those terms. Then click the label of a taxonomy term. 
- [ ] Verify that the page renders as expected

### DIFF
--- a/modules/mukurtu_taxonomy/src/EventSubscriber/TaxonomyFieldSearchIndexSubscriber.php
+++ b/modules/mukurtu_taxonomy/src/EventSubscriber/TaxonomyFieldSearchIndexSubscriber.php
@@ -44,7 +44,7 @@ class TaxonomyFieldSearchIndexSubscriber implements EventSubscriberInterface {
       $field_id = "{$event->entity_type_id}__{$field_name}__uuid";
       $property_path = "{$field_name}:entity:uuid";
       $label = "{$event->field_definition->getLabel()} » Taxonomy term » UUID";
-      $event->indexField($indexes, $field_id, $property_path, $label);
+      $event->indexField($indexes, $field_id, $property_path, $label, 'string');
     }
   }
 


m4olivei:
Hum, something is still off here with the taxonomy term pages. Reviewing...

PR #1330, opened 2026-01-29
723:  Restore full set of dictionary facets

### LINKED ISSUE #723: Browse pages missing search box
/browse and /digital-heritage should have a search/sort/reset box at the top of the right column. Unsure exactly which release lost this but it was recent.

Image

m4olivei:
Resolves #723 

## Testing instructions

This PR restores the full set of facets as well as the search box on the `/dictionary` page, as well as touching up styles. We're also making use of a proper form `<label>` element here for accessibility.

- [ ] Create a bunch of full features Dictionary Words (with keyword, word type, etc.) 
- [ ] Add a couple of Word Lists
- [ ] Navigate to `/dictionary`
- [ ] Verify all the facets are present and working correctly
- [ ] Verify the keyword search is present and works properly


m4olivei:
> I'm not opposed to moving this page up the priority list if you want to stick with this train of focus though.

Thats OK, lets shift to Roundtrip and either leave this for a palate cleanser, or come back to it after we're through with Roundtrip.

PR #1372, opened 2026-02-13
1359: Import should support all allowed file extensions

m4olivei:
Resolves #1359 

### Description

This will aggregate all of the allowed extensions across all media bundles and allow any of those extensions to be uploaded as media to the importer.

### Testing instructions

- [ ] Visit `/admin/import`
- [ ] Attempt to upload a file with an extension outside of this default allowed set (what was happening): `jpg jpeg gif png txt doc xls pdf ppt pps odt ods od`
- [ ] You should be able to upload any file type that any media bundle allows. I used AI to quickly aggregate those, they are: `png gif jpg jpeg webp mp3 wav aac m4a ogg mp4 webm ogv txt rtf doc docx ppt pptx xls xlsx pdf odf odg odp ods odt fodt fods fodp fodg key numbers pages csv sxw zip rar gz 7z ta`
- [ ] Verify that the file is updloaded to the field successfully without error

PR #1373, opened 2026-02-13
1360: Avoid setting a default for field_thumbnail on Video and RemoteVideo

### LINKED ISSUE #1360: Remote media shouldn't require a thumbnail on import
Remote video (youtube, vimeo) and soundcloud are requiring a thumbnail on import. Ideally the system would draw that from the host site the same way it does when created through the non-roundtrip workflow. 

Image

m4olivei:
Resolves #1360 

### Description

The error message noted in the issue stemmed from the `field_thumbnail` fielld having a default value set in its definition to an empty string. It is supposed to be an integer. This unsets the default to avoid issues with validation, especially when it's not even mapped in the import in the first place.

### Testing instructions

- [ ] Import media with at least Name, Cultural Protocols > Protocols, Cultural Protocols > Sharing Setting, and Video URL (or the relevant file field for other media types)
- [ ] Verify that you no longer get the error reported and the import succeeds
- [ ] Test with all types of media


michael-wynne-wsu:
Need to wait for #1372  to be merged before I can test, right?

Issue #254, opened 2024-02-19
Need autogenerated thumbnails


michael-wynne-wsu:
- [x] Images - yes, it works.
- [x] Audio - no thumbnail needed (will be a generic image or a player).
- [ ] File - at least required for PDFs, not present.
- [x] External embed - not required, manual upload only.
- [x] External video - yes, it works, but also the remote players work as expected.
- [x] Uploaded video - no, would be nice. But the player preview does work.
- [x] Soundcloud - no thumbnail needed, will use the SC player.

michael-wynne-wsu:
@jonscottwsu @brian-wsu @nick-deer Any chance we can look into PDF thumbnail generation? I'll do some exploring myself...

michael-wynne-wsu:
These both seem plausible:
- https://www.drupal.org/project/media_thumbnails_pdf 
- https://www.drupal.org/project/media_pdf_thumbnail

brian-wsu:
Media thumbnails (https://www.drupal.org/project/media_thumbnails_pdf) was easy to install and worked fine.

michael-wynne-wsu:
@bronzehedwick Is the media thumbnails module something you're up for incorporating while you work on the map previews as well? If not, I'll ask Nick to take a look next week.

bronzehedwick:
@michael-wynne-wsu I have the media thumbnails PDFs on my task list, so I'm up for doing that! Thanks for the direction.

quicksketch:
@bronzehedwick and I ran into some problems with PDF Previews. We tried using https://www.drupal.org/project/media_thumbnails_pdf, which seemed as though it would be straight-forward, but the module uses a not-well-maintained PECL extension. It hasn't had a [release that is compatible with PHP 8.3 or PHP 8.4 in years](https://github.com/Imagick/imagick/issues/640). I filed an [issue with the module to request it uses an alternative approach](https://www.drupal.org/project/media_thumbnails_pdf/issues/3515202) which would work on more environments.

Unfortunately the other module `media_pdf_thumbnails` (vs. `media_thumbnails_pdf`) also uses the same extension.

The PR at https://github.com/MukurtuCMS/Mukurtu-CMS/pull/754 solves the problem for Tugboat, and seemingly DDEV is fine too. But I don't think the ImageMagick PECL extension is available on Pantheon or Acquia (and probably other web hosts).

[...]

PR #1377, opened 2026-02-13
1362: Call prepareSave on entities that support it (media) in order to set a missing name

### LINKED ISSUE #1362: For importing NEW new media, is Name required?
- When working through the regular interface, the system automatically populates the required name field with the filename (for uploaded media) or the provided name (for remote media). 
- The only case where a name isn't auto populated should be external embeds.
- In the VAST majority of cases, people are happy to leave the name field as is. In rare cases they may want to change it.
- In roundtrip import, users must provide a name. In practice, that means they either waste time writing something (which is probably near-identical to the filename), OR just duplicating the filename entirely.
- So there seems to be very limited benefit to requiring the name field on new import (it may have more necessity on export/update import).
- Any reason we would not want to try and automate that so it more closely aligns with the regular create media asset workflow?

m4olivei:
Resolves #1362 

### Description

There is some handy logic within the Media entity code (`\Drupal\media\Entity\Media::prepareSave`) that will work out a default name for media when a name is not already set. As it was, we were validating the entity prior to save (and susequent `prepareSave` calls) such that the validation would fail before this default name logic could be applied. This corrects that. At the same time we take the opportunity to clean up the `\Drupal\mukurtu_import\Plugin\migrate\destination\ProtocolAwareEntityContent` migrate destination plugin code to tigten it up and ensure we're falling back as much as possible to the parent class.

NOTE: Especially for remote media, if possible, you should still map the Name field. If you do not, the media subsystem will make HTTP requests to oembed endpoints to fetch an appropriate name. Depending on how many remote videos you are attempting to import, this may become a performance issue, and break the import.

### Testing instructions

- [ ] Import media without mapping the name field
- [ ] Verify that a name is automatically set. In the case of Image, or other file based media, the filename should be used

### DIFF
--- a/modules/mukurtu_media/src/Entity/ExternalEmbed.php
+++ b/modules/mukurtu_media/src/Entity/ExternalEmbed.php
@@ -18,13 +18,15 @@ class ExternalEmbed extends Media implements ExternalEmbedInterface, CulturalPro
   use CulturalProtocolControlledTrait;
   use PeopleTrait;
 
-  public static function bundleFieldDefinitions(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions)
-  {
+  /**
+   * {@inheritdoc}
+   */
+  public static function bundleFieldDefinitions(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions) {
     $definitions = self::getProtocolFieldDefinitions();
 
     $definitions['field_media_external_embed'] = BaseFieldDefinition::create('text_long')
       ->setLabel('External Embed')
-      ->setDescription(t('Embed code from an external website. Note that while the media asset will be managed by cultural protocols, the originating website may not have similar privacy settings.	</br>External embeds are usually some kind of code wrapped in <iframe></iframe> tags.'))
+      ->setDescription(t('Embed code from an external website. Note that while the media asset will be managed by cultural protocols, the originating website may not have similar privacy settings. External embeds are usually some kind of code wrapped in &lt;iframe&gt;&lt;/iframe&gt; tags.'))
       ->setCardinality(1)
       ->setRequired(TRUE)
       ->setRevisionable(TRUE)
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu:

Link:
- https://pr1377-bwiopdajhwvgzk2y18aoggacxsoikhdv.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/698f62c4017fd55f32d65088

michael-wynne-wsu:

Link:
- https://pr1377-bwiopdajhwvgzk2y18aoggacxsoikhdv.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/698f62c4017fd55f32d65088

michael-wynne-wsu:
Okay, running into partial success, some issues. This spurred me to create an issue logging the TARGET behaviour of all media types and fields: https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1381 (happy to expand on that if needed).

So, in testing against that new document, and ONLY focusing on the required fields, the issues I am running into are....

**Image**

- Image > Alternative text is required (it should autofill from Image > File ID, or from Name, but Name autofills from Image > File ID).

**Audio**

- No issues, works as expected.

**Document**

- No issues, but thumbnail did not generate (that is not isolated to import, see https://github.com/MukurtuCMS/Mukurtu-CMS/issues/254)

**Video**

- No issues, works as expected.

**Remote Video**

- No issues, works as expected.
- It did work fine with no Name fields, but it's still recommended to provide those.

**Soundcloud**

- No issues, works as expected.
- NOTE: If no Name is provided, it just uses the system generated UUID.

**External Embed**

 - Working with required thumbnails exposes an issue here. It requires you to manage those thumbnails as independent image media assets, which I understand WHY, but it feels a bit counterintuitive here (as it would for managing any other thumbnails).

PR #1400, opened 2026-02-24
1394: Don't set default value on reference fields for collection, tighten up ValidCollectionItemValidator

### LINKED ISSUE #1394: Minimal collection import is failing
Just testing a barebones new collection import and it fails with the following message

> Collection-required-only.csv: [node]: field_child_collections.0.target_id=This value should be of the correct primitive type.||field_collection_image.0.target_id=This value should be of the correct primitive type.||field_items_in_collection=A collection cannot contain itself.||field_items_in_collection.0.target_id=This value should be of the correct primitive type.||field_keywords.0.target_id=This value should be of the correct primitive type.||field_related_content.0.target_id=This value should be of the correct primitive type.

Theoretically the only required fields should be collection name, and the two protocol fields (although in practice more will be used).

[Collection-required-only.csv](https://github.com/user-attachments/files/25424970/Collection-required-only.csv)

Image

m4olivei:
Resolves #1394 

## Description

Fixes validation constraint errors with importing collections. For most of the errors, what was happening was at a base field level, we had configured reference fields to have a default value of empty string, which doesn't make any sense. In one other case of a constraint checking for duplicate referenced in collections, we were improperly handling that empty reference. All tidied up now.

## Testing instructions

- [ ] Navigate to `/admin/import`
- [ ] Upload a basic, required fields only collection
- [ ] Run through the import, should finish successfully

### DIFF
--- a/modules/mukurtu_collection/src/Entity/Collection.php
+++ b/modules/mukurtu_collection/src/Entity/Collection.php
@@ -42,7 +42,6 @@ class Collection extends Node implements CollectionInterface, CulturalProtocolCo
           'auto_create_bundle' => '',
         ]
       ])
-      ->setDefaultValue('')
       ->setCardinality(-1)
       ->setRequired(FALSE)
       ->setRevisionable(TRUE)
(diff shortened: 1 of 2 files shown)


michael-wynne-wsu (commented):
@m4olivei Yes that worked for the basic required fields!

I did also run a populated `Items in collection` field referencing multiple DH names and that worked well. Do you want to merge this in as is or let me test a few more entity reference scenarios in this PR?

PR #1402, opened 2026-02-24
1384: Community records bug fixes

### LINKED ISSUE #1384: Determine the correct structure for importing NEW community records

m4olivei:
Resolves #1384 

## Description

The implementation was mostly already there wrt Community Records. There was just a bug in a validator that was causing errors.

Some important notes for CR import:

* While not visible on the forms, CR's make use of the "Original Record" field to reference the canonical content. As such `*.csv` files meant to import CR's need to include this column. The value can be an `id`, `uuid`, or `label` (aka title, name) for existing content. 
* If a row in a `*.csv` for a CR wants to reference a DH or other original record in the current import session, it can do so by using a value for "Original Record" equal to the `label` or `uuid`
* If a single `*.csv` is used both for adding the original records and adding CR's of those, the CR's must come later in the `*.csv`. If the wrong ordering is set, it will silently fail, where the CR is created with an empty reference to an Original Record. Effectively making it a plain old DH item (if DH was the intended Original Record).
* The importer will attempt to intelligently order multiple `*.csv` files where one `*.csv` is for new original records and the other is for CR's of those.

### Testing instructions

- [ ] Create a DH item on the site
- [ ] Import one or more CRs by using the `id` / `uuid` / `label` of the existing DH item as the value for the Original Record column
- [ ] Verify the CR was created with the correct values on import
- [ ] Import a DH item and CR for it in the same import session, with separate `*.csv` files for the DH item, the CR, and the Media needed.
- [ ] Verify the CR and DH items were both created and correctly linked together
- [ ] Import a DH item and CR item in the same `*.csv`. Order the DH item first. Add the combined `*.csv` as well as a CSV for media if needed.
- [ ] Verify the CR and DH items were both created and correctly linked together

### DIFF
--- a/modules/mukurtu_dictionary/src/Entity/WordList.php
+++ b/modules/mukurtu_dictionary/src/Entity/WordList.php
@@ -79,7 +79,6 @@ class WordList extends Node implements WordListInterface, CulturalProtocolContro
           'auto_create_bundle' => '',
         ]
       ])
-      ->setDefaultValue('')
       ->setCardinality(-1)
       ->setRequired(FALSE)
       ->setRevisionable(TRUE)
@@ -105,7 +104,6 @@ class WordList extends Node implements WordListInterface, CulturalProtocolContro
           'auto_create_bundle' => '',
         ]
       ])
-      ->setDefaultValue('')
       ->setCardinality(1)
       ->setRequired(FALSE)
       ->setRevisionable(TRUE)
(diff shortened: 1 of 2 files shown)


michael-wynne-wsu (approved):
@m4olivei All DH tests have worked successfully for me.

I'm happy to approve this as is, but if you want someone else to review other content types earlier on Wednesday @christa-kavanagh-wsu can probably do so.

PR #1405, opened 2026-02-25
1383 : Multipage item import adjustments

### LINKED ISSUE #1383: Determine the correct import structure for NEW multipage items.
- [ ] Is dependent on the settings at `/admin/config/mukurtu/multipage-item` (but is expected to be mostly DH items).
- [ ] How to indicate parent/child nodes within the same sheet (ALL new).
- [ ] How to indicate parent node (new children being added to an existing parent).
- [ ] How to set order of nodes.

m4olivei:
Resolves #1383 

## Description

Multipage item support was mostly there already! The one gap was we didn't have a results listing that shows which new multipage items were created during the import session. That gets added here.

Lets go through some of the open questions in the issue as well:

> Is dependent on the settings at /admin/config/mukurtu/multipage-item (but is expected to be mostly DH items).

Yep. Lets turn these all on for testing. The rendering of the created multipage item may be off for some, but thats a separate issue. For purpose of the importer, we're mainly looking for the Multipage item Pages field to correctly point at the items.

> How to indicate parent/child nodes within the same sheet (ALL new).

Multipage items are a custom entity. For this scenario, users need to upload at least two `*.csv` files, eg. `Multipage item.csv` and `Content - Digital Heritage.csv`. The `Multipage item.csv` should have a column mapped to the Pages field, which is an entity reference. The value can be either a nid, uuid, or label. The importer should work out the content `*.csv` files to import first, creating the content, followed by the multipage item `*.csv`. Here is a set of `*.csv` to illustrate:

[Multipage Item - Item in Import.csv](https://github.com/user-attachments/files/25556721/Multipage.Item.-.Item.in.Import.csv)
[Media - Image.csv](https://github.com/user-attachments/files/25556723/Media.-.Image.csv)
[Content - Digital Heritage.csv](https://github.com/user-attachments/files/25556724/Content.-.Digital.Heritage.csv)
[soccer.jpg](https://github.com/user-attachments/assets/731ebec5-390e-4651-a06c-8752ff5d979f)

> How to set order of nodes.

Setting multiple items for a Multipage item looks like this:

```csv
Title,Pages
Test Multipage Import,5;4;3;2
```
Notice the value for `Pages`: `5;4;3;2`. Multiple values are separated by `;`. The order set here is respected on import.

## Testing

- [ ] Configure multipage import settings at `/admin/config/mukurtu/multipage-item`
- [ ] Add one or more DH items or other content
- [ ] Author a `*.csv` for Multipage items, referencing existing content using label / id / uuid
- [ ] Import Multipage items
- [ ] Test an import with multipage items referencing new content in the same import (see above)

### DIFF
--- a/modules/mukurtu_multipage_items/src/Plugin/Validation/Constraint/MultipageValidNodeConstraintValidator.php
+++ b/modules/mukurtu_multipage_items/src/Plugin/Validation/Constraint/MultipageValidNodeConstraintValidator.php
@@ -6,6 +6,7 @@ namespace Drupal\mukurtu_multipage_items\Plugin\Validation\Constraint;
 
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Field\EntityReferenceFieldItemListInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu (approved):
Yes that is now working as expected, thank you!

I'd like to make sure the additional robustness around trailing spaces and empty values can be brought across to all importers. Let me know if a separate issue is needed for that.

PR #1410, opened 2026-02-26
1363: Dictionary word import

### LINKED ISSUE #1363: Dictionary word import goals
- Dictionary words have a level of internal complexity that isn't present in most other content types (or at least not to the same extent).
- They contain repeatable paragraphs (word entries).
- And the word entry paragraphs also contain repeatable paragraphs (sample sentences).
- The legacy reasons for this....
   - Originally (v3) dictionary words were single entry and had no paragraphs (no word entries, and sample sentences were a repeatable text field).
   - Later (v3) word entries as a concept were added. They are intended to capture things like conjugations, alternate forms, declensions, etc, without requiring a completely unique dictionary word. So the base word includes a subset of fields that was repeated as a word entry paragraph. Users always found this unusual, since seemingly "basic" fields, like translation, required creating an initial word entry instead of being part of the base word. Note that sample sentences are a part of word entries, not the base word. 
   - Now (v4) the word entry paragraphs have been maintained and sample sentences have been converted to paragraphs so that a single sample sentence can bundle text and audio together. So now we have a paragraph within a paragraph.
- In the UI form `/node/add/dictionary_word`, it works pretty logically. Users can create a "basic" word (meaning a word with all fields, but NOT requiring the use of word entries) using obvious field groups, and if needed, additional word entries are in their own field group.
   - The Mukurtu Essentials group contains most of the "base" fields we associate with a word. This is essential one iteration of a word entry, but presented as the base word, with a few fields that are about the word itself (protocols, language, glossary entry). This is not currently reflected in the importer `/admin/import/format/node/dictionary_word`.
   - The Word Entries group is the repeatable paragraph and contains most of the Mukurtu Essentials field, except as noted above.
   - The Additional Fields and Related Content groups are again fields for the base word, but grouped out to align with other content types for consistency.
- While we don't have any data, I suspect that MOST words in current Mukurtu sites DO NOT use word entries (although their use is not rare), and MANY words in current sites DO use multiple sample sentences.
- So keeping the structural complexity in mind, ideally, we would have, at a minimum,
   - A robust import tool that is faithful to the underlying data. Meaning yes, we probably have to have complex words imported across multiple spreadsheets because of paragraphs.
      - This should include paragraph/entity references that are as user-friendly as possible.
   - A more "bespoke" import tool that makes the process more user-friendly, and mirrors the UI experience.
      - This likely means including direct field mapping for Mukurtu Essentials, Additional Fields, and Related Content groups. Word entries would need to be in a separate sheet as usual.
      - Unsure how to handle multiple sample sentences in this case.

m4olivei:
Resolves #1363 

## Description

Adds the ability to configure an identifier column for any given import `*.csv`, which can then be used in reference fields to connect imported entities.

## Testing instructions

- [ ] Import one or more Dictionary Words. This will require:

* A `*.csv` file for all Sample Sentences (both for the main Dictionary Word, and any Word Entries). Include a column with some unique ids. The values don't matter, as long as they are unique.
* A `*.csv` file for all Word Entries. Include a column with some unique ids. The values don't matter, as long as they are unique. Include a column for "Sample Sentences", where each value is one of the unique ids from above
* A `*.csv` file for all Dictionary Words. Include a column for "Sample Sentences", where each value is one of the unique ids from above. Include a column for "Word Entries", where each value is one of the unique ids from above

- [ ] During the mapping step of the import, you do not need to map your custom id column
image

- [ ] During the mapping step of the import, set the "Identifier Column" under "File Settings" to the column with unique ids
image

### DIFF
--- a/modules/mukurtu_import/src/Form/CustomStrategyFromFileForm.php
+++ b/modules/mukurtu_import/src/Form/CustomStrategyFromFileForm.php
@@ -90,6 +90,18 @@ class CustomStrategyFromFileForm extends ImportBaseForm {
       ],
     ];
 
+    $header_options = ['' => $this->t('- Computed -')];
+    foreach ($this->getCSVHeaders($file) as $header) {
+      $header_options[$header] = $header;
+    }
+    $form['identifier_column'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Identifier Column'),
+      '#description' => $this->t('Optional. Select a column to use as the unique identifier for each row. When set, this takes precedence over entity ID, UUID, and label columns for tracking rows in the import. Use this when importing entities without a natural label (e.g. paragraphs) so they can be referenced by other CSVs in the same import session.'),
+      '#options' => $header_options,
+      '#default_value' => $this->importConfig->getConfig('identifier_column') ?? '',
+    ];
+
     $this->buildMappingTable($form, $form_state, $file);
 
     // File options for import like multivalue delimiter, enclosure, etc.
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu:

Link:
- https://pr1410-0bpmxllpic6cinpp7grjldb8y68fa2eo.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69a09ae9befda58aeb32174c

michael-wynne-wsu:
@m4olivei gah, I swore I triple checked that, thanks...

However, I tried again, copy/pasted in a different identifier, and am still getting the same error.

[samplesentence.csv](https://github.com/user-attachments/files/25589782/samplesentence.csv)
[word-sample-sentence.csv](https://github.com/user-attachments/files/25589783/word-sample-sentence.csv)

michael-wynne-wsu:
Related to making the identifier column more prominent, can we make the name of the sheet being mapped more prominent? I always go hunting for it when my import had multiple sheets to map. I would think as a heading at the top of the page?

Screenshot 2026-03-03 at 11 06 21 AM

PR #1425, opened 2026-03-09
1111: Theming person content type

### LINKED ISSUE #1111: Person record display updates
- [ ] "Biography sections" label is appearing even when there are no sections added.
- [ ] Relabel that to "Biography" too.
- [x] Sidebar is not displaying the correct CSS.
- [ ] Related people should be "name (relation)" (and linked as a whole to the term, no need to link the relation type).
- [ ] Referenced content is not displaying (it did work previously)
- [ ] All other issues are relevant to all content types, and logged elsewhere.

Javi-er:
Closes #1111 

This PR adjust the theming of the person full display page to match the requirements.

### Changes

- Related People now display as "Person Name (Relation)" linked to the person's page, instead of separate unlinked fields
- Related Content section now uses the featured view mode, showing a 3-column thumbnail grid instead of teasers
- Biography sections label renamed from "Biography sections" to "Biography"
- Name field removed from the sidebar.
- Section headings (h2/h3) styled correctly for biography sections on the person page

### How to test
1. create or edit a person and fill out all the information.

### Notes
Regarding this point:

> There is a "referenced content" feature that aggregates all site content that the person record is connect to via the "Other names" taxonomies. https://docs.mukurtu.org/person-records/PersonRecords/#other-names
> 
> That seems to be pulling in the content properly, but the section is labelled "related content", which is incorrect.

The related content works through the related content section, I think there is a miss-understanding here, the "other names" field is actually displayed in the sidebar.


Javi-er:
@michael-wynne-wsu @m4olivei I changed the following:

- Person preprocess hook renames"Related Content" to "Referenced Content" label only for the person content type.
- Keywords field moved to the bottom of the sidebar in the person full view display.
- Entity browser dependencies fixed, both mukurtu_content_browser and mukurtu_taxonomy_record_term_browser now listed. This seems to have been an artifact of feature export.

PR #1428, opened 2026-03-10
1408: Person import edge cases

### LINKED ISSUE #1408: Person record import goals
- [ ] Minimally required fields should import as expected:
   - [ ] Name
   - [ ] Cultural protocol fields
- [ ] All other "common" fields must import as expected (these are replicated across multiple content types, so hopefully they are already functional):
   - [ ]Media assets
   - [ ] Keywords
   - [ ] Locations fields
   - [ ] Local contexts projects/labels
   - [ ] Related content
- [ ] Person record specific fields should import as expected (these may get very messy):
   - [ ] Other names (taxonomy reference)
      - [ ] Note the configuration settings at `/admin/config/mukurtu/person-records`
      - [ ] See also #1406 
   - [ ] Date born (ISO date)
   - [ ] Date died (ISO date)
   - [ ] Deceased (boolean)
   - [ ] Related people (paragraph that includes taxonomy references and references to other person records)
   - [ ] Text sections (paragraph)

m4olivei:
Resolves #1408 

### Description

Person import was mostly good, but there was a bug that I found around Other Names, or any taxonomy import for that matter. For taxonomy fields that are configured to allow multiple bundles, the code was just taking the first one as allowed for assignment, rather than passing through the list of allowed bundles (vocabularies in the case of Other Names field).

I also took the opportunity to fix an unrelated issue where we missed specifying that the configuration (including Identifier Column) on Import Strategy config items should be exported and saved. This meant that if you returned the importer with a new set of files and attempted to reuse an Import Strategy, it would not have the configuration originally set for Identifier Column or any config under File Settings. That should now work.

## Testing instructions

NOTE: Generally speaking, if your Person record has values for "Related People", the Related Person paragraphs included in the same import need to reference previously existing Person content.

- [ ] Import a full set of Person data
- [ ] During the "Import - File Configuration", under "Import Configuration Template" Save the template, especially for Paragraphs where your setting a Identifier Column.
- [ ] Verify that all the expected entities were created
- [ ] Be sure to test with multiple values of Other Names, across the various configured vocabularies (`/admin/config/mukurtu/person-records`).
- [ ] Come back and run a brand new import (use Reset function or "Start New Import" at the end of a previous import)
- [ ] Reuse a import configuration you previously saved where you had set an "Identifier Column". Click Customize Settings
- [ ] Verify that the Identifier Column is set to the previously configured value

### DIFF
--- a/modules/mukurtu_person/src/Entity/Person.php
+++ b/modules/mukurtu_person/src/Entity/Person.php
@@ -39,7 +39,6 @@ class Person extends Node implements PersonInterface, CulturalProtocolControlled
           'auto_create_bundle' => '',
         ]
       ])
-      ->setDefaultValue('')
       ->setCardinality(-1)
       ->setRequired(FALSE)
       ->setRevisionable(TRUE)
@@ -69,7 +68,6 @@ class Person extends Node implements PersonInterface, CulturalProtocolControlled
           'auto_create_bundle' => 'audio',
         ]
       ])
-      ->setDefaultValue('')
       ->setCardinality(-1)
       ->setRequired(FALSE)
       ->setRevisionable(TRUE)
@@ -196,7 +194,6 @@ class Person extends Node implements PersonInterface, CulturalProtocolControlled
           'auto_create_bundle' => 'people',
         ]
       ])
-      ->setDefaultValue('')
       ->setCardinality(-1)
       ->setRequired(FALSE)
       ->setRevisionable(TRUE)


michael-wynne-wsu (approved):
@m4olivei I think this is all working as expected. In practice we will likely recommend that when creating new person records and relationships that it be done in two steps: import person records, then update those records with the related people paragraphs.

PR #1429, opened 2026-03-10
1427: Make Local Contexts import delimiter configurable, default to >

### LINKED ISSUE #1427: Import Local Contexts, update delimeter
In import spreadsheets we currently use `:` to separate Project:Label names. Based on feedback from LC, we should use `>` instead.

m4olivei:
Resolves #1427 

## Description

Change the compound project/label delimiter from a hardcoded colon to a configurable value (default ">") to avoid conflicts with the internal stored format (project_id:label_id:display). The delimiter is exposed as a "Local Contexts Delimiter" textfield under File Settings in the import strategy form.

## Testing instructions

- [ ] Setup the environment including importing local context data
- [x] Try an import for content that accepts a local context project / label with the new default delimiter of `>`
- [ ] Try an import configuring an alternative delimiter, eg. `|`. You can configure this on the Import - File Configuration page under the File Settings. Look for the "Local Contexts Delimiter" configuration option.


michael-wynne-wsu (approved):
Yes, various delimiters all worked as expected. Good to merge!

Issue #1406, opened 2026-02-25
"Other Names" potential updates


michael-wynne-wsu:
- In person records and place names (in progress), we have the excellent "other names" entity reference that allows users to select existing terms from defined taxonomies in order to aggregate site content that references those terms in one place.
- For example, we can connect all DH items and dictionary words where a person is named in either creator, contributor, or people field. Or if a person has been identified by multiple names.
- Doing more exploration into person records and place names, there are definitely cases where a place or person should have multiple names provided in the content, but those are not entered as terms in the corresponding taxonomies.
- As it stands a user would need to create those terms in a taxonomy of choice, and then reference them in the content. That is kind of annoying.

- Two possible solutions:
   1. We add a new field to person and place records for other names. That is not directly connected to the current other names aggregation field. Probably a taxonomy field, but could also be plain text.
      a. I think renaming "Other names" to "linked names" and then calling the new field "other names" would make sense to users. 
   3. Explore if the current "other names" modal could be modified to create new terms in the allowed taxonomies on the fly.
<img width="1642" height="708" alt="Image" src="https://github.com/user-attachments/assets/aa393179-cd6e-4f1c-8e74-737f9203158d" />

michael-wynne-wsu:
@nick-deer Just a shot in the dark, but do you have any sense if the "other names" modal in person records could be adapted to allow on the fly taxonomy term creation? I SUSPECT it would be a big hassle, but just curious if you have any initial thoughts.

michael-wynne-wsu:
Actually, it does already have People as the target vocabulary, just doesn't have on the fly creation enabled. Enabling that doesn't quite do it, I think we're using a custom form there so presumably it would need updating. I think that's `modules/mukurtu_taxonomy/src/Plugin/EntityBrowser/Widget/MukurtuTaxonomyRecordSelection.php` ?

michael-wynne-wsu:
Ignore previous comment, I wanted the "tagify" widget, not "tagify select". See #1436

PR #1432, opened 2026-03-11
931: Correctly handle other names' fields respecting taxonomy settings

### LINKED ISSUE #931: Person record - representative term issue
Received the following error when trying to select representative terms from several different (seemingly properly enabled) taxonomies.

Have not tested exhaustively, but category, language, both gave an error, Not that we would expect to use those vocabularies.

Creator worked fine.

Nothing in the recent log messages.

Image

m4olivei:
Resolves #931

### Description

Removes the limited `target_types` restrictions for Other Names fields on Person and Place in favor of a new field constraint which will correctly enforce the configured allowed vocabularies for each field.

Also resolving a couple of issues in testing this with the importer.

### Testing instructions

- [x] Minimal setup with Community, Protocol as usual.  Also create some Taxonomy terms under Creator, Contributor, Keywords, Location, Publisher, and People
- [x] Create a new Person and Place
- [x] Set values for Other Names fields. Verify that the dialog limits the list to only those configured
- [x] Verify you are able to save the Person and Place without error
- [x] Adjust the configuration at `/admin/config/mukurtu/person-records` and `/admin/config/mukurtu/place-records`
- [x] Edit the existing Person and Place records and add/remove values from the Other Names fields. Verify expected behavior

NOTE: If you remove a vocabulary from the allowlist and we have content with Other Names terms that include the newly disallowed vocabulary, that value is effectively grandfathered in, and should be allowed to be saved. 

- [x] Also try a Person import with a value for Other Names. Note that for the time being, the terms referenced must already exist.
- [x] Try a Person import with a value for Other Names in disallowed a vocabulary for the field. It should display an error.

### DIFF
--- a/modules/mukurtu_import/src/Plugin/MukurtuImportFieldProcess/EntityReference.php
+++ b/modules/mukurtu_import/src/Plugin/MukurtuImportFieldProcess/EntityReference.php
@@ -94,15 +94,35 @@ class EntityReference extends MukurtuImportFieldProcessPluginBase implements Con
     if ($ref_type == 'taxonomy_term') {
       $target_bundles = $field_config->getSetting('handler_settings')['target_bundles'] ?? [];
       $all_target_bundles = array_keys($target_bundles);
+      $auto_create = $field_config->getSetting('handler_settings')['auto_create'];
+      $auto_create_bundle = $field_config->getSetting('handler_settings')['auto_create_bundle'] ?? NULL;
 
-      $ref_process = [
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu (approved):
@m4olivei Yes works completely as expected!

The only note, which I assume is dependent on https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1406 and we'll address later is that including a new term doesn't throw an error, it just creates the person and ignores that term.

PR #1435, opened 2026-03-11
1343: Fix edge cases and misc Roundtrip issues related to limited access testing

### LINKED ISSUE #1343: Determine roles and permissions related to import and export
- Which roles get import permissions?
- In v3 it was only administrator and mukurtu admin. 
- But we have had to requests to expand that to other user roles so that people can run import/export WITHOUT having all the site-wide admin access.
   - It is essential that users can't bypass the scope of existing content creation or protocols.
- So at a minimum
   - Drupal administrator (site role)
   - Mukurtu manager (site role)
- Beyond that, we could consider...
   - Making it a per-user granted permission (I kind of like that)
   - Protocol steward? But again that could expose unwanted permissions.
   - Contributor? That might be giving too many people access.

m4olivei:
Resolves #1343 

### Description

Main pieces tackled here:

* (Bonus) Fix issues that would sometimes arise when toggling to media on "Customize Import Settings" form
* (Bonus) Clean up a placeholder settings page at `/admin/config/system/mukurtu-import` which was never used
* Add access checks for the import process, so entity creation that results from the importer respects permissions
* Add a new `mukurtu_roundtrip_manager` role with roundtrip permissions. Assign this role to authenticated users who otherwise just have permissions on the Community/Protocol groups
* Fixed some weirdness with the results table where for Word Lists, even though the importer created content, it would not immediately show for limited access users. This turned out to be a great find, and could potentially fix other very, very sneaky bugs.
* Fixed supressed errors when multiple `*.csv` files are imported in the same import, and referenced entities have issues (eg. Related Person paragraph)

### Testing Instructions

- [ ] Create a new user and assign them the site-wide `mukurtu_roundtrip_manager` role. Also assign them as a member of a Community and Protocol. Test with different assigned Protocol roles (Protocol Steward, Contributor, etc.)
- [ ] Run through the import for various content / media. Verify that limited access users can only create content they are allowed to create
- [ ] Test editing existing content
- [ ] As a group level Conributor user, test editing content that is not your own. Should recieve an error

There are lots of iterations to hit here. I have not yet tested exhaustively. Will be continuing more testing, but do feel free to report issues! Would love the help.

### DIFF
--- a/modules/mukurtu_import/src/Plugin/migrate/process/MukurtuEntityLookup.php
+++ b/modules/mukurtu_import/src/Plugin/migrate/process/MukurtuEntityLookup.php
@@ -30,7 +30,7 @@ class MukurtuEntityLookup extends EntityLookup {
     $id = is_array($value) ? reset($value) : $value;
     if (is_numeric($id)) {
       $entity = $this->entityTypeManager->getStorage($this->lookupEntityType)->load($id);
-      if ($entity && (!$this->accessCheck || $entity->access('view'))) {
+      if ($entity) {
         return TRUE;
       }
     }
@@ -46,8 +46,10 @@ class MukurtuEntityLookup extends EntityLookup {
     }
 
     // Below is basically a copy of migrate_plus EntityLookup::query but with
-    // some tweaks to always respect access and to throw an exception when
-    // multiple query results are found.
+    // a tweak to throw an exception when multiple query results are found.
+    // Access checking is deferred to entity validation, specifically
+    // ValidReferenceConstraintValidator, which grandfathers pre-existing
+    // references the user cannot view while blocking new ones.
 
     // Entity queries typically are case-insensitive. Therefore, we need to
     // handle case-sensitive filtering as a post-query step. By default, it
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu:
Fantastic, sounds like lots of little snarls got identified here too. Will jump on this in the morning and probably throw together some kind of tracking list/matrix to make sure all roles are explored.

michael-wynne-wsu:

Link:
- https://pr1435-b0bazouluxc12tk8dxexaxrk6glhtdob.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69b1dde728647a72ffa7299d

michael-wynne-wsu:
- There is no "community record" import type available in the mapping (which makes sense).
- Community Record Stewards, unless they have an additional role that gives content type creation permissions cannot import a community record (and I suspect it would be rare for this arrangement to happen, but it's a new role so we don't have experience with it). 
- They CAN create one through the web interface though.

Screenshot 2026-03-12 at 1 48 16 PM

michael-wynne-wsu:

Link:
- https://pr1435-b0bazouluxc12tk8dxexaxrk6glhtdob.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69b1dde728647a72ffa7299d

m4olivei:
> * When trying to update an MPI, receive an error that pages cannot be null, whether or not the field is present.
> * Affects both contributors and protocol stewards, does not matter who is the node author (edit own/edit any)
> 
> `MPI-edit-any.csv: [multipage_item]: field_pages=This value should not be null.`

Hum, can I see the `*.csv`? I'm not able to replicate this one. For an import that edits content, or multipage items or any entity, either `ID` or `UUID` must be mapped. Otherwise it will consider the row as a new item, in which case this error would be correct since Pages is a required field.

m4olivei:
> * Curator could import and populate a new collection with content from other users, but did not get an error when referencing invalid content, it was simply skipped.
>   
>   * Eg: referencing DH items that are part of a multipage item. Correctly disallowed, but did not get a warning or error.
> 
> [Collection-reqd-only-failure.csv](https://github.com/user-attachments/files/25944241/Collection-reqd-only-failure.csv) [Collection-reqd-only-success.csv](https://github.com/user-attachments/files/25944243/Collection-reqd-only-success.csv)

Hum, I'm not able to reproduce the issue here. Also, I don't believe that we have any validation that prevents users from adding items to Collections that are in an MPI. I seem to be able to do that from the UI. There is validation that prevents content from being in more than one MPI, eg. a DH item can only be in a single MPI.

PR #1436, opened 2026-03-13
Adding tagify module for testing with "other name" fields

michael-wynne-wsu:
Trying to make improvements as discussed in https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1406 
Matt, if this is not what you were expecting or this seems way off base, let me know!

- [Tagify](https://www.drupal.org/project/tagify) module installed and set as default for the Person Record > Other Names field.

**Testing**

- [ ] Create community and protocol.
- [ ] Add some terms to creator, contributor, and/or people vocabularies.
- [ ] Create a person record.
- [ ] Other Names field should display all available terms and filter based on entered text.
- [ ] Try creating a new term on the fly.

**Things that are not working**

- [x] The field seems to be showed ALL vocabularies, not just those enabled by the other names field...
   - [x] This does not seem to be an issue with regular taxonomy fields, so likely due to this custom work.
- [x] Adding terms on the fly creates them but they're not added to any vocabulary, which breaks when trying to edit.
- [x] Create new terms in people bundle SHOULD be enabled but it's not working.
- [x] There should be an [info label](https://www.drupal.org/docs/extending-drupal/contributed-modules/contributed-module-documentation/tagify/info-label-optional) option so we can display vocabulary names but I don't see that option.
- [x] Text colour is low contrast in dark gin mode.

**Other planned work**

- [x] If we like this, expand it to place names.
- [ ] Consider expanding it to all/select taxonomy fields site-wide.

### DIFF
--- a/modules/mukurtu_person/config/install/core.entity_form_display.node.person.default.yml
+++ b/modules/mukurtu_person/config/install/core.entity_form_display.node.person.default.yml
@@ -282,14 +282,18 @@ content:
       display_label: true
     third_party_settings: {  }
   field_keywords:
-    type: entity_reference_autocomplete
+    type: tagify_entity_reference_autocomplete_widget
     weight: 9
     region: content
     settings:
       match_operator: CONTAINS
-      match_limit: 10
-      size: 60
+      match_limit: 0
       placeholder: ''
+      suggestions_dropdown: 0
+      show_entity_id: 0
+      show_info_label: 0
+      info_label: ''
+      parent_selection: 1
     third_party_settings: {  }
   field_local_contexts_labels_and_notices:
     type: local_contexts_label_and_notice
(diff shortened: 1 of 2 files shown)


michael-wynne-wsu:

Link:
- https://pr1436-dntfjavs9plcdmmbcckhulmqu1r3ecwh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69b4544a28647a72ffd62a44

michael-wynne-wsu:
@m4olivei the fields are now properly referencing the enabled taxonomy/ies, great!

On the fly creation is not working though.

Screenshot 2026-03-17 at 12 44 16 PM
Screenshot 2026-03-17 at 12 44 21 PM

michael-wynne-wsu:
@m4olivei this is PERFECT, thank you! If people need different behaviour on autocreate, that's far enough off of "normal" use that I'm not too concerned.


@joelle-birano-wsu @christa-kavanagh-wsu Do you want to check this again? And I'm considering rolling it out to all taxonomy fields.

PR #1439, opened 2026-03-16
1420  place record display updates

### LINKED ISSUE #1420: Place record display updates
Should be able to mostly copy appearance from Person records since this is built off that module. This is new, I don't have existing content to reference.

The one big difference would be a more prominent map display. Ideally that would include the referenced content as well.

**Main section**

- [x] Media assets (same as person records)
- [x] Text Sections (similar to person records "biography", but I don't think we need the "text sections" label)
- [x] Map (should include the place records map points AND the referenced/related content map points)
- [x] Location description (same as person records)
- [x] Location (same as person records)
- [x] Referenced content (same as person records - a bit redundant with the map, but not all content will have map points)

**Sidebar**

- [x] Communities and Cultural Protocols (same as person records)
- [x] Place type (handle like other taxonomy fields)
- [x] Other names (same as person records)
- [x] Keywords (same as person records)
- [ ] Local Contexts (same as person records)

Javi-er:
Closes #1420 

This PR themes the "place" content type and modifies `MukurtuLeafletFormatter` for adding an extra field setting for retrieving all the related content on the field coverage map, this is an opt-in feature, it means that the same setting can be enabled for other content types that also need to retrieve all the related content for `field_coverage`, for now it's enabled just for "place".

### How to test
1- Create a few "location" terms.
2- Edit or create a dictionary word and make sure that it has a location.
3- Create a "place" node making sure:
   1- That it has a location selected.
   2- On related content choose the dictionary word.
   3- Under "other names" choose one of the location terms.
   4- Create another location (with a point selected in the map) that uses the same location for "other names".

When saved, the location should show it's own location in the map with the location for the related dictionary word and the location of the other place that also uses the same term for "other names" .


michael-wynne-wsu:

Link:
- https://pr1439-mpneikpbjw1rqp1xwgkizdkwa9lyq7kw.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69b86b33d5d87f969e5f7d49

modules/mukurtu_place/config/install/core.entity_form_display.node.place.default.yml
michael-wynne-wsu: ... this, which uses the new widget.

michael-wynne-wsu:
@javier this look good then thank you!

PR #1452, opened 2026-03-19
1451: Fix cultural protocol data loss when user saves content with inaccessible protocols

### LINKED ISSUE #1451: Data loss on Cultural Protocols field
Where you have a content item that is set for multiple cultural protocols, set to Any, such that any user with one of the protocols has access to edit, when a user that has access to only a subset of the protocols the content is set to edits the item and Saves, all the protocols the user does not have access to are removed.

**Steps to reproduce**

1. Create 2 communities, with a protocol each, (eg. Community A > Protocol A, Community B > Protocol B). 
2. As an admin create 2 DH items in Protocol A, we're going to throw these into a collection next
3. As an admin create a Collection, set it as part of Protocol A and Protocol B, mark as Any. Add the two DH Items to the collection
4. As a curator for Protocol B, login and edit the Collection
5. Change the title and Save. 
6. As the admin owner, edit the Collection to observe the Cultural Protocols field

**NOTE: Using Collections as an example here to observe that the Items in collection field has the correct behavior, in that even though the user from Protocol B does not have access to the items, they are still retained. This issue is generally present on any content with a Cultural Protocol field.**

**Expected result**

Collection is under Protocol A and Protocol B

**Actual result**

When the curator saved the Collection, connection to Protocol A was lost

m4olivei:
Resolves #1451 

### Description

Pre-existing protocols the user cannot apply were being stripped on save because `preSave()` treated them as unauthorized additions. Now only genuinely new protocols (not in the original) are checked against the user's apply permission, so grandfathered-in protocols are preserved.

### Testing instuctions

- [ ] Run through steps to reproduce from the issue description. Should have desired result
- [ ] Try with media, and various content types, generally any entities that have a `cultural_protocol` type field.

### DIFF
--- a/modules/mukurtu_protocol/src/Plugin/Field/FieldType/CulturalProtocolItem.php
+++ b/modules/mukurtu_protocol/src/Plugin/Field/FieldType/CulturalProtocolItem.php
@@ -19,6 +19,7 @@ use Drupal\mukurtu_protocol\CulturalProtocols;
  * )
  */
 class CulturalProtocolItem extends FieldItemBase {
+
   /**
    * {@inheritdoc}
    */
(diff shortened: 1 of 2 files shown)


michael-wynne-wsu:
Yes! So far this has worked as expected when import/editing a collection, neither the protocols not the content in collection were messed up. I'll do some broader testing, but if this carries across other content as expected it'll be perfect.

PR #1463, opened 2026-03-24
1448 : Fix various issues surrounding import for Creative Commons and Rights Statements

### LINKED ISSUE #1448: Import not accepting expected Creative Commons and Rights Statements as expected
Noted by @christa-kavanagh-wsu 

- [ ] When importing DH items, in the Creative Commons Licenses field, only the labels of the licenses are accepted, not the URLs, even though the importer says either value is allowed.
- [ ] Conversely, Rights Statement field is only accepting URLs, when both URLs and labels are stated.
- [ ] Rights Statements and Creative Commons use slightly different URL formats. One has a trailing `/`, one doesn't. I feel like we should set one for consistency, any preference/recommendation?
- [ ] RS and CC fields in the import don't tolerate a trailing space, `/`, etc. Can that be accommodated to minimize user annoyance?

Image

Image

m4olivei:
Resolves #1448 

## Description

Resolves the issues noted in the issue namely:

* Add flexibility to allow CC and right statement values to include or exclude a trailing slash when the value is a URL
* Allow the use of the value for either field, in particular Rights Statements needed specialized handling
* Standardize on including the trailing slash for Creative Commons URL values. This includes an update hook to update any existing values to add the trailing slash. This also includes handling in the migration to transform v3 values without a trailing slash to have one.

## Testing instructions

- [ ] Try a migration with content that has values for Creative Commons field. Verify the values migrated correctly.
- [ ] Try an import with various combinations of values to ensure the expected results, per the issue description
- [ ] For content with Right and CC fields set, view the front-end and ensure there were no visual regressions

### DIFF
--- a/modules/mukurtu_digital_heritage/mukurtu_digital_heritage.install
+++ b/modules/mukurtu_digital_heritage/mukurtu_digital_heritage.install
@@ -49,3 +49,54 @@ function mukurtu_digital_heritage_update_11001(&$sandbox) {
     ? min($sandbox['current'] / $sandbox['total'], 1)
     : 1;
 }
+
+/**
+ * Update Rights Statements URLs from http to https.
+ */
+function mukurtu_digital_heritage_update_11002(&$sandbox) {
+  $url_map = [
+    'http://rightsstatements.org/vocab/InC/1.0/' => 'https://rightsstatements.org/vocab/InC/1.0/',
+    'http://rightsstatements.org/vocab/InC-OW-EU/1.0/' => 'https://rightsstatements.org/vocab/InC-OW-EU/1.0/',
+    'http://rightsstatements.org/vocab/InC-EDU/1.0/' => 'https://rightsstatements.org/vocab/InC-EDU/1.0/',
+    'http://rightsstatements.org/vocab/InC-NC/1.0/' => 'https://rightsstatements.org/vocab/InC-NC/1.0/',
+    'http://rightsstatements.org/vocab/InC-RUU/1.0/' => 'https://rightsstatements.org/vocab/InC-RUU/1.0/',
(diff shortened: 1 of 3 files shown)


modules/mukurtu_digital_heritage/mukurtu_digital_heritage.install
```diff
@@ -4,3 +4,48 @@
...
+  $url_map = [
+    'http://creativecommons.org/licenses/by/4.0' => 'http://creativecommons.org/licenses/by/4.0/',
+    'http://creativecommons.org/licenses/by-nc/4.0' => 'http://creativecommons.org/licenses/by-nc/4.0/',
+    'http://creativecommons.org/licenses/by-sa/4.0' => 'http://creativecommons.org/licenses/by-sa/4.0/',
+    'http://creativecommons.org/licenses/by-nc-sa/4.0' => 'http://creativecommons.org/licenses/by-nc-sa/4.0/',
+    'http://creativecommons.org/licenses/by-nd/4.0' => 'http://creativecommons.org/licenses/by-nd/4.0/',
+    'http://creativecommons.org/licenses/by-nc-nd/4.0' => 'http://creativecommons.org/licenses/by-nc-nd/4.0/',
+  ];
+
+  if (!isset($sandbox['nids'])) {
```
m4olivei: This is a bit advanced for update hooks, but what's going on here is update hooks can be setup to run progressively, so you can split up the work and run a bit at a time to avoid out of memory errors. Since we don't know what the situation is on a given site, and they could have thousands of DH items, we update 25 at a time here. See [batch updates](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/11.x#sec_bulk) in the docs for more info.

PR #1488, opened 2026-03-31
Fix protocol all-mode access control and activate image alt constraint

alexmerrill:
- MukurtuProtocolNodeAccessControlHandler: under the 'all' sharing setting, every protocol must now individually grant edit/delete access. Previously, a qualifying role in any single protocol was sufficient, allowing unintended access to content requiring membership in all protocols.

- Image.php: attach the ImageAltRequired constraint to field_media_image so that $media->validate() actually enforces it. The constraint class existed but was never wired to the field definition.  There are other places this could also be added -- like thumbnails -- but we should confirm this approach too.

--these are two possible production fixes that came up in my CI investigation and build-out (that I will push separately).  We should confirm our any / all settings are working the way we want them to  as proposed fix could just be over tightening a screw (if yes we can disregard the changes to the MukurtuProtocolNodeAccessControlHandler).

### DIFF
--- a/modules/mukurtu_media/src/Entity/Image.php
+++ b/modules/mukurtu_media/src/Entity/Image.php
@@ -77,7 +77,8 @@ class Image extends Media implements ImageInterface, CulturalProtocolControlledI
       ->setRevisionable(TRUE)
       ->setTranslatable(TRUE)
       ->setDisplayConfigurable('view', TRUE)
-      ->setDisplayConfigurable('form', TRUE);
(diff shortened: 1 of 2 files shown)


michael-wynne-wsu:

Link:
- https://pr1488-ad73d1ys0xwdrttw99n0ldrel9ih1lhc.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69cc5b3a624c69af455b75d3

PR #1531, opened 2026-04-16
Fix access handler cache metadata to avoid manual cache clears (#1017, #1029)

### LINKED ISSUE #1017: Protocol local tasks require cache clear to update.
- Protocols can only be deleted when there is no content in them (good!)
- After deleting the last piece of content in a protocol, the delete button does not immediately appear, a cache clear is required for that to resolve.
- Similarly, in some cases, after creating the first piece of content in a protocol, the delete buttons does not DISAPPEAR, a cache clear resolves this too.

### LINKED ISSUE #1029: Cache clear on permission changes
In general, a cache clear is required to reflect updates to user permissions. Eg: adding/removing/changing group roles

- This can be done manually, but that's not a great user experience.
- A cache clear can be force triggered, but has performance impacts, especially if changes are happening often.
   - Is there a smaller OG/permissions cache that can be cleared?
- User logout/login does not address it.
- When else is it triggered? 
   - Maybe a cron run?
- In practice this may not be a big issue, but TBD.

michael-wynne-wsu:
## Summary

Closes #1017, closes #1029.

Access results across all protocol/community/collection handlers were returned without cache tags or dependencies. Drupal's render cache had no way to surgically invalidate them when underlying data changed, forcing full manual cache clears after routine operations like role assignments, protocol content changes, and privacy setting updates.

- **`ProtocolAccessControlHandler`**: add OG membership deps on `view`/`update` results; add `node_list`/`media_list` tags on `delete` results so the protocol delete button appears/disappears without a cache clear; use `CacheableMetadata` accumulator in the community loop
- **`Community::addMember/removeMember/setRoles`**: remove `cache.render->invalidateAll()` — now that access results carry OG membership cache tags (auto-invalidated by Drupal when `$membership->save()/delete()` is called), the nuclear option is no longer needed
- **`CommunityAccessControlHandler`**: carry entity dep on all returns (sharing setting and parent/protocol field changes); carry membership dep or `user:{id}` tag on membership-dependent results
- **`MukurtuProtocolNodeAccessControlHandler`**: carry entity dep on empty-protocol-set and non-member denials; `CacheableMetadata` accumulator in `checkCreateAccess` membership loop
- **`MukurtuProtocolMediaAccessControlHandler`**: same changes as node handler
- **`MukurtuCommentAccessControlHandler`**: accumulate membership deps in protocol loop for `approve` operation; guard against null membership with `user:{id}` tag
- **`PersonalCollectionAccessControlHandler`**: carry entity dep on owner/privacy/published checks

## How the invalidation chain works

| Event | Mechanism |
|---|---|
| User gains a community/protocol membership | `Community::addMember` calls `Cache::invalidateTags(["user:{id}"])`, clearing any cached result tagged with `user:{id}` |
| User's role changes | `$membership->save()` auto-invalidates `og_membership:{id}`, clearing any cached result that carries that membership as a dep |
| User loses a membership | `$membership->delete()` auto-invalidates `og_membership:{id}` |
| Content is added to a protocol | `node_list`/`media_list` tags are invalidated, clearing the cached "allowed to delete" result on the protocol |
| Content is removed from a protocol | Same — clearing the cached "forbidden to delete" result |
| Community sharing setting changes | Entity save invalidates `community:{id}`, clearing view access results that carry the entity as a dep |

## Test plan

- [ ] **#1017 — protocol delete button**: add content to a protocol → delete tab disappears without cache clear; remove all content → delete tab reappears without cache clear
- [ ] **#1029 — role changes**: assign a user a community role → their UI reflects the new permissions without cache clear; remove them from the community → access reverts without cache clear
- [ ] Regression: community/protocol browsing works correctly across users with different roles
- [ ] Regression: node/media create access works correctly for users with and without protocol memberships


michael-wynne-wsu:
@nick-deer @alexmerrill So this is work that I am not as confident in. Basically we have a handful of processes that require a cache clear to run before users see the right actions, but that could be a notable performance issue. This was an effort to tackle all of that, but other than the usual end user testing, this is all code that is beyond my grasp to review. Any thoughts?

nick-deer:
> @nick-deer @alexmerrill So this is work that I am not as confident in. Basically we have a handful of processes that require a cache clear to run before users see the right actions, but that could be a notable performance issue. This was an effort to tackle all of that, but other than the usual end user testing, this is all code that is beyond my grasp to review. Any thoughts?

It's definitely an improvement over what we have now! We can always tweak cache issues as they come up, but this is a great starting point.

nick-deer (approved):

michael-wynne-wsu:
Great, thanks Nick!

PR #1543, opened 2026-04-19
Fix message_digest_ui fatal error and stale bulk actions for missing email_user flag

### LINKED ISSUE #1239: Action menu items at admin/people break site when applied
Bulk applying the 3  "send notifications" options (daily, weekly or immediately) from the action menu breaks the site. I thought it had to do with email not being set up in previews, but I tried to edit this setting in the V4 sandbox as well and it's the same.

Image

Image

I noticed that when I'm editing an individual user, there are no other interval options besides "send immediately" so I'm guessing that's related.

Image

Here's the error log:

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1239

In the end I don't think that's a neccessary set of bulk options and removed them instead.

## Summary

- `DigestInterval::access()` was calling `->actionAccess()` on a null `$flag` property, causing a fatal error when a Views bulk form action was submitted. The constructor only loaded the flag from `$configuration['flag_id']`, but the deriver never sets that key — so the flag was never loaded.
- The `DigestIntervalActionDeriver` hardcodes `email_user` as one of the three flags to create action derivatives for, but that flag does not exist in Mukurtu. This caused actions like "Send notifications immediately on User X" to appear in bulk forms and then fail with an access error.

## Changes

- **Patch (`DigestInterval`)**: Extract `flag_id` from the derivative plugin ID when it's not in configuration; add null guard in `access()` to prevent the fatal error
- **Patch (`DigestIntervalActionDeriver`)**: Inject the flag service and skip creating derivatives for flags that don't exist on the site
- **`mukurtu_update_40004`**: Deletes the three `system.action.message_digest_interval.email_user.*` config entities already stored on existing sites

## Test plan

- [ ] Run `drush updb` and confirm update hook 40004 runs cleanly
- [ ] Confirm "Send notifications immediately/daily/weekly on User" actions no longer appear in the Users view bulk actions
- [ ] Confirm digest interval bulk actions still work correctly on nodes and taxonomy terms
- [ ] Confirm no fatal errors in Drupal logs related to `DigestInterval`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -121,6 +121,9 @@
             },
             "drupal/search_api_glossary": {
                 "TypeError: Cannot access offset of type string on string": "web/profiles/mukurtu/patches/search_api_glossary_type-error-3492668-8.patch"
+            },
+            "drupal/message_digest_ui": {
+                "Fix null flag fatal error in DigestInterval::access()": "web/profiles/mukurtu/patches/message_digest_ui-digest-interval-null-flag.patch"
             }
         }
     }
--- a/modules/mukurtu_core/src/Hook/FormHooks.php
+++ b/modules/mukurtu_core/src/Hook/FormHooks.php
@@ -88,4 +88,31 @@ class FormHooks {
       }
     }
   }
+
+  /**
+   * Removes message_digest notification actions from the user admin bulk form.
+   *
+   * These come from message_digest_ui optional config and should not be
+   * exposed in Mukurtu's user management UI.
+   */
+  #[Hook('form_alter')]
+  public function formAlterRemoveNotificationBulkActions(array &$form, FormStateInterface $form_state, string $form_id): void {
+    if (!str_starts_with($form_id, 'views_form_user_admin_people_') &&
+        !str_starts_with($form_id, 'views_form_mukurtu_people_')) {
+      return;
+    }
+
+    $actions_to_remove = [
+      'message_digest_interval.email_user.immediate',
+      'message_digest_interval.email_user.daily',
+      'message_digest_interval.email_user.weekly',
+    ];
+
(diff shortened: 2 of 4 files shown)


joelle-birano-wsu:
Looks like notification intervals options are still there.

Screenshot 2026-04-20 124729

michael-wynne-wsu:
Ah I was just focusing on the other user list. I was thinking this one might keep more stuff since it's for higher level admins?

christa-kavanagh-wsu:
I was a little confused about this one because they are still showing. They are still returning error messages.

joelle-birano-wsu:
> Ah I was just focusing on the other user list. I was thinking this one might keep more stuff since it's for higher level admins?

Which list? I checked admin/people/list on main and it looks the same as this PR. 

I'm also a little unclear about how to test this step:

- [ ]  Confirm digest interval bulk actions still work correctly on nodes and taxonomy terms

michael-wynne-wsu:
@joelle-birano-wsu @christa-kavanagh-wsu I definitely don't see them at `/admin/people/list`, but do at `/admin/people` (I'll remove them there)

And ignore " Confirm digest interval bulk actions still work correctly on nodes and taxonomy terms", that's because I didn't edit the PR description...
Screenshot 2026-04-21 at 9 32 30 AM
Screenshot 2026-04-21 at 9 32 38 AM

michael-wynne-wsu:
Okay this seems fixed to me @joelle-birano-wsu @christa-kavanagh-wsu 
Screenshot 2026-04-21 at 10 11 07 AM
Screenshot 2026-04-21 at 10 11 20 AM

christa-kavanagh-wsu (approved):
Looks good to me!

PR #1549, opened 2026-04-19
Add user memberships and roles view (#451)

michael-wynne-wsu:
## Summary

- Adds `/user/memberships` (dashboard, current user) and `/user/{uid}/memberships` (owner or admin) pages showing a hierarchical table of all community and protocol memberships with roles
- Adds a "My memberships" link to the dashboard's "My Content" menu
- Adds a "Memberships" operation to each row on `/admin/people` for quick admin access

## Test plan

- [ ] Log in as a user with community/protocol memberships → dashboard "My Content" → "My memberships" shows correct communities, roles, and nested protocols
- [ ] Log in as admin → `/admin/people` → Operations dropdown → "Memberships" opens the correct user's page
- [ ] Admin can view `/user/{uid}/memberships` for any user; non-admin cannot view another user's page (403)
- [ ] A user with no memberships sees the empty-state message
- [ ] A protocol belonging to multiple communities appears under each community the user is a member of

Closes #451

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1548, opened 2026-04-19
Change role select to checkboxes in bulk operations (#1221)

michael-wynne-wsu:
## Summary

- Adds a `hook_form_FORM_ID_alter` targeting OG's `og_membership_add_multiple_roles_action_form`
- Changes the roles field from a multi-select (`#type => select, #multiple => TRUE`) to checkboxes
- Adds a validation handler to require at least one checkbox be checked
- Adds a pre-submit handler to filter unchecked values before OG's submit handler runs

Closes #1221

## Test plan

- [ ] Navigate to a group members overview and select one or more members
- [ ] Choose "Add roles" from the bulk operations dropdown
- [ ] Confirm the roles field renders as checkboxes instead of a multi-select
- [ ] Verify submitting with no roles checked shows a validation error
- [ ] Verify selecting roles and submitting correctly assigns those roles to the selected members

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1559, opened 2026-04-24
User memberships page: WCAG accessibility + cache fix (#1549)

michael-wynne-wsu:
This is a post-hoc review and fix for https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1548

## Summary

- **Table accessibility** — adds `<caption>`, `scope="col"` + `id` on all column headers, and `headers` attributes on every data cell (including rowspan cells that need compound header chains). Orphan-protocols section gets its own `<tbody>` with a proper `<th scope="rowgroup" id>` section label referenced by each data cell. Empty cells use translatable `None` instead of a bare em dash. Fixes WCAG 1.3.1 (Info and Relationships) for complex tables with rowspans.
- **Cache metadata** — replaces `url->toString()` with `toString(TRUE)` → `GeneratedUrl` so cache contexts and tags from entity URLs are collected and applied to the render array via `BubbleableMetadata`. Prevents stale page caches when community or protocol URL routes carry per-user or per-role cache contexts.
- **Menu hierarchy** — adds `parent: mukurtu_dashboard` so "My memberships" nests correctly under the dashboard menu item.
- **Drupal CS** — adds docblocks to all public controller methods.

## Test plan

- [ ] Log in as a user with community and protocol memberships; visit `/user/{uid}/memberships` and confirm table renders correctly
- [ ] Log in as a user with orphan protocol memberships (protocols not under any of their communities) and confirm the second table section appears with correct labels
- [ ] Log in as a user with no memberships and confirm the empty-state message appears
- [ ] Run WAVE or axe on the memberships page — no errors on table headers/cells
- [ ] Confirm admin can visit `/user/{other-uid}/memberships` and anonymous gets 403


michael-wynne-wsu:
This is a post-hoc review and fix for https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1548 - seems like a few small improvements.

There are no front end changes.

nick-deer (approved):

PR #1546, opened 2026-04-19
Reorder roles on protocol and community members/add forms

michael-wynne-wsu:
Fixes #1090

## Summary
- Reorders protocol roles on `/admin/protocols/{id}/members/add` to: Protocol Member, Protocol Affiliate, Contributor, Curator, Community Record Steward, Language Contributor, Language Steward, Protocol Steward
- Reorders community roles on `/admin/communities/{id}/members/add` to: Community Member, Community Affiliate, Community Manager
- Sorting is applied in `hook_form_og_membership_form_alter` by loading each role's weight from config — the `options_buttons` widget does not respect weight natively
- Updated role config weights and added `mukurtu_protocol_update_40005` for existing installs

## Test plan
- [ ] Visit `/admin/protocols/{id}/members/add` and confirm role order matches: Protocol Member, Protocol Affiliate, Contributor, Curator, Community Record Steward, Language Contributor, Language Steward, Protocol Steward
- [ ] Visit `/admin/communities/{id}/members/add` and confirm role order matches: Community Member, Community Affiliate, Community Manager
- [ ] Run `drush updb` on an existing install and confirm roles are reordered

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1561, opened 2026-04-24
Follow-up to #1546: consistent role sort on bulk add-roles form

michael-wynne-wsu:
## Summary

Follow-up to #1546 (merged), which reordered roles on the individual add-member forms. This PR extends that work based on a post-merge review:

- Extracts `_mukurtu_protocol_sort_role_options()` as a shared helper so the sort logic lives in one place
- Applies weight-based role sorting to `og_membership_add_multiple_roles_action_form_alter` — the bulk "add roles" action form was showing roles in a different order than the individual add-member form
- Replaces the per-role `load()` loop with `loadMultiple()` (single call, cleaner)
- Adds a `mukurtu_protocol` channel warning log when an `og_role` config entity is unexpectedly missing, instead of silently falling back to weight 99

## Test plan

- [ ] Visit `/admin/protocols/{id}/members/add` — confirm role order matches: Protocol Member, Protocol Affiliate, Contributor, Curator, Community Record Steward, Language Contributor, Language Steward, Protocol Steward
- [ ] Visit `/admin/communities/{id}/members/add` — confirm role order matches: Community Member, Community Affiliate, Community Manager
- [ ] Select multiple members from a protocol member list and use the "Add roles" bulk action — confirm the roles checkbox list shows the same order as the individual form above
- [ ] Select multiple members from a community member list and use the "Add roles" bulk action — confirm Community Member, Community Affiliate, Community Manager order


michael-wynne-wsu:
@nick-deer Another review of previously merged work. No front end changes, just minor code improvements.

nick-deer (approved):

PR #1564, opened 2026-04-25
Fix three cache metadata gaps missed by #1531

michael-wynne-wsu:
## Summary

Code review of #1531 identified three cache metadata gaps that were not addressed in the original PR. All three allow stale access-control results to survive longer than they should — in two cases indefinitely.

- **Open protocol view has no entity dep** — `ProtocolAccessControlHandler::checkAccess` returned `AccessResult::allowed()` with no cache dependency when `$entity->isOpen()`. If a protocol transitions from open to restricted its cached "allowed" view result would never be invalidated. Fixed with `addCacheableDependency($entity)`.

- **`checkCreateAccess` forbidden fallthrough has no user tag** — `ProtocolAccessControlHandler::checkCreateAccess` returned a bare `AccessResult::forbidden()` when no qualifying membership was found. `Community::addMember()` calls `Cache::invalidateTags(["user:{id}"])`, but that tag was absent from this result, so a user gaining their first community membership would still be blocked from creating protocols until the cache naturally expired. Fixed with `addCacheTags(["user:{$account->id()}"])`.

- **`break` in comment approval loop skips remaining protocol memberships** — `MukurtuCommentAccessControlHandler` broke out of the protocol loop on the first missing membership, so memberships for subsequent protocols were never added to `$cacheability`. A role change on any protocol after the first would not invalidate the cached result. Changed `break` to `continue` so all protocols are visited and their deps accumulated before the result is returned.

## Context

These were found during a structured review of #1531 against the merged codebase. The CI failure on that PR (`ensure_https` plugin not found in `ImportListStringTest`) is pre-existing and unrelated.

## Test plan

- [ ] Open protocol: change protocol from open to restricted → cached view access revokes without manual cache clear
- [ ] Protocol create: add user to a community → protocol create tab appears without manual cache clear  
- [ ] Comment approval: user promoted to comment admin on a later protocol in a multi-protocol set → approve access granted without manual cache clear


michael-wynne-wsu:
Another code update from earlier work to catch missed issues. No front end changes.

nick-deer (approved):

PR #1537, opened 2026-04-17
New user workflow

michael-wynne-wsu:
Closes #291 
Closes #293

- Still in development. 
- This builds on both the admin and community manager "create user" forms.
- Adds an expandable section to notify other users about the new account. This is largely intended so that someone can ping other community managers or protocol stewards in case they can't enrol the new user in those groups themselves, or think the user may need additional memberships and wants someone else to confirm.
- The options (which can be used simultaneously) are:
  - Notify all Mukurtu Managers.
  - Notify all Community Managers in... (select from communities where you are a CM)
  - Notify all Protocol Stewards in... (select from protocols where you are a PS)
  - Notify.... (search by username)
- At the moment this sends a fairly generic "there is a new user, review their account here" email. 

To do:

- [x] Potentially build in an in-site notification center (really that's a separate issue). https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1541
- [x] Make sure that the notification options are actually useful.
- [x] Update labels.
- [x] Clean up formatting.
- [x] Restore user selection widget.
- [x] Move display name below password.
- [x] Debug error.
- [x] Resolve new "notify all mukurtu managers" error.
- [x] Ensure that new accounts created via `/admin/communities/create-user` are assigned the authenticated user site-wide role.
- [x] Resolve crash on `/admin/communities/create-user` form submit.

I would really appreciate some testing and input here before I press on. This needs additional input from the team.

Pre-merge checklist:

- [x] I have checked for and if required, created update hooks.
- [x] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [x] I have updated from `main` and resolved any merge conflicts.
- [x] I have verified that all expected checks pass.

PR #1565, opened 2026-04-26
Fix bugs and accessibility gaps in new user workflow (#1537)

michael-wynne-wsu:
## Summary

Follow-up to #1537. Addresses bugs and accessibility issues found during code review.

### Security / access control

- **Anonymous users no longer see the notify section** on `/user/register`. The form alter now returns early for unauthenticated users, preventing community and protocol names from being exposed during self-registration.
- **Non-admin users see only their managed communities and protocols** in the notify dropdowns. Previously `formUserRegisterFormAlter` loaded all communities site-wide regardless of the current user's membership. Now only site admins (with `administer users`) see everything; Mukurtu Managers see only the communities where they have the `manage members` permission, consistent with how `CommunityManagerUserCreationForm` already behaved.

### Accessibility (WCAG 2.2 AA — 1.3.1 Info and Relationships)

- Added `role="group"` and `aria-labelledby` to the protocol steward checkbox group container and the "Notify specific users" AJAX wrapper in both forms. Without these, screen readers had no programmatic association between the visible group labels and their child controls.
- Moved `margin-inline-start: .5rem` from inline `style` attributes on checkbox group elements into `mukurtu-notify-form.css`. Inline styles are not overridable by Windows High Contrast Mode user stylesheets.

### Code quality

- Extracted the duplicated notification submit logic (UID resolution from form state + group expansion + send) into a single `mukurtu_notifications_extract_notify_uids(FormStateInterface $form_state): array` helper in `mukurtu_notifications.module`. Both `FormHooks::userRegisterNotifySubmit()` and `CommunityManagerUserCreationForm::submitForm()` now call it.
- Removed `$this->t($roleValue->getLabel())` in `CommunityManagerUserCreationForm` — passing a variable to `t()` is a Drupal anti-pattern that runs untranslatable DB values through the translation system.
- Declared `mukurtu_notifications` as a formal dependency in `mukurtu_core.info.yml` (it was already declared in `mukurtu_protocol`). Replaced `function_exists()` guards with direct calls.
- Added a static request-level cache to `MukurtuManagerUserSelection::getPrivilegedUserIds()`, which previously ran 3 entity queries on every autocomplete keystroke.

## Pre-merge checklist

- [x] I have checked for and if required, created update hooks. *(no schema changes)*
- [x] I have recompiled SCSS if required. *(CSS only, no SCSS changes)*


michael-wynne-wsu:
@nick-deer this should be the last of the re-review of committed code from the past few weeks.

modules/mukurtu_core/src/Hook/FormHooks.php
```diff
@@ -81,12 +83,31 @@ public function formUserRegisterFormAlter(array &$form, FormStateInterface $form
...
-    // Build community options and protocol options grouped by community (site-wide).
-    $communityOptions = [];
-    $protocolsByCommunity = [];
+    // The notify section is only useful to authenticated users with create
+    // access; skip it for anonymous self-registration.
+    $currentAccount = \Drupal::currentUser();
+    if ($currentAccount->isAnonymous()) {
+      return;
+    }
+
```
nick-deer, line 92: I'm not sure this is necessary, since anonymous users cannot access the `/user/register` page.

nick-deer (approved):

modules/mukurtu_core/src/Hook/FormHooks.php
michael-wynne-wsu: @nick-deer If you enable it at `/admin/config/people/accounts` it is. But yikes it's exposed when that is enabled, we didn't test that clearly...

I will look at removing that.

Screenshot 2026-04-29 at 2 52 52 PM

michael-wynne-wsu: @nick-deer never mind, I was looking at that on main. On this PR that notifications section IS properly hidden, so I think that fix might be accurate?

PR #1570, opened 2026-04-27
Index taxonomy term names as fulltext for DB search backend (issue #622)

### LINKED ISSUE #622: What fields are indexed in search?
Getting expected (very limited) search results and I don't think all expected fields are included.

michael-wynne-wsu:
## Summary

- Taxonomy reference fields (keywords, category, people, subject, language, location, publisher) were indexed as `string` in `mukurtu_default_content_index` — usable for browse facets but invisible to the `search_api_fulltext` filter
- The original PR only updated `mukurtu_browse_auto_index`, which is not used by the browse page views; this fix targets the correct index (`mukurtu_default_content_index`)
- Adds parallel `type: text` fields alongside the existing `type: string` fields for taxonomy terms that have facets (keywords, category, people, subject), so faceting is preserved
- Adds new `type: text` fields for language, location, and publisher, which were missing from the index entirely
- Contributor and creator were already indexed as `type: text` and required no changes
- Update hook `mukurtu_browse_update_40003` reimports the index config and marks all content for reindexing on existing sites
- Removes all `__uuid` fields from `mukurtu_browse_auto_index` — they served no purpose for search or faceting and were pushing the index table over MySQL's 64-index-per-table limit; static field count reduced from 52 to 36
- Reduces dynamic key additions to `mukurtu_browse_auto_index` from 3 per taxonomy field to 1 by removing `__name` (string) from `BaseFieldsSearchIndexSubscriber` and removing `mukurtu_browse_auto_index` from `TaxonomyFieldSearchIndexSubscriber` — string/facet fields are defined explicitly in the YAML; UUID indexing is only needed for the Solr backend
- `FieldAvailableForIndexing::indexField()` now skips fields whose ID, type, and property path already match what the YAML defines, avoiding unnecessary index saves during rebuild

Closes #622

## Test plan

- [ ] Run `drush updb` — update hooks `mukurtu_browse_update_40003` and `mukurtu_search_update_40001` fire without errors
- [ ] Run `drush sapi-i mukurtu_default_content_index` to reindex content
- [ ] Search for a keyword term name (e.g. a term used on a DH item) — confirm results appear on `/browse`
- [ ] Search for a language, location, or publisher term name — confirm results appear
- [ ] Confirm browse facets still work (string fields for keywords, category, people, subject are unchanged)
- [ ] Verify no "Too many keys" warnings in the Drupal logs after `drush updb`
- [ ] Verify field list at `/admin/config/search/search-api/index/mukurtu_default_content_index/fields` shows the new text fields

## Pre-merge checklist:


michael-wynne-wsu:
## PR Review: Full Text Search — Taxonomy Term Names (#622) — Updated

**Blast Radius:** LOW — changes contained within `mukurtu_search` module; only `mukurtu_browse_auto_index` (DB backend) is affected. Solr index and browse facets untouched.
**Security:** No findings.
**Tests:** No new tests — pre-existing gap, tracked as `@todo` in `BaseFieldsSearchIndexSubscriber.php`.
**Breaking Changes:** None. Pre-existing field IDs (`node__field_keywords__name`, `node__field_category__name`) are unchanged.

---

### MUST FIX

None.

### SHOULD FIX

None.

---

### TRACKED AS @todo

- **Unit tests** — `BaseFieldsSearchIndexSubscriber.php:74` — `@todo` documents the need for tests covering the taxonomy branch and `field_communities` exclusion.
- **Redundant UUID entries in `ignorecase` processor** — `FieldAvailableForIndexing.php:87` — `@todo` notes that UUID fields in the `ignorecase` list are always lowercase and can be cleaned up in a future update hook.
- **Update hook `@see` link** — `mukurtu_search.install` — `@todo` to add a canonical issue queue URL once available.

---

### LOOKS GOOD

- **Null guard on `getSetting('target_type')`** — now uses `?? ''` to safely handle malformed or in-progress entity reference configs; won't throw on null.
- **Mirrors Solr implementation** — `SolrBaseFieldsSearchIndexSubscriber` already does this exact pattern. DB backend now at parity.
- **Field IDs are stable** — `node__field_keywords__name` and `node__field_category__name` string fields keep the same IDs through the rebuild.
- **`field_communities` correctly excluded** — its target type is not `taxonomy_term`, so the new handler doesn't fire for it.
- **`ignorecase` auto-registration covered on both paths** — `indexField()` adds `text` fields dynamically at runtime; YAML lists them explicitly for new installs.
- **Update hook is correct** — `mukurtu_search_rebuild_index()` clears and rebuilds field definitions using the updated subscriber, then `$index->reindex()` marks all content for reindexing. Numbering follows the `40001` convention.
- **All four processor lists updated consistently** — `html_filter`, `ignorecase`, `tokenizer`, `transliteration` each correctly include the new `__name__text` fields (and `__name` string fields where applicable).
- **No security issues.**

michael-wynne-wsu:
@alexmerrill @nick-deer No real front end changes, just fixing some fields missing from search indices.

modules/mukurtu_search/mukurtu_search.install
```diff
@@ -8,3 +8,16 @@
 function mukurtu_search_install() {
   mukurtu_search_rebuild_index();
 }
+
+/**
+ * Index taxonomy term names as fulltext for search (issue #622).
+ *
+ * @todo Follow Drupal docblock convention and add a @see link to the issue
+ *   queue once a canonical URL is available for this update.
```
nick-deer: Non-blocking: I'm not sure what this means--why would we need a canonical URL for this update? And where would it lead?

nick-deer (approved):
Still willing to approve this in spite of my question regarding the todo in `mukurtu_search.install`

modules/mukurtu_search/mukurtu_search.install
michael-wynne-wsu: Safe to say it's erroneous/slop. I will remove that!

michael-wynne-wsu:
@nick-deer I think this warrants a re-review. It was failing to build properly because of too many keys in the default content index. The changes SEEM fine, but it's a notable code change.

modules/mukurtu_taxonomy/src/EventSubscriber/TaxonomyFieldSearchIndexSubscriber.php
```diff
@@ -38,7 +38,7 @@ public function __construct(MessengerInterface $messenger) {
    *   Response event.
    */
   public function indexTaxonomyField(FieldAvailableForIndexing $event) {
-    $indexes = ['mukurtu_browse_auto_index', 'mukurtu_default_solr_index'];
```
nick-deer, line 41: I think removing `mukurtu_browse_auto_index` here is a mistake. See the above comment:

```
* This is responsible for indexing the UUIDs of taxonomy fields. This is how
* we display/browse referenced content on the canonical taxonomy term pages.
```
The referenced content view may break on canonical taxonomy term pages if we make this change.

nick-deer (changes requested):

modules/mukurtu_taxonomy/src/EventSubscriber/TaxonomyFieldSearchIndexSubscriber.php
michael-wynne-wsu: Interesting, I wonder if that's because I instructed Claude not to work on solr. Happy to scrap that edit.

modules/mukurtu_taxonomy/src/EventSubscriber/TaxonomyFieldSearchIndexSubscriber.php
```diff
@@ -38,7 +38,7 @@ public function __construct(MessengerInterface $messenger) {
...
+    $indexes = ['mukurtu_default_solr_index'];
```
michael-wynne-wsu:
```suggestion
    $indexes = ['mukurtu_browse_auto_index', 'mukurtu_default_solr_index'];
```

nick-deer (approved):

PR #1572, opened 2026-04-28
Move update hooks out of mukurtu.install into owning modules

michael-wynne-wsu:
## Summary

- Update hooks belong in the module responsible for the config/entities they touch, not in the root profile install file
- Moves four hooks from `mukurtu.install` to their owning modules:

| Old | New |
|---|---|
| `mukurtu_update_40001` (roundtrip_manager role label) | `mukurtu_import_update_40025` |
| `mukurtu_update_40002` (audio media view displays) | `mukurtu_media_update_40002` |
| `mukurtu_update_40003` (digital heritage featured view) | `mukurtu_digital_heritage_update_40003` |
| `mukurtu_update_40004` (message digest cleanup) | `mukurtu_notifications_update_40004` |

## Test plan

- [ ] Run `drush updatedb` on a site that has not yet run these hooks — all four should execute from their new locations
- [ ] Confirm `mukurtu.install` no longer contains any `mukurtu_update_NNNNN` functions (only `mukurtu_install()`)
- [ ] Confirm no functional change to what each hook does


michael-wynne-wsu:
Noticed some of my earlier update hooks slipped into the wrong file. This corrects that, and shouldn't really impact anything.

modules/mukurtu_notifications/mukurtu_notifications.install
```diff
@@ -53,4 +53,20 @@ function mukurtu_notifications_update_40003(): void {
...
+ */
+function mukurtu_notifications_update_40004(): void {
+  $config_factory = \Drupal::configFactory();
+  $actions = [
+    'system.action.message_digest_interval.email_user.daily',
+    'system.action.message_digest_interval.email_user.immediate',
+    'system.action.message_digest_interval.email_user.weekly',
+  ];
+  foreach ($actions as $action) {
+    $config_factory->getEditable($action)->delete();
```
nick-deer, line 70: Non-blocking: found a potential inefficiency. If the config passed to `getEditable()` does not exist, `getEditable()` will actually create one and then return it. For repeat runs of this hook, if the config entities for the `$actions` have already been deleted, this hook will re-create them, only to delete them right after with `delete()`.

nick-deer (approved):

modules/mukurtu_notifications/mukurtu_notifications.install
michael-wynne-wsu: Since we are probably going to clear out the stale hooks after we release the initial 4.0, this is probably fine to leave. But I'm happy to try and clean it up if you prefer.

PR #1578, opened 2026-04-29
Allow media reference by name or filename in formatted text embeds

### LINKED ISSUE #1513: Importing media assets into full HTML/rich text fields
- Currently the rich text/full HTML fields require a UUID to include a media asset in the field. 
- Is there a way to import media assets as part of a rich text field by using another identifier (filename, etc.)?

michael-wynne-wsu:
Closes #1513

## Summary

- Adds a new migrate process plugin `mukurtu_resolve_media_embeds` that runs on formatted text field values during CSV import
- Supports two new `drupal-media` tag attributes as alternatives to `data-entity-uuid`:
  - `data-entity-name="My Image"` — resolves by media entity label
  - `data-entity-filename="image.jpg"` — resolves by the filename of the media's source file
- Both attributes are resolved to a proper UUID and removed before the content is saved; existing `data-entity-uuid` tags are unaffected
- Unresolvable references (not found, ambiguous) log a migration message and leave the tag unchanged so the import still completes

## Supported in CSV sheets

| Syntax | Lookup | Attributes |
|---|---|---|
| `{{media:My Image}}` or `{{media:photo.jpg}}` | Name, then filename fallback | Default view mode |
| `[media name="My Image"]` | Name only | `view-mode`, `align`, `caption`, `alt` |
| `[media filename="photo.jpg"]` | Filename only | `view-mode`, `align`, `caption`, `alt` |
| `<drupal-media data-entity-name="...">` | Name only | Full tag control |
| `<drupal-media data-entity-filename="...">` | Filename only | Full tag control |

## Test plan

- [x] `testResolveByName` — `data-entity-name` resolves to correct UUID
- [x] `testResolveByFilename` — `data-entity-filename` resolves to correct UUID
- [x] `testExistingUuidPassesThrough` — existing UUID tags are unchanged
- [x] `testUnknownNameLeavesTagUnchanged` — unresolvable name leaves tag intact and migration completes

Run: `vendor/bin/phpunit --bootstrap web/core/tests/bootstrap.php web/profiles/mukurtu/modules/mukurtu_import/tests/src/Kernel/ImportFormattedTextMediaEmbedTest.php`

### DIFF
--- a/modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php
+++ b/modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php
@@ -57,6 +57,13 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 class ResolveMediaEmbeds extends ProcessPluginBase implements ContainerFactoryPluginInterface {
 
   /**
+   * Cached map of media type ID to source field name, built on first use.
+   *
+   * @var array<string, string>
+   */
+  protected array $sourceFieldNames = [];
+
+  /**
    * Constructs a ResolveMediaEmbeds object.
    */
   public function __construct(
@@ -171,10 +178,18 @@ class ResolveMediaEmbeds extends ProcessPluginBase implements ContainerFactoryPl
         }
         catch (MigrateException $e) {
           $migrate_executable->saveMessage($e->getMessage());
+          // Replace the intermediate attribute with data-entity-name so the
+          // stored tag is consistent with unresolved data-entity-name tags.
+          $element->setAttribute('data-entity-name', $identifier);
+          $element->removeAttribute('data-entity-identifier');
+          $resolved = TRUE;
           continue;
         }
         if (!$uuid) {
           $migrate_executable->saveMessage(sprintf('Could not find a media entity matching "%s".', $identifier));
+          $element->setAttribute('data-entity-name', $identifier);
+          $element->removeAttribute('data-entity-identifier');
+          $resolved = TRUE;
           continue;
         }
         $element->setAttribute('data-entity-uuid', $uuid);
@@ -319,20 +334,21 @@ class ResolveMediaEmbeds extends ProcessPluginBase implements ContainerFactoryPl
 
     $file_ids = array_values($file_ids);
     $media_storage = $this->entityTypeManager->getStorage('media');
-    $media_type_storage = $this->entityTypeManager->getStorage('media_type');
+
+    if (!$this->sourceFieldNames) {
+      foreach ($this->entityTypeManager->getStorage('media_type')->loadMultiple() as $media_type) {
+        $source_field_def = $media_type->getSource()->getSourceFieldDefinition($media_type);
+        if ($source_field_def) {
+          $this->sourceFieldNames[$media_type->id()] = $source_field_def->getName();
+        }
+      }
+    }
 
     $matches = [];
-    foreach ($media_type_storage->loadMultiple() as $media_type) {
-      $source_plugin = $media_type->getSource();
-      $source_field_def = $source_plugin->getSourceFieldDefinition($media_type);
-      if (!$source_field_def) {
-        continue;
-      }
-      $source_field_name = $source_field_def->getName();
-
+    foreach ($this->sourceFieldNames as $type_id => $source_field_name) {
       $results = $media_storage->getQuery()
         ->accessCheck(FALSE)
-        ->condition('bundle', $media_type->id())
+        ->condition('bundle', $type_id)
         ->condition($source_field_name, $file_ids, 'IN')
         ->execute();
 
--- a/modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php
+++ b/modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php
@@ -1,0 +1,230 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\mukurtu_import\Plugin\migrate\process;
+
+use DOMDocument;
+use DOMXPath;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\migrate\MigrateException;
+use Drupal\migrate\MigrateExecutableInterface;
+use Drupal\migrate\ProcessPluginBase;
+use Drupal\migrate\Row;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Resolves drupal-media embed tags using media name or filename attributes.
+ *
+ * During CSV import, users can reference media assets in formatted text fields
+ * using human-readable identifiers instead of UUIDs:
+ *
+ * By media entity name:
+ * @code
+ * <drupal-media data-entity-type="media" data-entity-name="My Image" data-view-mode="default">
+ * @endcode
+ *
+ * By source file filename:
+ * @code
+ * <drupal-media data-entity-type="media" data-entity-filename="my-image.jpg" data-view-mode="default">
+ * @endcode
+ *
+ * Both attributes are resolved to a proper data-entity-uuid and removed from
+ * the tag before the content is saved.
+ *
+ * @MigrateProcessPlugin(
+ *   id = "mukurtu_resolve_media_embeds",
+ *   handle_multiples = FALSE
+ * )
+ */
+class ResolveMediaEmbeds extends ProcessPluginBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * Constructs a ResolveMediaEmbeds object.
+   */
+  public function __construct(
+    array $configuration,
+    $plugin_id,
+    $plugin_definition,
+    protected EntityTypeManagerInterface $entityTypeManager,
+  ) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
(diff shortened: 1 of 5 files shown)


christa-kavanagh-wsu:
Two of them did not find the media asset: `[media name="My Image"]` and `<drupal-media data-entity-name="...">`. The rest of them worked as expected.

michael-wynne-wsu:
> Two of them did not find the media asset: `[media name="My Image"]` and `<drupal-media data-entity-name="...">`. The rest of them worked as expected.

I think this is due to smart quotes getting inserted. working to normalize those.

christa-kavanagh-wsu (approved):

michael-wynne-wsu (commented):
## Accessibility

**No WCAG concerns.** This is a pure backend migration process plugin — no HTML rendered to end users, no UI elements, no ARIA impact.

---

## PR Review

### Bug — `MediaEmbed.php` hardcodes `libxml_use_internal_errors(FALSE)`

`modules/mukurtu_migrate/src/Plugin/migrate/process/MediaEmbed.php`

The new change always resets to `FALSE` after parsing:

```php
libxml_use_internal_errors(TRUE);
$doc->loadHTML('<?xml encoding="utf-8"?>' . $value, LIBXML_HTML_NODEFDTD);
libxml_use_internal_errors(FALSE);  // ← BUG
```

`ResolveMediaEmbeds` does this correctly by saving and restoring:

```php
$previous = libxml_use_internal_errors(TRUE);
// ...
libxml_use_internal_errors($previous);
```

If any caller previously set `libxml_use_internal_errors(TRUE)`, this silently clears it. Should use the same save/restore pattern.

---

### Bug — Unresolved `{{media:...}}` shortcodes leave intermediate `data-entity-identifier` in saved content

`modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php`

When a `{{media:VALUE}}` shortcode cannot be resolved, the flow is:

1. `convertShortcodes()` converts `{{media:VALUE}}` → `<drupal-media data-entity-identifier="VALUE">` (intermediate)
2. DOM lookup fails, `continue` is called, `$resolved` stays `FALSE`
3. Because `$resolved === FALSE`, `saveHTML()` is skipped and `$value` — already containing `<drupal-media data-entity-identifier="...">` from step 1 — is returned as-is

`data-entity-identifier` is not a recognized Drupal attribute; Drupal's media embed filter will silently fail to render this tag. Compare: unresolvable `data-entity-name` tags leave the original `<drupal-media data-entity-name="...">` in place, which is at least recognizable. The two failure modes are inconsistent.

Suggested fix: in the unresolved `data-entity-identifier` branch, either restore the original `{{media:VALUE}}` text or swap to `data-entity-name` as the preserved attribute.

---

### Performance — `resolveByFilename` runs one query per media type

`modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php:290-318`

The method loads all media types and executes one entity query per type to find which media entities reference the file. For sites with many media types or documents with many embeds, this is an N+1 pattern per embed.

Low severity (import-time only, not page load), but worth noting. The source field mapping per media type could at minimum be cached for the lifetime of the plugin instance.

---

### Missing test — ambiguous name/filename

The ambiguous-match code path (`count($results) > 1` → `MigrateException`) is documented in error messages but not covered by a test. Suggested additions:

- `testAmbiguousNameThrowsMigrateMessage` — two media entities with the same name
- `testAmbiguousFilenameThrowsMigrateMessage` — two files/media with the same filename

---

### Minor — `&nbsp;` in generated `drupal-media` tags

`convertShortcodes()` produces `<drupal-media ...>&nbsp;</drupal-media>`. Drupal's media embed filter generally expects empty or self-closing tags. Worth verifying that the `&nbsp;` doesn't get passed through to the rendered output.

michael-wynne-wsu:

Link:
- https://pr1578-lkbf5vxz2xyqp88uwdu0hrjeh3ydy29o.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69f248827e27abc39c402d02

michael-wynne-wsu:
All four issues resolved and pushed. Here's a summary of what was done:

**1. `MediaEmbed.php` — libxml state leak** (`modules/mukurtu_migrate/src/Plugin/migrate/process/MediaEmbed.php:117`)
Save and restore `$previous` instead of hardcoding `FALSE`.

**2. `ResolveMediaEmbeds.php` — unresolved shortcode remnant** (`modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php:173-181`)
When a `{{media:...}}` shortcode can't be resolved, the intermediate `data-entity-identifier` attribute is now swapped to `data-entity-name` before writing back via `saveHTML()`. Unresolvable shortcodes now leave a consistent `<drupal-media data-entity-name="...">` tag rather than the opaque intermediate attribute.

**3. `ResolveMediaEmbeds.php` — N+1 query performance** (`modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php:64`)
Added a `$sourceFieldNames` cache property. Media type storage is now loaded only once per plugin instance rather than once per embed processed.

**4. Missing ambiguity tests** (`modules/mukurtu_import/tests/src/Kernel/ImportFormattedTextMediaEmbedTest.php`)
Added three new tests: `testAmbiguousNameLogsMessageAndLeavesTagUnchanged`, `testAmbiguousFilenameLogsMessageAndLeavesTagUnchanged`, and `testUnresolvableCurlyBraceShortcodeLeavesSafeAttribute`. All 10 tests pass.

modules/mukurtu_import/src/Plugin/migrate/process/ResolveMediaEmbeds.php
```diff
@@ -0,0 +1,371 @@
...
+use Drupal\migrate\MigrateException;
+use Drupal\migrate\MigrateExecutableInterface;
+use Drupal\migrate\ProcessPluginBase;
+use Drupal\migrate\Row;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Resolves media embed references in formatted text fields during import.
+ *
+ * Three syntaxes are supported:
```
nick-deer: Says only 3 syntaxes are supported, but lists 4

michael-wynne-wsu:
```suggestion
 * Four syntaxes are supported:
```

modules/mukurtu_import/tests/src/Kernel/ImportFormattedTextMediaEmbedTest.php
```diff
@@ -0,0 +1,476 @@
...
+    $this->assertEquals(MigrationInterface::RESULT_COMPLETED, $result);
+
+    $body = $this->entityTypeManager->getStorage('node')->load($this->node->id())->get('field_body')->value;
+    $this->assertStringContainsString('data-entity-uuid="' . $this->mediaByName->uuid() . '"', $body);
+  }
+
+  /**
+   * Tests that an ambiguous name logs a message and leaves the tag unchanged.
+   */
+  public function testAmbiguousNameLogsMessageAndLeavesTagUnchanged(): void {
```
nick-deer, line 347: This test does check that the tag remains unchanged, but it does not check that a message was logged. I think either the title of this test change to reflect that, or an assertion should be added that actually checks that the expected error message was logged.

modules/mukurtu_import/tests/src/Kernel/ImportFormattedTextMediaEmbedTest.php
```diff
@@ -0,0 +1,476 @@
...
+
+    $body = $this->entityTypeManager->getStorage('node')->load($this->node->id())->get('field_body')->value;
+    $this->assertStringNotContainsString('data-entity-uuid=', $body);
+    $this->assertStringContainsString('data-entity-name="My Named Asset"', $body);
+  }
+
+  /**
+   * Tests that an ambiguous filename logs a message and leaves the tag unchanged.
+   */
+  public function testAmbiguousFilenameLogsMessageAndLeavesTagUnchanged(): void {
```
nick-deer, line 392: Same issue as the preceding test

nick-deer (changes requested):
Just a few minor things to change, but overall the PR looks great!

michael-wynne-wsu:
> Just a few minor things to change, but overall the PR looks great!

Fantastic, I will get those cleaned up and ping for re review when done!

michael-wynne-wsu:
Hopefully this actually has the correct test assertion

joelle-birano-wsu:
Everything else is working except for `data-entity-name` and `data-entity-filename.` The upload for both is successful, the media asset is in the library. The media asset sheet and the DH sheet don't seem to notice each other at all. The image doesn't display. 

Screenshot 2026-04-30 164616

michael-wynne-wsu:
@joelle-birano-wsu can we see the plain text you tested with?

joelle-birano-wsu:
> @joelle-birano-wsu can we see the plain text you tested with?

Here are the sheets for the DH items I uploaded. 
[Media Asset Test DH_5.csv](https://github.com/user-attachments/files/27279320/Media.Asset.Test.DH_5.csv)
[Media Asset Test DH_6.csv](https://github.com/user-attachments/files/27279337/Media.Asset.Test.DH_6.csv)

michael-wynne-wsu:
@joelle-birano-wsu Looking at the code that's generated by embedding in site (below, ignore the uuid but see the rest), you might need to include `data-entity-type="media"` in there as well?

`<drupal-media data-entity-type="media" data-entity-uuid="c7289630-de78-470c-b3bb-48764beb9d9e">&nbsp;</drupal-media>`

michael-wynne-wsu:
@joelle-birano-wsu we can also just omit that format from the documentation, like Christa suggested. I don't think it needs to block this.

joelle-birano-wsu:
> @joelle-birano-wsu we can also just omit that format from the documentation, like Christa suggested. I don't think it needs to block this.

Yeah I think that's best. It seems like the least user-friendly option.

joelle-birano-wsu (approved):

nick-deer (approved):

PR #1580, opened 2026-04-29
Dashboard panels cleanup (#1577)

### LINKED ISSUE #1577: Clean up dashboard sections
Dashboard clean-up notes:
- Mukurtu managers - do they need Migration?
- Remove migration for all users except administrators
- Do all users need to view site-wide Local Contexts projects?

By-role breakdown:
- Community managers 
  - [ ] Remove migration
- Protocol stewards
  - [ ] Remove Migration
- Language stewards
  - [ ] Remove Migration
- Contributors
  - [ ] Remove Migration
- Curators
  - [ ] Remove Migration
- Community record stewards
  - [ ] Remove migration
- Community members, community affiliates, protocol members, protocol affiliates, and authenticated users who don't have community or protocol roles
  - [ ] Remove Site settings
  - [ ] Remove Migration
  - [ ] Remove site information

michael-wynne-wsu:
## Summary

- [ ] Only administrators and mukurtu managers can see the "Site settings" block (this was exposing "view site-wide local contexts projects" to all users which was unnecessary).
   - [ ] The LC directory is still accessible as before, I anticipate people will be providing front end links when they use it.
- [ ] Only administrators can see the "migration" block (since that should only be run by UID one it seems to make sense).

## Details

- Adds a layout builder event subscriber (`DashboardBlockAccessSubscriber`) that intercepts `SECTION_COMPONENT_BUILD_RENDER_ARRAY` at priority 200 (before the standard subscriber at 100)
- The **Migration** panel (`system_menu_block:dashboard-migration`) is now hidden for all roles except `administrator`
- The **Site settings** panel (`system_menu_block:dashboard-site-settings`) is now hidden for all roles except `administrator` and `mukurtu_manager`
- `user.roles` cache context is applied when access is denied, so the page cache varies correctly per role

## Implementation notes

`hook_block_access` was not usable here — it only fires for block config entities, but the dashboard renders block plugins directly via layout builder's `SectionComponent::toRenderArray()`. The event subscriber approach intercepts the layout builder render pipeline before the block is built.

## Test plan

- [ ] Log in as `administrator` → both Migration and Site settings panels visible
- [ ] Log in as `mukurtu_manager` → Site settings panel visible, Migration panel hidden
- [ ] Log in as any other authenticated user → neither panel visible
- [ ] Confirm `/local-contexts/projects` remains publicly accessible (no route change)
- [ ] `ddev drush cr` after deploy (new service registered)

Closes #1577

### DIFF
--- a/config/install/dashboards.dashboard.mukurtu_dashboard.yml
+++ b/config/install/dashboards.dashboard.mukurtu_dashboard.yml
@@ -12,7 +12,7 @@ third_party_settings:
       restricted_categories: {  }
       allowed_layouts: {  }
 _core:
-  default_config_hash: _O8NOA9JSEt9Zy3WdKUs4xOL7aHvfwUmBz1oifvnvEw
+  default_config_hash: rVGaJigae-beV372fO5BQUxNElb3GXG-yxGn7pPVzko
 id: mukurtu_dashboard
 admin_label: 'Mukurtu Dashboard'
 category: ''
@@ -25,19 +25,6 @@ sections:
       label: 'main content'
       context_mapping: {  }
     components:
-      0d6f9095-ddc7-4e04-b102-a47eab86f9dc:
-        uuid: 0d6f9095-ddc7-4e04-b102-a47eab86f9dc
-        region: three
-        configuration:
-          id: 'views_block:content_recent-block_1'
-          label: ''
-          label_display: visible
-          provider: views
-          context_mapping: {  }
-          views_label: ''
-          items_per_page: null
-        weight: 3
-        additional: {  }
       177c8838-ecba-43f7-9acb-813187ae16f3:
         uuid: 177c8838-ecba-43f7-9acb-813187ae16f3
         region: three
@@ -66,34 +53,6 @@ sections:
           expand_all_items: false
         weight: 7
         additional: {  }
-      7ea8f4fe-f816-4b62-80ed-597ea6d8d39a:
-        uuid: 7ea8f4fe-f816-4b62-80ed-597ea6d8d39a
-        region: two
-        configuration:
-          id: 'system_menu_block:dashboard-site-info'
-          label: 'Site information'
-          label_display: visible
-          provider: system
-          context_mapping: {  }
-          level: 1
-          depth: null
-          expand_all_items: false
-        weight: 8
-        additional: {  }
-      4baf6b8d-6deb-4742-aac9-5a4ca97be575:
-        uuid: 4baf6b8d-6deb-4742-aac9-5a4ca97be575
-        region: one
-        configuration:
-          id: 'system_menu_block:dashboard-users'
-          label: Users
-          label_display: visible
-          provider: system
-          context_mapping: {  }
-          level: 1
-          depth: null
-          expand_all_items: false
-        weight: 6
-        additional: {  }
       8e52ecf9-f13d-4974-b4e5-9ecff8af170a:
         uuid: 8e52ecf9-f13d-4974-b4e5-9ecff8af170a
         region: one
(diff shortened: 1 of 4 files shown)


christa-kavanagh-wsu (commented):
All the sections that were supposed to be hidden now are. 

For all roles except admin and Mukurtu manager, I would switch the position of the Site information and Recent content sections.

michael-wynne-wsu:
> All the sections that were supposed to be hidden now are.
> 
> For all roles except admin and Mukurtu manager, I would switch the position of the Site information and Recent content sections.

I'm not sure we can rearrange sections per role. At least not without a good amount of work. is there a different all-user arrangement that works instead? Maybe "my account" in the first column instead?

christa-kavanagh-wsu:
> > All the sections that were supposed to be hidden now are.
> > For all roles except admin and Mukurtu manager, I would switch the position of the Site information and Recent content sections.
> 
> I'm not sure we can rearrange sections per role. At least not without a good amount of work. is there a different all-user arrangement that works instead? Maybe "my account" in the first column instead?

I like the "my account" settings being on the right. I think "recent content would work in the center for admin and Mukurtu manager roles and just shift the other sections to the right-hand column below Roundtrip?

christa-kavanagh-wsu:
Recommended order (based on admin account dashboard)

|Left-hand column|Center|Right-hand column|
|---|---|---|
|Communities, cultural protocols, categories|Recent Content|My Account|
|Users|Content settings|Site Information|
|Content|Migration|Roundtrip|
|Media| | Look and Feel|
|Taxonomies| |Site Settings|

joelle-birano-wsu:
That all looks good to me. I like @christa-kavanagh-wsu's arrangement. The only thing I'm debating a little is whether content should be above users...but for admins it's probably more important to access users more?

michael-wynne-wsu:
I made some tweaks, here's how I currently have it...

Edit-layout-for-Mukurtu-Dashboard-Mukurtu-CMS-04-30-2026_12_21_PM

christa-kavanagh-wsu:
> That all looks good to me. I like @christa-kavanagh-wsu's arrangement. The only thing I'm debating a little is whether content should be above users...but for admins it's probably more important to access users more?

I'm on the fence about that one, too, but as an admin I would rather have easier access to users.

joelle-birano-wsu:
> I'm on the fence about that one, too, but as an admin I would rather have easier access to users.

I can see that. And for other content creating roles it'll be up higher...

christa-kavanagh-wsu (approved):

michael-wynne-wsu:
**Accessibility & PR Review — `1577-dashboard-role-restrictions`**

No WCAG 2.2 AA violations introduced. No security vulnerabilities. One correctness issue fixed, one update hook added.

**Fixed in this commit:**

- **Cache context gap** — `user.roles` was only attached to the event when access was denied. Allowed renders for the two restricted plugins (`dashboard-migration`, `dashboard-site-settings`) were cached without role context, risking an admin's cached render being served to a lower-privileged user. Now attached on both paths.
- **Redundant role lookups** — Added an early return for non-restricted plugins so `getRoles()` is never called for the 11 unrestricted dashboard components.
- **Update hook `mukurtu_core_update_40014`** — Dashboard block reorganization (admin panels grouped into region three, recent content moved to region two) was only in `config/install/` and would not apply to existing sites. The hook applies the 10 region/weight changes by component UUID.

**Already good:** strict `in_array` comparison, correct `@current_user` injection, event priority 200 correctly pre-empts the default block build subscriber, no pre-existing config entities orphaned.

joelle-birano-wsu (approved):

nick-deer (approved):

PR #1595, opened 2026-05-06
Update site-wide user list display for issue #1552

### LINKED ISSUE #1552: Update site wide user list displays
- Need to update `/admin/people` and `/admin/people/list`.
- At top of both pages, there should be view-filters:
   - Name or email contains (as on `/admin/people`)
   - Status (as on `/admin/people`)
   - Site role (as on `/admin/people`, just named "Site role" instead of "Role")
   - Communities (new)
   - "Filter" button
- The field order on both pages should be:
   - Username
   - Display name
   - Site roles (renamed from "Roles" to "Site roles")
      - As a comma separated list
   - Communities
      - As links in a comma separated list
   - Status
   - Member since
   - Last access
   - Operations 
      - Edit
      - Block or Approve, following current logic
      - Memberships
- The existing bulk actions menu should not be changed.

michael-wynne-wsu:
Closes #1552 

- The core functionality is here from previous work, this is minor cleanup. 

## `/admin/people/list`

- Added a "Site roles" filter.
- Ensured that users with no additional site role display "Authenticated user".

## `/admin/people`

- Updated all filters and table fields to match `/admin/people/list`.

## Testing (on both forms)

- [ ] Do all filters and search work?
- [ ] Do all per-user operations work?
- [ ] Do all bulk operations work?
- [ ] Does `/admin/people/list` work as expected as a community manager as well? (with no Administrator/Mukurtu manager role)

Pre-merge checklist:


joelle-birano-wsu:
- Maybe a separate issue, but I tried blocking myself as admin (with no backups) and I could, which is how I got locked out. I wonder if we shouldn't allow that if there are no other admin accounts.


- As a community manager - if a user not in my community is blocked, I cannot unblock them, but I CAN add them to my community, and once I do that, I can approve their account and they can be active again. Do we want that? I'm thinking about situations where someone may have been blocked by someone else - is that a potential loophole?

- I ran into some caching issues - removing users from a community was not reflecting on admin/people/list or /admin/people. Fixed with cache clear.


- If I add a user to my community from the Operations menu, I still get that "assign protocol roles" button (I'm ONLY a community manager). Not sure if we're leaving that or  be addressed somewhere else (I know we discussed it before but couldn't find that conversation) so sorry if that's redundant. Clicking "assign protocol role" does redirect to admin/people/list though which is good.

Screenshot 2026-05-06 at 4 19 16 PM


Screenshot 2026-05-06 at 4 21 32 PM


- These are slightly different - not sure if it matters:

Screenshot 2026-05-06 at 1 45 04 PM
Screenshot 2026-05-06 at 1 44 56 PM

michael-wynne-wsu:
> Maybe a separate issue, but I tried blocking myself as admin (with no backups) and I could, which is how I got locked out. I wonder if we shouldn't allow that if there are no other admin accounts.

I'm guessing that was through the bulk actions right? We probably should disable self-blocking.

> As a community manager - if a user not in my community is blocked, I cannot unblock them, but I CAN add them to my community, and once I do that, I can approve their account and they can be active again. Do we want that? I'm thinking about situations where someone may have been blocked by someone else - is that a potential loophole?

Yeah I was weighing that earlier and I had the same gut reaction that it was a loophole. But at the end of the day I think it's more about sensible management and not something we need to enforce on our end. Open to discuss but I think it's a case where we can't reasonably spec a one size fits all solution.

> I ran into some caching issues - removing users from a community was not reflecting on admin/people/list or /admin/people. Fixed with cache clear.

That's known in general. At one point I cleared that up but with the new folks stuff is probbably missed. Can you make a new issue to assess user role caching after we're done the rest of the user management changes.

> If I add a user to my community from the Operations menu, I still get that "assign protocol roles" button (I'm ONLY a community manager). Not sure if we're leaving that or be addressed somewhere else (I know we discussed it before but couldn't find that conversation) so sorry if that's redundant. Clicking "assign protocol role" does redirect to admin/people/list though which is good.

Personally I think it's a fringe enough use case that I don't want to address it. I'm not opposed to including it in post launch work but I think it's pretty low priority.

> These are slightly different - not sure if it matters:

Thought I caught all those. Will fix!

michael-wynne-wsu:
1. Self-blocking is prevented on both pages.
2. "Site role" filter label is updated.
3. After removing users, cache is cleared.

@joelle-birano-wsu I think those are the main ones. We can discuss the others, but again, I don't necesarily think we need to address them.

michael-wynne-wsu:

Link:
- https://pr1595-jymc5fw1ispztsspohovbwr3oxumnnop.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69fb7e0f703fa3bab1e63561

christa-kavanagh-wsu:
I didn't notice any of the problems Joelle was tracking above; almost everything else looks good and I was able to apply all of the user batch edit options. 

I was wondering about the "Require email confirmation" toggle for cancelling user accounts: you can apply it even if the user does not have a registered email address, and it looks like it goes ahead and cancels the user account anyway even without an email confirmation. Is this intended? Is this toggle a Drupal standard that is best if we don't remove?
image

michael-wynne-wsu:
That's stock drupal stuff, I don't think it's a priority to modify at this point. We may already have an open issue for clarifying the cancel account settings.

christa-kavanagh-wsu (approved):
This works as expected so far as I can tell.

joelle-birano-wsu:
> 1. Self-blocking is prevented on both pages.
> 2. "Site role" filter label is updated.
> 3. After removing users, cache is cleared.
> 
> @joelle-birano-wsu I think those are the main ones. We can discuss the others, but again, I don't necesarily think we need to address them.

@michael-wynne-wsu Agreed! Yes I understand not prioritizing/addressing the other issues. Functionality isn't being compromised so that's fine with me!

michael-wynne-wsu:
Great, getting this ready for code review

michael-wynne-wsu:
## PR Review: User List Display Improvements - `issue/1552-user-list-displays`

**Blast Radius:** MEDIUM - admin-facing only; touches the two user admin views (`/admin/people` and `/admin/mukurtu/people`), VBO bulk actions, and a new Views filter plugin. No public-facing pages are affected.
**Tests:** No new tests. Coverage delta: 0.
**Breaking Changes:** 1 found (missing update hook for `mukurtu_people` view).

---

### MUST FIX (Blocking)

**1. Missing update hook for `mukurtu_people` view changes**
`modules/mukurtu_protocol/mukurtu_protocol.install`

The last hook that re-imports `mukurtu_people` is `40006`. This branch makes significant changes to that view  l new columns (Site Roles now visible, Communities, Status, Member since, Last access, Operations), column reordering, and two new exposed filters (Communities, Site role). Existing sites that have already run `40006` will never see these changes without a new update hook.

Add `update_40010` that re-imports `views.view.mukurtu_people.yml`, following the same pattern as `40006` and `40008`.

---

**2. Update hooks 40008 and 40009 are identical**
`modules/mukurtu_protocol/mukurtu_protocol.install:290-305`

Both hooks parse the same YAML file and write the same config. A fresh install via `40008` already contains the "Site role" label change, making `40009` a no-op for new installs and a wasteful duplicate write for existing sites. Either remove `40009` and fold its intent into `40008`'s docblock, or make `40009` do a targeted surgical config update (e.g., use `$config->set('display.default.display_options.filters.roles_target_id.expose.label', 'Site role')->save()`). The full-overwrite approach is fine when intentional, but two identical hooks in a row signals an incomplete refactor.

---

### SHOULD FIX (Non-blocking)

**3. Disabled bulk checkbox has no accessible name**
`modules/mukurtu_core/src/Hook/FormHooks.php:64-66`

The `disableSelfBulkFormCheckbox` callback sets `#disabled = TRUE` but gives screen reader users no context for why the control is non-interactive. Add a `title` attribute (announced by most screen readers for disabled controls):

```php
$element[$key]['#disabled'] = TRUE;
$element[$key]['#attributes']['title'] = $this->t('You cannot apply bulk actions to your own account.');
```

*(This was also flagged by the WCAG audit as the one required fix.)*

---

**4. `user_roles` is not a real Drupal cache tag**
`modules/mukurtu_protocol/src/Entity/Protocol.php:384`, `modules/mukurtu_protocol/src/Plugin/Action/MukurtuDeleteOgMembershipAction.php:88`

```php
Cache::invalidateTags(["user:{$account->id()}", "user_roles"]);
```

Drupal core does not use a `user_roles` tag. Role config is tagged as `config:user.role.{role_name}`. The `user_roles` tag here will only invalidate caches that were explicitly tagged with that string - which is likely nothing in a standard install. The `user:{uid}` tag is correct and sufficient for invalidating per-user cached Views rows. Remove `"user_roles"` from both call sites, or replace it with the specific role config tags if broader invalidation is needed.

---

### SUGGESTIONS

**5. Brittle bulk form key decoding**
`modules/mukurtu_core/src/Hook/FormHooks.php:62-63`

The base64+JSON format of `#return_value` is an internal implementation detail of Drupal's `BulkForm` plugin. The docblock acknowledges the format but the code still silently breaks if it ever changes. Consider adding a guard:

```php
$key_parts = json_decode(base64_decode($value), TRUE);
if (!is_array($key_parts) || !isset($key_parts[1])) {
  continue;
}
```

This at least prevents a PHP notice/error on malformed values.

---

**6. `UserCommunityFilter::getValueOptions()` loads all communities**
`modules/mukurtu_protocol/src/Plugin/views/filter/UserCommunityFilter.php:32-33`

`loadMultiple()` with no arguments loads every community entity. Fine at typical Mukurtu scale, but worth noting as a potential performance concern for large deployments. If this becomes an issue, consider `loadMultiple(NULL)` with a `getQuery()` approach or adding a result count check.

---

**7. `MukurtuUnblockUserAction` self-unblock guard is logically unreachable**
`modules/mukurtu_protocol/src/Plugin/Action/MukurtuUnblockUserAction.php:63-66`

A blocked user cannot be logged in, so the current user can never be the target of an unblock action against themselves. The guard is defensive and harmless, but the comment "Users cannot unblock themselves via bulk action" may be confusing to future readers. Worth a brief inline note explaining it's a defensive guard.

---

### LOOKS GOOD

- **Defense-in-depth on self-block**  - three independent layers: `entity_presave` fallback, VBO `access()` check, and form `#after_build` checkbox disable. Solid.
- **`entityOperation` access check** - `$url->access()` is checked before returning the Memberships link. Correct.
- **`UserCommunityFilter` query** - The subquery on `og_membership` is properly parameterized (no string interpolation). The `state = active` condition is the right filter.
- **`user_admin_people` view ownership** - Taking full ownership of this core view via `config/install` + update hook is the standard Drupal distribution pattern. Well-executed.
- **`UserRoles` fallback label** - Changing "None" > "Authenticated user" is a meaningful improvement for clarity.
- **Cache tag on `Protocol::removeMember()`** - The `user:{uid}` invalidation is the right place to put this; previously the Views list would show stale community memberships after bulk removal.

---

**Summary:** Two blockers - a missing update hook for `mukurtu_people` (existing sites won't get the new columns/filters), and duplicate update hooks 40008/40009. The accessibility fix on the disabled checkbox is required per the WCAG audit. Everything else is solid.

michael-wynne-wsu:
All six issues resolved:

1. **`update_40010` added**  - re-imports `mukurtu_people` view so existing sites get the new columns and filters.
2. **`update_40009` fixed** - now does a targeted `->set()` on just the label instead of a full overwrite identical to 40008.
3. **Accessibility fix** - disabled checkbox now has a `title` attribute explaining why it's non-interactive.
4. **`user_roles` cache tag removed** - both `Protocol::removeMember()` and `MukurtuDeleteOgMembershipAction::execute()` now only invalidate the correct `user:{uid}` tag.
5. **Bulk form key guard added** - `disableSelfBulkFormCheckbox` now validates the decoded array before accessing index 1.
6. **Unblock comment clarified** - explains the guard is defensive for an edge case, not a normal scenario.

michael-wynne-wsu:
@nick-deer this includes the PR review and then the summary of fixes.

nick-deer (approved):

PR #1612, opened 2026-05-09
Fix map widget fallback zoom and enable locate button (#229)

### LINKED ISSUE #229: Map picker settings
Default starting location/zoom should EITHER be set by user's location OR by a site admin. Maybe both options?

AM: "Map Picker --  not geo-locate by browser -- can also in the geo location field a center can be set (do we just document how to set the location in Drupal config --- or Mukurtuize it)"

michael-wynne-wsu:
When geolocation is denied, the map now defaults to a world view (zoom 2) instead of the Atlantic Ocean at zoom 12. A locate button is shown in the widget so users can re-trigger geolocation after granting permission. When geolocation succeeds, the map still zooms to street level (zoom 12).

Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/229

Pre-merge checklist:

### DIFF
--- a/modules/mukurtu_collection/config/install/core.entity_form_display.node.collection.default.yml
+++ b/modules/mukurtu_collection/config/install/core.entity_form_display.node.collection.default.yml
@@ -175,7 +175,7 @@ content:
         auto_center: '1'
         scroll_zoom_enabled: '1'
         map_position:
-          zoom: '12'
+          zoom: '2'
           minZoom: '3'
           maxZoom: '18'
           zoomFiner: '0'
@@ -227,7 +227,7 @@ content:
       locate:
         options: '{"position":"topright","setView":"untilPanOrZoom","returnToPrevBounds":true,"keepCurrentZoomLevel":true,"strings":{"title":"Locate my position"}}'
         automatic: '1'
-        control: 0
+        control: 1
       feature_properties:
         values: ''
     third_party_settings: {  }
--- a/modules/mukurtu_collection/config/install/core.entity_form_display.node.collection.default.yml
+++ b/modules/mukurtu_collection/config/install/core.entity_form_display.node.collection.default.yml
@@ -176,7 +176,7 @@ content:
         scroll_zoom_enabled: '1'
         map_position:
           zoom: '2'
-          minZoom: '3'
+          minZoom: '1'
           maxZoom: '18'
           zoomFiner: '0'
           force: 0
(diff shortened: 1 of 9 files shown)


michael-wynne-wsu:
## Accessibility Review

The accessibility check flagged one pre-existing issue that this branch makes more likely to be triggered now that the locate button is enabled by default:

**WCAG 2.1.1 / 2.4.3 - Focus not moved to popup input on open.** When a user clicks a map point, a location description popup appears but focus stays on the map. Keyboard and screen reader users won't know the popup appeared. The fix is a small addition to the existing `popupopen` handler in `mukurtu-leaflet-widget.js`. Everything else passed.

---

## PR Review

**Blast Radius:** LOW - JS change is widget-context only (not browse maps). Config changes affect new installs only; update hook handles existing.
**Security:** Clean - no user input paths, no SQL, no XSS.
**Breaking Changes:** None. All changes are additive or fallback improvements.

---

### MUST FIX

**1. Initial zoom (2) is below minZoom (3) in all configs**

Every form display has `minZoom: '3'` but we're now setting `zoom: '2'`. Leaflet will clamp the initial view to zoom 3 at runtime, making the `'2'` config value misleading. Either lower `minZoom` to `'1'` (allows full world view), or change `zoom` to `'3'` to match.

---

### SHOULD FIX

**2. `keepCurrentZoomLevel: true` in locate button options conflicts with fallback zoom**

The L.Control.Locate button options include `"keepCurrentZoomLevel":true`. Now that the fallback zoom is 2 (world view), a user who denied location, then allows it and clicks the locate button, will be centered on their location but stay at zoom 2 - still a world view, no better than before. Remove `keepCurrentZoomLevel` from the options string in all 6 YAMLs so the button zooms to the accuracy circle.

---

### SUGGESTIONS

**3. `locationfound` listener registered even when content has an existing location**

When editing content that already has a saved location point, `update_leaflet_widget_map()` skips the `map.locate()` call (there's already data to display). Our `map.once('locationfound', ...)` listener still gets registered but will never fire. It's harmless but could mirror the parent's `value.length === 0` guard.

**4. Locate zoom hardcoded to 12**

The `locationfound` handler calls `map.setView(e.latlng, 12)` directly. Leaflet's native accuracy-based zoom computation (capped at maxZoom) is bypassed - so a user with imprecise geolocation (e.g., IP-based, 50km radius) will be zoomed to street level regardless. Consider using `e.accuracy` to compute a reasonable zoom, or accept the current behavior as "good enough for GPS users."

---

The two blocking items are quick - the minZoom/zoom mismatch is a one-line change per YAML, and removing `keepCurrentZoomLevel` is a string edit. Want me to fix both?

michael-wynne-wsu:
All fixes applied and live in DDEV. Here's a summary of what changed:

- **JS**: `locationfound` handler now uses `fitBounds` with the accuracy circle (capped at zoom 16) instead of hardcoded zoom 12, and only registers when there's no existing location data
- **All 6 YAMLs**: `minZoom` lowered from `'3'` to `'1'` (so zoom 2 fallback is actually reachable), and `keepCurrentZoomLevel` removed from locate button options (so clicking the button zooms to the accuracy circle instead of staying at world view)
- **Update hook 40016**: applies the minZoom and locate options changes to existing installations

michael-wynne-wsu:
Pretty isolated changes.

nick-deer (approved):

PR #1613, opened 2026-05-10
Warn when no Categories or Languages exist on content add forms

### LINKED ISSUE #809: Cannot save DH item due to required Categories field
The Category field on DH items is required. We used to provide a default category named Default (ha), but it was removed in https://github.com/MukurtuCMS/Mukurtu-CMS/issues/746. Now, if you make a DH item without first making a category taxonomy term, you will be forced to navigate off the DH creation page, make a term, then come back, and by then all your field values will be lost:

Image

Image

I had a couple ideas on how to address this:

- Make the 'Categories' field optional 
- Inform the user that there are no category taxonomy terms on the site, then prompt them to make one (perhaps via a link)

Open to other ideas of course, @michael-wynne-wsu any thoughts?

michael-wynne-wsu:
- Shows a warning at the top of....
   - `/node/add/digital_heritage` when no Category terms exist
   - `/node/add/dictionary-word` when no Language terms exist. 
   - Each warning links to the relevant vocabulary so editors can add terms before creating content.
   - all `node/add/....` forms if no protocols are available.
- This is probably MOSTLY going to be seen by administrators, maybe mukurtu managers.

Closes #809 

Pre-merge checklist:

### DIFF
--- a/modules/mukurtu_core/src/Hook/FormHooks.php
+++ b/modules/mukurtu_core/src/Hook/FormHooks.php
@@ -485,6 +485,51 @@ class FormHooks {
    * alongside the composer patch and update hook 40004: if the patch fails to
    * apply on a given environment the actions are still hidden from the UI.
    */
+  /**
+   * Implements hook_form_alter().
+   *
+   * Warns editors on content add forms when no Cultural Protocols exist yet.
+   * Protocols are required to control access to content; without at least one,
+   * newly created items cannot be properly shared.
+   */
+  #[Hook('form_alter')]
+  public function formAlterWarnNoProtocols(array &$form, FormStateInterface $form_state, string $form_id): void {
+    $affected_forms = [
+      'node_digital_heritage_form',
+      'node_collection_form',
+      'node_dictionary_word_form',
+      'node_word_list_form',
+      'node_person_form',
+      'node_place_form',
+    ];
+
+    if (!in_array($form_id, $affected_forms, TRUE)) {
+      return;
+    }
+
+    $node = $form_state->getFormObject()->getEntity();
+    if (!$node->isNew()) {
+      return;
+    }
+
+    $protocol_count = \Drupal::entityQuery('protocol')
+      ->accessCheck(FALSE)
+      ->count()
+      ->execute();
+
+    if ($protocol_count > 0) {
+      return;
+    }
+
+    $message = t('There are no Cultural Protocols available. Cultural Protocols are required to control who can access your content. Please create a Cultural Protocol before adding content.');
+
+    $form['no_protocols_warning'] = [
+      '#type' => 'markup',
+      '#markup' => '<div class="messages messages--warning" role="alert">' . $message . '</div>',
+      '#weight' => -100,
+    ];
+  }
+
   #[Hook('form_alter')]
   public function formAlterRemoveNotificationBulkActions(array &$form, FormStateInterface $form_state, string $form_id): void {
     if (!str_starts_with($form_id, 'views_form_user_admin_people_') &&
--- a/modules/mukurtu_core/src/Hook/FormHooks.php
+++ b/modules/mukurtu_core/src/Hook/FormHooks.php
@@ -7,6 +7,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Hook\Attribute\Hook;
 use Drupal\Core\Hook\Order\OrderAfter;
 use Drupal\Core\Render\Element;
+use Drupal\Core\Url;
 use Drupal\og\Og;
 use Drupal\user\Entity\User;
 
@@ -694,8 +695,14 @@ class FormHooks
             return;
         }
 
+        $communities_url = Url::fromUri("internal:/admin/communities-protocols");
+        $link = \Drupal::service("link_generator")->generate(
+            t("communities and cultural protocols"),
+            $communities_url,
+        );
         $message = t(
-            "You do not have permission to create content in any cultural protocols, which is a requirement for all content. If you think you should have access to existing protocol(s), contact your site administrator. If you are the site administrator, ensure that you have created appropriate communities and cultural protocols before creating content.",
+            "You do not have permission to create content in any cultural protocols, which is a requirement for all content. If you think you should have access to existing protocol(s), contact your site administrator. If you are the site administrator, ensure that you have created appropriate @link before creating content.",
+            ["@link" => $link],
         );
 
         $form["no_protocols_warning"] = [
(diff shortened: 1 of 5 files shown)


joelle-birano-wsu (approved):
Love this! Works great!

christa-kavanagh-wsu (commented):
This is a nice feature!

As a Mukurtu Manager:
- There is no option to create content when no communities or protocol are present. 
- There is no option to create a dictionary word when no language is set. Mukurtu managers can add a language; is this something we should duplicate with DH items?

Image

- Mukurtu managers do have the option to create a DH item when no categories have been added, this is inconsistent with what is happening with dictionary words and language.
- The Add categories link works well!

As an admin:
- Admins get a warning when there are no communities and protocols but there is no link. I think a link would be helpful.
- The Add categories and Add languages links work well!

michael-wynne-wsu:
>There is no option to create a dictionary word when no language is set. Mukurtu managers can add a language; is this something we should duplicate with DH items?

I will explore, good call.

michael-wynne-wsu:
# Plan: Block digital heritage creation when category taxonomy is empty

## Context

Mukurtu managers can currently reach `/node/add/digital-heritage` even when no Category taxonomy terms exist. Since `field_category` is required on digital heritage items, letting someone start that form with no categories available is a dead end. There is already a form-level warning (in `FormHooks.php`), but it doesn't block access — the user can still see and partially fill out the form.

The `DictionaryWord` entity solves the identical problem for languages by implementing `BundleSpecificCheckCreateAccessInterface`, which is checked by `MukurtuProtocolNodeAccessControlHandler::checkCreateAccess()` before the form is ever rendered. The fix is to apply the same pattern to `DigitalHeritage`.

## Approach

Implement `BundleSpecificCheckCreateAccessInterface` on the `DigitalHeritage` entity class, returning `AccessResult::forbidden()` when the `category` vocabulary has no terms.

## What was changed

**modules/mukurtu_digital_heritage/src/Entity/DigitalHeritage.php**

- Added `use` imports for `AccessResult`, `AccountInterface`, and `BundleSpecificCheckCreateAccessInterface`
- Added `BundleSpecificCheckCreateAccessInterface` to the `implements` list
- Added `bundleCheckCreateAccess()` static method — queries `category` vocabulary and returns `AccessResult::allowedIf(!empty($categories))`

## Verification

1. `ddev drush cr`
2. Log in as a Mukurtu manager with no Category terms — `/node/add/digital-heritage` should return 403
3. Add a Category term, clear caches — route should be accessible again
4. Confirm editing existing digital heritage nodes is unaffected

michael-wynne-wsu:
## PR Review: feature/809-no-categories-language-warning

**Blast Radius:** LOW — changes are isolated to two module-level entity classes and form hook files. No shared libraries, API contracts, or DB schema touched.
**Security:** No findings.
**Tests:** None added (no existing test suite for these modules visible in the diff).
**Breaking Changes:** None — access tightening only; no removals or API changes.

---

### MUST FIX (Blocking)

**1. Orphaned docblock in `mukurtu_core/src/Hook/FormHooks.php`**

The `formAlterRemoveNotificationBulkActions` docblock ended up immediately above `formAlterWarnNoProtocols` instead of its own method. Two consecutive docblocks appear before `formAlterWarnNoProtocols` — PHP only associates the nearest one, so the `message_digest` docblock is orphaned and `formAlterRemoveNotificationBulkActions` now has no docblock at all. This is a direct result of the reformat reshuffling method order.

**Fix:** Move the `message_digest` docblock to sit directly above `formAlterRemoveNotificationBulkActions`.

---

### SHOULD FIX (Non-blocking)

**2. Massive reformat diff obscures the actual changes**

`mukurtu_core/src/Hook/FormHooks.php` shows 1161 insertions / 462 deletions, but the only functional changes are adding the `Url` import and the link in `formAlterWarnNoProtocols`. The rest is a style reformat (single quotes to double quotes, PSR-12 brace style). This makes the diff extremely hard to review and could mask unintended changes.

Recommend either: (a) land the reformat as a separate commit/PR so this PR's diff is readable, or (b) accept it as-is but call it out in the PR description.

**3. Style inconsistency in `DigitalHeritage.php`**

The new `bundleCheckCreateAccess` method uses 2-space indentation and single quotes, while the `FormHooks.php` files in the same PR use 4-space indentation and double quotes. Minor, but inconsistent within the same PR.

---

### SUGGESTIONS

**4. No tests**

`bundleCheckCreateAccess` is a pure static method with a simple query — it's straightforward to unit test with a mocked entity type manager. The `DictionaryWord` equivalent also has no test, so this is consistent with the codebase pattern, but worth noting.

**5. `bundleCheckCreateAccess` result is not cached**

The method runs an entity query on every access check. If this is called repeatedly (e.g. on a listing page with many nodes), it will fire a query each time. Consider adding `->addCacheTags(['taxonomy_term_list:category'])` to the `AccessResult` so Drupal can cache the result properly:

```php
return AccessResult::allowedIf(!empty($categories))
    ->addCacheTags(['taxonomy_term_list:category']);
```

The `DictionaryWord` equivalent has the same gap.

---

### LOOKS GOOD

- The `BundleSpecificCheckCreateAccessInterface` pattern is the correct approach — mirrors `DictionaryWord` exactly and hooks into the existing `MukurtuProtocolNodeAccessControlHandler`
- `Url::fromUri("internal:/admin/communities-protocols")` is the right call for a path with no registered route name
- New `FormHooks` files for both modules follow the established pattern from `mukurtu_digital_heritage` and `mukurtu_dictionary`
- `accessCheck(TRUE)` in `bundleCheckCreateAccess` is correct (respects access) vs `accessCheck(false)` in the warning hooks (intentional bypass for the count check)

christa-kavanagh-wsu:
> Log in as a Mukurtu manager with no Category terms — /node/add/digital-heritage should return 403

It doesn't return a 403, but it does return the Page not Found message.

christa-kavanagh-wsu (approved):
Looks good!

michael-wynne-wsu:
Three fixes applied:

1. **`FormHooks.php`** — orphaned `message_digest` docblock moved to sit directly above `formAlterRemoveNotificationBulkActions` where it belongs
2. **`DigitalHeritage.php`** — `bundleCheckCreateAccess` now returns with `->addCacheTags(['taxonomy_term_list:category'])`
3. **`DictionaryWord.php`** — same cache tag fix with `taxonomy_term_list:language`

michael-wynne-wsu:

Link:
- https://pr1613-0cm14d40mxctdwu20jejvrbxdvupwigq.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/69ffe3d53d79c52189743c41

nick-deer (approved):

PR #1621, opened 2026-05-12
349 dictionary browse fixes

### LINKED ISSUE #349: Dictionary browse page
At /dictionary.

- [ ] 1) Remove "add dictionary word" and "add word list" buttons.
- [ ] 2) Needs map browse.
   - [ ] I think the map/list toggle would go in the beige search form panel (obv open to recommendations). And the outlined word list space would be swapped to the map (and vice versa) when switching views. 
   - [ ] Should be able to re-use the map browse from /digital-heritage
- [ ] 3) Add a "reset/clear" button (this is needed on all browse pages).
- [ ] 4) Facets need to be changed. Should be (not necessarily in order):
   - [ ] Language (only if there are multiple languages)
   - [ ] Community
   - [ ] Cultural protocol (possibly merge this with community?)
   - [ ] Content type (word, word list - this is PRESENT but is not including word lists)
   - [ ] Word type
   - [ ] Keywords
   - [ ] (Remove Content type)
- [ ] 5) Current "sort by" selections don't fit for a dictionary, and might not even be needed. Default should be alphabetical, with search relevance as the only other option.
- [ ] 6) Add sort by term/translation toggle (maybe - now that there are multiple translations that might be an issue).
- [ ] 7) Media player may need an outline, hard to see against the background.
- [ ] 8) Make translation fields larger text.
- [ ] 9) Spacing, placement, and labelling of the "contributor"/speaker field is a bit off.

michael-wynne-wsu:
Closes #349 

## Summary

- [ ] Added and updated facets (note these are configured a bit different from the other browse pages, they're all single select instead of multi-select checkboxes).
- [ ] Included word lists in the results.
- [ ] Added display options, map view, and sort by controls.

## To do/discuss

- [ ] Currently word lists display after words because they aren't included in the glossary order. That might be fine?
   - [ ] They also don't have a language field so aren't included when language filter is used.
- [ ] Do we still need the "sort by term/translation" toggle?
- [x] Improve the "display options" appearance.
- [x] Add a "reset" button.
- [x] Add glossary filter.

Pre-merge checklist:


michael-wynne-wsu:
Alright, can you test and give feedback on the search/filter tools appearance and functionality? 

I still need to address the previews, but I may do that in a separate PR to keep the scope simpler.

@joelle-birano-wsu @christa-kavanagh-wsu 

Screenshot 2026-05-13 at 10 58 23 AM

christa-kavanagh-wsu:
I like the facets; they seem really helpful.

- The search doesn't return results when you search for part of a word, only if you search for a whole word. For instance it returns two results if I search for "Term" but does not filter out any results if I search for "Þ" as in "Þe". It does return the correct result if I search for the entire word "Þe".
- The word lists facet box is not appearing for me; is there a reason for that?

michael-wynne-wsu:
@christa-kavanagh-wsu good catches.

1. Partial term searches are enabled. Not sure if you'll get results off a single letter, but see how it works.
2. Word list filter should be working, had to add a thing to re-index words after they were added to a word list.

michael-wynne-wsu:
FYI I am going to move the teasers to a separate PR and issue, so this IS just about the actual search and facet behaviour.

joelle-birano-wsu:
Looking pretty good so far! A couple of things I spotted:

- Something funky is going on with the Word Lists facet...I'm having a hard time getting all my word lists to show up in the facet, but one thing that seems to work for me is when I add map points to a word in the word list which doesn't make much sense but I've recreated it a few times now.

- I'm also just having trouble with word lists in general - If I try to add a word to a word list and save it, it breaks the site but the word is added. So not sure if that's related.

Dictionary-Mukurtu-CMS-05-13-2026_12_14_PM

I like how facets update dynamically based on other selections, but sometimes they'll disappear and rearrange, which is a little disorienting.

For example, here's before filtering by word list:
Screenshot 2026-05-13 121653

and after:
Screenshot 2026-05-13 121849

michael-wynne-wsu:
@joelle-birano-wsu Okay I'll dig into the word list errors and stuff.

As for the facets, it doesn't look like they're rearranging, but if a facet isn't applicable to the current results it's not appearing/being removed. We actually do the same on the other browse pages, but it's more noticeable here. I will see if I can think up a good way around that.

christa-kavanagh-wsu:
Ok, as a second pass I am getting some of the same weirdness with word lists. I didn't get that weirdness when I made one on this PR initially; not sure what happened there. 
The reappearing and disappearing facets are a little disconcerting at first, but I don't think they're ultimately going to be a huge deal because it seems unlikely that folks will be adding and removing metadata fields that often.
You can only view so many word lists; you can't scroll down in the facets menu? I think it's the same funkiness Joelle was having.

michael-wynne-wsu:
Latest updates (which I have note tested on Tugboat yet)

- Fixed breaking change when adding a word to a word list.
- Set word list facet to select single, not multiple, which may resolve some of the noted issues.
- Fixed a glossary order bug.
- Fixed display of pagination at end of results.

Facets showing or not has nothing to do with changing metadata fields, the displayed facets correspond to the active search results. So if you select content types>word list, the language facet goes away because word lists don't have that field. I could probably make that stuff static, but then you get choices that won't actually work/affect your results.

joelle-birano-wsu:
Word lists are working and showing up in facets! I didn't catch what was wrong with glossary order or pagination but those appear to be fine.

michael-wynne-wsu:
Sometimes the glossary order wasn't working properly. And the pagination just wasn't styled yet. Neither major issues!

michael-wynne-wsu:
## WCAG 2.2 AA Accessibility Audit

**Files audited:** `mukurtu_browse.module` (browse-links injection), `_dictionary.scss`, `mukurtu-dictionary-view-switch.js`, `node--word-list--teaser.html.twig`

### Issues Found

**Issue 1 — FAIL: Active pager link uses low-contrast border (1.4.11 Non-text Contrast)**
`.pager .is-active a` removes underline and relies solely on a `border-block-end` in `--brand-secondary` (#e6ab49) on white — contrast ~2.1:1, below the 3:1 minimum for UI component boundaries. Fix: use `--brand-secondary-dark` (#9d6915, ~3.6:1) or add a non-color indicator (font-weight, background).

**Issue 2 — FAIL: No explicit focus ring on `.browse-links` buttons (2.4.7 Focus Visible)**
The `block-link` mixin sets `outline-offset: 1px` but no `outline` style, leaving browser-default focus rings. The pattern is already in the codebase (`.search-collapse-toggle__input` uses `outline: 3px solid var(--brand-primary-dark); outline-offset: 2px`). Add the same to `[data-browse-mode]` and `.filter-reset` in `_dictionary.scss`.

**Issue 3 — FAIL: Glossary link contrast on `--brand-secondary-accent` background (1.4.3 Contrast)**
`--brand-secondary-accent` (#f1b85a) with `--brand-primary-dark` (#107996) link color gives ~2.1:1. Fix: set `color: var(--brand-text-color)` (#464646) on `.glossary .facets-widget-links a` — that gives ~4.8:1.

**Issue 4 — WARNING: Filter label contrast (1.4.3)**
White on `--brand-primary-dark` (#107996) is ~3.1:1. Passes large text (≥19px or ≥14px bold) but not normal text. Verify `--font-size-m` value.

**Passed:** Semantic HTML in teasers; `aria-hidden`/`tabindex="-1"` on decorative image links; `aria-label` on all toggle buttons; `hidden` attribute for show/hide; visually-hidden pattern on the search-collapse checkbox; "Reset" link has a real `href`.

---

## PR Review: Dictionary Browse Fixes (#349)

**Blast radius:** MEDIUM-HIGH — `mukurtu_browse.module` is shared by `/browse` and `/digital-heritage`; `search_api.server` config is site-wide.

---

### MUST FIX (Blocking)

**1. Hardcoded `/dictionary` URL in reset link**
`mukurtu_browse.module:40` — `href="/dictionary"` will break on any Drupal install not at the server root (subdirectory installs) and doesn't respect the language prefix.

```php
// Fix: compute the URL in PHP and pass it as context
'reset_url' => \Drupal::url('mukurtu_dictionary.page'),
// Template:
'<a href="{{ reset_url }}" class="filter-reset">{{ reset_text }}</a>'
```

**2. `search_api.server` matching change is site-wide**
`search_api.server.mukurtu_database_server.yml` changes `matching: words` → `matching: partial`. This server is shared by all indexes (browse, digital heritage, collections, etc.), not just the dictionary. Partial matching is significantly slower on large datasets and changes search behavior everywhere. This needs an intentional decision and a note in the PR — or scope it to only affect the dictionary if that's the intent.

---

### SHOULD FIX (Non-blocking)

**3. `search_api.index` word_type property path fix needs an update hook**
`search_api.index.mukurtu_dictionary_index.yml` changes `property_path` from `field_additional_word_entries:entity:field_word_type:entity:name` → `field_word_type:entity:name`. This is a config change that won't auto-apply on existing sites without re-importing config — the existing `mukurtu_dictionary.install` update hooks should include a step to re-save or rebuild the search index after this change. Verify this is covered.

**4. `display options toggle` weight change affects all browse pages**
`mukurtu_browse.module:28-29` — `$toggle['#weight']` changed from `-5` to `100` (globally, not conditionally). This reorders the display options toggle on `/browse` and `/digital-heritage` too, not just `/dictionary`. Verify this is intentional.

**5. `word_list` added to dictionary content type filter**
`views.view.mukurtu_dictionary.yml:281` adds `word_list: word_list` to the content type filter. The `MukurtuDictionaryController::access()` already checks for word_list create access, so the intent is clear — but the `node--word-list--teaser.html.twig` template should be confirmed to handle display in the dictionary context gracefully (e.g., no broken field references).

---

### SUGGESTIONS

**6. `NodeHooks.php` — verify hook registration**
The new `src/Hook/NodeHooks.php` file uses Drupal 11-style `#[Hook]` attributes. Confirm the hook name and that it's registered correctly (the `.module` file or `services.yml` shouldn't duplicate it procedurally).

**7. Pager `quantity: 9` is an unusual value**
Full pager with 9 page links (`views.view.mukurtu_dictionary.yml:142`) — most Drupal defaults use 5 or 7. This is fine but worth a comment if intentional for the glossary layout.

---

### Looks Good

- Update hook 40017 is defensive (guards against double-application with `str_contains`)
- `aria-label` attributes on all toggle buttons
- `hidden` attribute pattern for list/map JS toggle is correct
- Reset link consolidated into `.browse-links` with `flex: 1` — clean implementation
- Glossary facet `css_class: list` is appropriate for the facet widget output
- Facet reordering (language → community → type → word_type → glossary_entry) makes semantic sense

michael-wynne-wsu:
All done. Here's a summary of every change made:

**PHP — mukurtu_browse.module**
- Consolidated three identical `plugin_id` string comparisons into one `$is_dictionary` variable
- Reset link URL changed from hardcoded `"/dictionary"` to `\Drupal::url('mukurtu_dictionary.dictionary_page')` — safe for subdirectory installs and language prefixes
- Display options toggle weight scoped: weight `100` only on dictionary, `-5` on all other browse pages (restoring previous behaviour on `/browse` and `/digital-heritage`)

**SCSS — _dictionary.scss**
- Added `:focus-visible { outline: 3px solid var(--brand-primary-dark); outline-offset: 2px }` to `.browse-links a, button` (WCAG 2.4.7)
- Active pager border changed from `--brand-secondary` (#e6ab49, ~2.1:1) to `--brand-secondary-dark` (#9d6915, ~3.6:1) (WCAG 1.4.11)
- Glossary links: added `color: var(--brand-text-color)` (#464646, ~4.8:1 on the accent background) (WCAG 1.4.3)

**Note on filter label contrast (Issue 4):** `--font-size-m` is 18px bold — technically just under the large-text threshold (18.67px), so white on `--brand-primary-dark` (~3.1:1) is a pre-existing failure that predates this branch and would need a separate fix.

michael-wynne-wsu:

Link:
- https://pr1621-g8mhguckwemi5es04eu8043yphk97zmi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a0391933d79c52189d8a4fe

modules/mukurtu_core/config/install/search_api.server.mukurtu_database_server.yml
nick-deer, line 1: To confirm, we _do_ want partial matching on all the browse pages?

michael-wynne-wsu: I think it's a good change, yes. unless I'm misunderstanding how partial matching works.

christa-kavanagh-wsu (commented):
I am still getting some weirdness with the word lists facet: I created 5 word lists, but it's only showing one. 
image

michael-wynne-wsu:
@christa-kavanagh-wsu Is that the only word list that has words in it?

IF so, that's expected, it's a "filter words by their word lists" filter, not a "list all word lists" (that's the content type filter basically)

christa-kavanagh-wsu:
@michael-wynne-wsu Yes, it was. I went ahead and added words to some of the others and they popped up as expected, so nevermind!

nick-deer (approved):

PR #1633, opened 2026-05-15
Person creation on the fly 225

### LINKED ISSUE #225: Add person record references on the fly
Any way to add "related people" on the fly? Creating many in advance to relate later is tedious.

alexmerrill:
Pre-merge checklist:

## Summary

Closes #225. Editors who can't find a person in the related-person entity
browser can now create one without leaving their current form.

- Clicking "Create a new person record" opens the person form in a new tab
- After saving, the new tab automatically closes and the editor is returned
  to the original form, where the entity browser has already opened
  pre-filtered to show only the person they just created
- One click selects the new person — no searching required
- Includes `aria-live` announcement for screen readers and visually-hidden
  "(opens in new tab)" text on the create link (WCAG 2.1 SC 3.2.2)

## Why BroadcastChannel instead of Inline Entity Form

The natural Drupal approach for on-the-fly record creation is
`inline_entity_form_complex`, which embeds the creation form inside the
parent form. For person records this hits a hard structural wall: the
person form contains a `media_library_widget` and an entity browser, both
of which open their own modals and make their own AJAX calls. When IEF
nests the full person form inside the parent person form, those inner AJAX
callbacks lose track of which form instance they belong to — the same class
of DOM conflict that caused the original modal approach on this issue to be
abandoned. IEF would be a deeper version of the same problem, not a
solution.

There are also practical costs: `inline_entity_form` is not currently a
Mukurtu dependency (new Composer requirement); a stripped-down
`related_person_inline` form display would need to be maintained alongside
the default; and because the inline form would have to omit media and entity
browser fields to avoid the conflicts, the created person would be
incomplete and need a follow-up edit anyway.

The BroadcastChannel approach avoids all of this by keeping the two forms
completely separate — no nesting, no shared DOM, no competing AJAX handlers.
The person is created on its own full-featured form. After save, the
browser's native BroadcastChannel API sends a small message (nid + title)
to any same-origin tab that's listening. The original tab receives it,
locates the correct entity browser using a UUID stored in sessionStorage
at click time, prepends `?title=<name>` to the entity browser's iframe
path input, and triggers the open — all before the new tab closes.

New surface area: two JS files, two PHP hooks, one libraries.yml. No new
dependencies, no config changes, no update hook required.

## Test plan

- [ ] Open a person record in edit mode; add a Related People paragraph
- [ ] Click "Create a new person record (opens in new tab)"
- [ ] Fill in a name and save the new person
- [ ] Confirm the new tab closes and focus returns to the original form
- [ ] Confirm the entity browser opens automatically, filtered to the new person
- [ ] Click the person to select; confirm the paragraph saves correctly
- [ ] Confirm the self-exclusion still works (current person does not appear
      in their own entity browser)
- [ ] Test with BroadcastChannel unavailable (old Safari): confirm the link
      still opens the new tab and the workflow degrades gracefully to manual
      selection


christa-kavanagh-wsu:
One thing I noticed going through this the other day:
At the moment you can add the on-the-fly created related person record to the original record, but you can't add the original record as a related person to the on-the-fly record because the original record hasn't been saved. When creating these related person records on the fly, is there a way to streamline this workflow by either:

a) Making the relationship with the original record reciprocal, or if that is not possible
b) Flagging the new related person's person record so that users don't forget to add the reciprocal relationship?

alexmerrill:
### Reciprocal relationship reminder

After Person A is saved with Person B in the Related People field, the
system checks whether Person B also lists Person A as a related person.
If not, a warning message appears on Person A's page:

> Need to add a reciprocal relationship to: [Person B] ← linked to edit form

This fires on both insert and update, covers multiple related people in
one save, and stops appearing once the reciprocal is in place. It
naturally handles the on-the-fly creation case — a newly created person
can never have the reciprocal set yet.

**Implementation:** `hook_node_insert` / `hook_node_update` →
`_mukurtu_person_warn_missing_reciprocals()` in `mukurtu_person.module`.
Walks `field_related_people` → `field_related_person` for each paragraph,
loads the referenced person, checks their `field_related_people` in the
opposite direction. No new JS, no new tempstore keys, no new dependencies.

alexmerrill:
# PR Review: Person creation on the fly (#1633)

**Blast Radius:** MEDIUM — new hooks in `mukurtu_person.module` fire on every person node save; two new global JS behaviors; tempstore writes on every person form render  
**Security:** No findings  
**Test Coverage:** No automated tests; comprehensive manual test plan in PR body  
**Breaking Changes:** None — additive only

---

## Must Fix (Blocking)

None.

---

## Should Fix (Non-blocking)

### 1. `_mukurtu_person_warn_missing_reciprocals()` has an N+N² entity load pattern

For a person with M related people, each of whom has their own related people, this function performs up to M entity loads for the outer loop plus M×avg(N) entity loads for the inner "does Person B link back?" check. Drupal's entity storage caches within a request, so repeated loads of the same entity are free — but a person record with many relationships pointing to other well-connected people could still hit several dozen loads per save.

This is acceptable today. Worth adding a comment acknowledging it:

```php
// Note: this loads paragraphs and nodes for each related person. Entity
// storage caches within the request, but consider a direct entity query
// approach if person records with large relationship graphs cause slow saves.
```

### 2. `hook_node_update()` fires on every person save, not just when related people change

If a person has no `field_related_people` the loop exits immediately with negligible cost. But if they do have related people, the warning fires on every single save — even saves that only changed the title or body. This is by design (persistent reminder until the reciprocal is set), but an editor fixing an unrelated typo will see the warning again.

Confirm whether that's the intended UX, or whether comparing against `$node->original->get('field_related_people')` to only warn on newly added relationships is worth the extra complexity.

---

## Suggestions

### 3. URL title-param stripping has a double-`&` edge case

```js
src = src.replace(/([?&])title=[^&]*/g, '$1').replace(/[?&]$/, '');
```

If the URL is `?other=val&title=foo&after=bar`, step 1 produces `?other=val&&after=bar` — the double `&` is preserved. Browsers handle this fine, but it's slightly malformed. A more robust alternative:

```js
src = src.replace(/[?&]title=[^&]*/g, '').replace(/^&/, '?');
```

Non-blocking — current behavior works correctly for all real entity browser URL patterns.

### 4. 500ms `window.close()` delay is a magic number with no comment

`setTimeout(() => window.close(), 500)` should have a brief comment explaining why the delay exists — it lets the BroadcastChannel message post and the browser paint the save confirmation before the tab is torn down. Future maintainers will otherwise wonder whether it can be safely shortened.

---

## Looks Good

**BroadcastChannel design is the right call.** The PR description's explanation of why `inline_entity_form` can't work here — modal-in-modal AJAX conflicts, a new Composer dependency, and a stripped-down creation form that would require a follow-up edit anyway — is exactly right. BroadcastChannel keeps both forms fully independent with no shared DOM or competing AJAX handlers.

**AJAX re-attach safety in `mukurtu-person-broadcast.js`.** `delete drupalSettings.mukurtuPerson.newPerson` immediately after reading prevents the behavior from re-firing if Drupal's behavior system re-attaches (e.g., after a subsequent AJAX response on the save page).

**`once()` is used correctly in `mukurtu-person-listener.js`.** The click handler uses `once('mukurtu-person-create-link', '.mukurtu-person-create-link', context)` so it rebinds correctly when new Related People paragraphs are added via AJAX. The channel setup uses `once('mukurtu-person-channel', document.body)` so the channel is only opened once per page regardless of how many paragraph adds fire.

**sessionStorage lifecycle is clean.** The UUID is written on click, read and immediately deleted on message receipt — no stale state can persist across unrelated entity browser opens.

**Graceful degradation on old Safari.** Both JS files check `'BroadcastChannel' in window` and bail safely. The link still opens in a new tab; the editor just has to find the person manually. No broken UI.

**Accessibility is thorough.** Visually-hidden `(opens in new tab)` text satisfies WCAG 2.1 SC 3.2.2. The `aria-live="polite"` region announces when the entity browser reopens and is cleaned up after 3 seconds to avoid polluting the accessibility tree.

**Tempstore is correctly scoped to private (per-user).** Both `new_person_broadcast` and `related_person_exclude` use `tempstore.private`, so no cross-user data leakage is possible.

**The self-exclusion TempStore limitation is honestly documented.** The comment in `hook_field_widget_single_element_form_alter` explicitly describes the known blast radius (the stored exclusion applies to all `mukurtu_content_browser` opens within the TTL window, not just the current form) and flags a per-form-build-ID key as the correct long-term fix. That's the right way to ship a known limitation.

**Reciprocal relationship warning is well-targeted.** `_mukurtu_person_warn_missing_reciprocals()` correctly walks `field_related_people` → `RelatedPerson` paragraph → `field_related_person` → the linked person node, then checks the reverse direction. The `$related->bundle() !== 'person'` guard prevents false positives if the reference ever resolves to non-person content.

---

## Checklist

| Item | Status |
|---|---|
| PR title and description explain the change | ✅ |
| Design rationale (vs inline_entity_form) documented | ✅ |
| BroadcastChannel graceful degradation on unsupported browsers | ✅ |
| No hardcoded secrets / SQL injection / XSS vectors | ✅ |
| User-controlled `title` safely handled in JS (`@title` escaped; URL-encoded in query string) | ✅ |
| sessionStorage cleaned up on use | ✅ |
| `once()` guards prevent double-binding on AJAX re-attach | ✅ |
| Tempstore scoped to private (per-user) | ✅ |
| Accessibility: visually-hidden new-tab indicator | ✅ |
| Accessibility: `aria-live` announcement on entity browser open | ✅ |
| No new Composer dependencies | ✅ |
| No DB schema changes or update hooks needed | ✅ |
| Updated from `main` (merge completed) | ✅ |
| N+N² entity load pattern — add explanatory comment | ⚠️ |
| `hook_node_update` fires on all saves — confirm UX intent | ⚠️ |
| CI checks passing | ⏳ Still running |

---

**Verdict: Ready to merge pending green CI.** The design is sound, the implementation is careful, and edge cases are handled or honestly documented. Resolve the CI status and confirm the "warns on every save" UX is intentional — both are minor. The core feature and the reciprocal relationship warning are solid.

alexmerrill:
## Review items addressed

| # | Severity | Item | Resolution |
|---|---|---|---|
| 1 | Should Fix | `_mukurtu_person_warn_missing_reciprocals()` N+N² entity load pattern | Added docblock explaining the load pattern and its request-level cache mitigation |
| 2 | Should Fix | `hook_node_update()` warned on every save, not just when relationships changed | Refactored to build a `$pre_existing_nids` map from `$node->original` before the loop — warnings now only fire for newly added related people; pre-existing relationships are skipped |
| 3 | Suggestion | URL title-param stripping regex left orphan `&&` when param was mid-string | Fixed: `replace(/[?&]title=[^&]*/g, '').replace(/^&/, '?')` — removes the delimiter together with the param and promotes a leading `&` back to `?` if needed |
| 4 | Suggestion | `setTimeout(() => window.close(), 500)` magic number | Already had an explanatory comment — no change needed |

christa-kavanagh-wsu (approved):
This looks good! I think the reminder works really well, even when adding multiple related persons.

joelle-birano-wsu:
This is really cool!

> This fires on both insert and update, covers multiple related people in one save, and stops appearing once the reciprocal is in place. It naturally handles the on-the-fly creation case — a newly created person can never have the reciprocal set yet.

Does this mean that the message to add a reciprocal relationship is supposed to be persistent even after leaving the page and coming back to it, or making edits and saving? If that's the case, I'm not seeing that, but if that isn't the case, then everything works great!

michael-wynne-wsu:
Functionally I think it's great and good for v4 launch (as discussed, a more complete system can be explored later).

The only changes I'd like are updating the message text a bit and adding some more helper text. Some suggestions, but please edit/suggest changes as needed.

1. "Do you need to add a reciprocal relationship to: [person record]?" > "Relationships are created on both person records separately. Add the relationship back to this person record here: [person record]"
2. Add description/helper text for the "Can't find the person? Create a new person record." button: "A new person record form will open in a new tab. When you save that person record the tab will close and you will be redirected back here so that you can reference that new person record. Note that you will still need to add the relationship to THIS person record after saving here."

joelle-birano-wsu:
One more thing...once Person A is saved with Person B created and added as a related person, is it possible to automatically add Person A to Person B as a related person so that all you have to do is define the relationship?

christa-kavanagh-wsu:
> Functionally I think it's great and good for v4 launch (as discussed, a more complete system can be explored later).
> 
> The only changes I'd like are updating the message text a bit and adding some more helper text. Some suggestions, but please edit/suggest changes as needed.
> 
>     1. "Do you need to add a reciprocal relationship to: [person record]?" > "Relationships are created on both person records separately. Add the relationship back to this person record here: [person record]"

Maybe "Relationships are created on both person records separately. Add the reciprocal relationship here: [Person record]"

I just don't know about "relationship back". 

> 
>     2. Add description/helper text for the "Can't find the person? Create a new person record." button: "A new person record form will open in a new tab. When you save that person record the tab will close and you will be redirected back here so that you can reference that new person record. Note that you will still need to add the relationship to THIS person record after saving here."

I think this works.

michael-wynne-wsu:
> One more thing...once Person A is saved with Person B created and added as a related person, is it possible to automatically add Person A to Person B as a related person so that all you have to do is define the relationship?

That's what we were talking about working on in the future so that we don't sink too much more time in it now. There are two challenges with that:

1. Our relationship types are just one way. We haven't built the system to know that if Person A is Person B's father.... what is Person B to Person A? As it stands we can't really logic that unless we overhaul that function (I do have ideas but it would be a big change).
2. We have limitations on one entity changing another - what if person records are in different protocols, can we be allowed to edit another protocol's content? And in this case it could make sense, but it would probably require carving out some safety exemptions for person records, which isn't trivial.

michael-wynne-wsu:
@christa-kavanagh-wsu I just don't know about "reciprocal relationship". It comes across as jargony to me and I don't know if we've used the term elsewhere?

christa-kavanagh-wsu:
> @christa-kavanagh-wsu I just don't know about "reciprocal relationship". It comes across as jargony to me and I don't know if we've used the term elsewhere?

Bi-directional instead? We use that in the helper text.

michael-wynne-wsu:
Paired? Complementary?

joelle-birano-wsu:
> 1. Our relationship types are just one way. We haven't built the system to know that if Person A is Person B's father.... what is Person B to Person A? As it stands we can't really logic that unless we overhaul that function (I do have ideas but it would be a big change).

Ahh I see. Yes that definitely is a future development kind of thing. I kind of don't mind having to add the relationship every time, unless there's some kind of ability for users to pre-map relationships or something, but THAT'S a whole can of worms and family systems are...complex. 

> 2. We have limitations on one entity changing another - what if person records are in different protocols, can we be allowed to edit another protocol's content? And in this case it could make sense, but it would probably require carving out some safety exemptions for person records, which isn't trivial.

Right. I see that. It's definitely larger in scope than I first appreciated. 

I think it's great as is!

christa-kavanagh-wsu:
> Paired? Complementary?

I like paired.

michael-wynne-wsu:

Link:
- https://pr1633-igkhprktxgotuqvkqmyi3eqrfi5c5b8o.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a0743ce6effb34728e446ff

alexmerrill:
> This is really cool!
> 
> > This fires on both insert and update, covers multiple related people in one save, and stops appearing once the reciprocal is in place. It naturally handles the on-the-fly creation case — a newly created person can never have the reciprocal set yet.
> 
> Does this mean that the message to add a reciprocal relationship is supposed to be persistent even after leaving the page and coming back to it, or making edits and saving? If that's the case, I'm not seeing that, but if that isn't the case, then everything works great!

The notification appears on the initial save and any update/save thereafter (until a relationship is added), not meant to be persistent on view just directly after those save/update actions happen.

joelle-birano-wsu:
> The notification appears on the initial save and any update/save thereafter (until a relationship is added), not meant to be persistent on view just directly after those save/update actions happen.

In that case, I'm only seeing it on the initial save, not updates.
Screenshot 2026-05-27 at 9 23 41 AM

alexmerrill:
> > The notification appears on the initial save and any update/save thereafter (until a relationship is added), not meant to be persistent on view just directly after those save/update actions happen.
> 
> In that case, I'm only seeing it on the initial save, not updates. Screenshot 2026-05-27 at 9 23 41 AM

Interesting,  I will look to see where that was lost in the process

alexmerrill:
Ok, everything is kinda working as expected again... kinda.  On my local everything works fine on Drupal 11.2.xxx but on tugboat using 11.3.1.xxx I am seeing a weird tagify / paragraphs subform issue described here: https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1680     it is also appearing on at least one paragraphs subform word type.....  @michael-wynne  there might be other items to peruse too...

michael-wynne-wsu (approved):
Knowing that the tagify focus issues are being resolved separately, I think this is good to go.

nick-deer (approved):
This is really cool! Clever fix!

PR #1634, opened 2026-05-15
1305 dictionary word display updates

### LINKED ISSUE #1305: Dictionary word display updates
- [ ] When a word includes word entries, especially if there are several and/or any of them are quite long, the page gets very long.
- [ ] One idea is to apply a tabbed view between words entries, similar to community records, but we are open to other ideas.
- [ ] Identifying the fields can be a bit hard. Maybe adjust spacing or field labels? Could do two-column content? 
- [ ] Could consider moving additional information fields to sidebar?

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1305

Pre-merge checklist:


joelle-birano-wsu:
I think itʻs looking really good. I like the layout and the word entry tabs. 

- I think the word entry term could be a bit more bold/bigger

- Not sure if itʻs my browser, but when I click on the base word, the border is broken. It doesn't happen when I click on the word entry.

Screenshot 2026-05-15 at 12 37 30 PM
Screenshot 2026-05-15 at 12 46 32 PM

- Looking at it now Iʻm wondering if definition should be higher up the page...maybe under translation?

christa-kavanagh-wsu:
This looks good. Here are some things I noticed:
- I am having the same problem with the border breaking that Joelle is.
- I think pronunciation and definition should both be higher; definitely above the sample sentences. Maybe: 
|Term|
|Translation|Pronunciation|
|Definition|
|Recording|
|Alternate spelling|Word origin|
|Sample sentences|
|Contributor|Source|
(I think I got all of them)?

michael-wynne-wsu:
> I think the word entry term could be a bit more bold/bigger

Do you mean in the tab switcher?

> Not sure if itʻs my browser, but when I click on the base word, the border is broken. It doesn't happen when I click on the word entry.

That's a browser selection indicator, not an intentionally styled border. It goes away when you interact with the page, I don't think it's worth worrying about.

Easy enough to make the field layout changes, if you both have similar preferences

joelle-birano-wsu:
> > I think the word entry term could be a bit more bold/bigger
> 
> Do you mean in the tab switcher?

Yes the text in the tab switcher
> 

> That's a browser selection indicator, not an intentionally styled border. It goes away when you interact with the page, I don't think it's worth worrying about.
> 
Fair enough!

> Easy enough to make the field layout changes, if you both have similar preferences

I like Christa's suggestions for field layout

michael-wynne-wsu:
Okay I got the focus border sorted and made the tab text more prominent.

I also implemented the field re-ordering, making some changes per my own discretion. Let me know!

joelle-birano-wsu (approved):
I think it looks great!

michael-wynne-wsu:
A decent handful in this one, which doesn't surprise me...


## PR Review: 1305 Dictionary Word Display Updates (#1634)

**Blast Radius:** LOW - Changes are scoped to dictionary word display (Twig templates, SCSS, JS, entity view display config). No shared libraries or API contracts affected.
**Security:** No findings
**Tests:** No new tests (consistent with codebase pattern for UI/config changes)
**Breaking Changes:** Update hook replaces entire entity view display config - see below

---

### MUST FIX (Blocking)

**1. Pre-merge checklist item unchecked: "Update from main"**
The PR checklist has "I have updated from `main` and resolved any merge conflicts." left unchecked. Per the workflow, this must be completed before merge.

**2. CI failing - `AccessByProtocolTest` (6 failures, 7 errors)**
The `Mukurtu v4 - PHP 8.3` check is failing on `AccessByProtocolTest::testTwoProtocolsTwoOpenAll` and related tests. These failures appear to be in an unrelated module (`mukurtu_protocol`), suggesting they are pre-existing and not introduced by this PR - but they need to be confirmed as pre-existing (check main branch CI) and documented, or fixed, before merge.

**3. Update hook wipes admin-customized display configurations**
In `mukurtu_dictionary.install`, `mukurtu_dictionary_update_40018` calls `$config->setData($data)` which wholesale replaces the entity view display config. The comment says "Preserve existing third_party_settings" but `setData()` does the opposite - it overwrites everything. On any site where an admin has customized field ordering or field_group layouts via the UI, those customizations will be silently destroyed. The safer approach is to only set the specific weight and order keys that changed:

```php
// Instead of $config->setData($data), update only what changed:
$config->set('content.field_definition.weight', $data['content']['field_definition']['weight']);
// etc. for each changed weight
$config->set('third_party_settings.field_group...', $data['third_party_settings']['field_group']...);
$config->save();
```

---

### SHOULD FIX (Accessibility - from WCAG 2.2 AA audit)

**4. Tab widget missing `aria-controls` / `aria-labelledby` association (WCAG 4.1.2)**
Each `<a role="tab">` needs an `id` and `aria-controls` pointing to its panel. Each `<div role="tabpanel">` needs `aria-labelledby` pointing back to the tab. Without this, screen readers cannot announce the tab/panel relationship.

In `node--dictionary-word--full.html.twig`:
```twig
{# Base tab #}
<a href="#tab-base-entry" role="tab" id="tab-btn-base-entry" aria-selected="true" aria-controls="tab-base-entry">

{# Base panel #}
<div id="tab-base-entry" class="horizontal-tabs-pane" role="tabpanel" aria-labelledby="tab-btn-base-entry">

{# Loop tabs #}
<a href="#tab-entry-{{ loop.index }}" role="tab" id="tab-btn-entry-{{ loop.index }}" aria-selected="false" aria-controls="tab-entry-{{ loop.index }}">

{# Loop panels #}
<div id="tab-entry-{{ loop.index }}" class="horizontal-tabs-pane" role="tabpanel" hidden aria-labelledby="tab-btn-entry-{{ loop.index }}">
```

**5. No keyboard navigation in tab widget (WCAG 2.1.1)**
In `dictionary-word-tabs.js`, only `click` events are handled. The ARIA APG tab pattern requires:
- Arrow keys (Left/Right) to move between tabs
- Home/End to jump to first/last tab
- Roving `tabindex` (active tab = `0`, others = `-1`)
- Enter/Space `keydown` with `preventDefault()`

Without this, keyboard-only users cannot navigate between tabs.

**6. Empty tab label fallback missing (WCAG 1.3.1)**
```twig
<span class="div-title">{{ entry.entity.field_word_entry_term.value }}</span>
```
If `field_word_entry_term` is empty, the tab has no accessible name. Add a fallback: `{{ entry.entity.field_word_entry_term.value ?: 'Entry ' ~ loop.index }}`.

---

### SUGGESTIONS

**7. Update hook comment is misleading**
The comment `// Preserve existing third_party_settings (field_group layout) and content weights.` implies preservation, but `setData()` is a replacement. Either update the comment to accurately describe the behavior ("Replace config with install defaults to apply new field ordering"), or change the approach (see item 3 above).

**8. `panes[index]` coupling in JS**
The JavaScript in `dictionary-word-tabs.js` maps tab index to pane index positionally (`panes[index].hidden = false`). This works correctly as written since tabs and panes are rendered in matching order, but would silently break if the template ever renders them out of sync. A more robust approach would use `href`-to-`id` matching (`document.querySelector(link.getAttribute('href'))`).

---

### LOOKS GOOD

- Contrast fix: white-on-yellow (~1.72:1) corrected to `var(--brand-text-color)` on yellow (~4.78:1) - passes WCAG 1.4.3 AA
- `hidden` attribute used correctly on inactive panels (hides from both visual display and assistive tech)
- `aria-selected` toggled correctly in JS on tab activation
- `:focus:not(:focus-visible)` pattern correctly suppresses mouse focus ring while preserving keyboard focus ring
- SCSS scoped to `.dictionary-word-tabs` / `:is(#extra-specificity-hack, .dictionary-word-tabs)` - no global bleed
- Drupal `once()` used correctly to prevent double-binding
- Library registration in `mukurtu_v4.libraries.yml` is clean; library is attached via `attach_library()` only when the tabbed UI is actually rendered
- No-tab fallback path (`{% else %}`) correctly renders the single-entry layout without the tab widget
- Sidebar empty-section hiding (`:not(:has(.field__item))`) is a good UX improvement

---

**Summary:** Three blockers: the unchecked "update from main" checklist item, the unconfirmed CI failure (likely pre-existing but needs verification), and the update hook's destructive `setData()` behavior. The accessibility issues from items 4-6 should also be resolved before merge - the tab widget is not keyboard-navigable and lacks required ARIA associations for screen readers.

michael-wynne-wsu:
All fixes committed. Here's a summary of what was done:

**Fixed (6 of 8 review items):**

1. **`aria-controls` + `aria-labelledby`** - Each `<a role="tab">` now has an `id` and `aria-controls` pointing to its panel; each `role="tabpanel"` has `aria-labelledby` pointing back to its tab.
2. **Keyboard navigation** - `dictionary-word-tabs.js` now handles ArrowLeft/Right, Home/End, Enter/Space, with roving `tabindex` (active tab = `0`, others = `-1`).
3. **href-to-id matching** - Pane activation uses `tabs.querySelector(link.getAttribute('href'))` instead of positional index.
4. **Empty tab label fallback** - `{{ entry.entity.field_word_entry_term.value ?: 'Entry ' ~ loop.index }}`.
5. **Misleading comment** - Updated to accurately describe what `setData()` does.
6. **Merged from main** - Already up to date.

**Not needing action:**
- **CI failure** - Confirmed pre-existing on main (unrelated `AccessByProtocolTest`); not caused by this PR.

christa-kavanagh-wsu (approved):
Looks great!

nick-deer (approved):

joelle-birano-wsu:
Updates [Creating dictionary words (and word entries)](https://github.com/MukurtuCMS/mukurtu-user-manual/issues/79)

PR #1641, opened 2026-05-18
776 exclude current entity from content browser

### LINKED ISSUE #776: Content browser should not display or at least not allow self-selection of the content being edited.
Eg: a collection can't be it's own sub-collection, or within it's own collection tree. A DH item can't have itself as a related item.

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/776

Prevent self-selection by filtering the currently-edited node out of the mukurtu_content_browser view results. The entity_browser Modal display already passes original_path (e.g. /node/42/edit) as a query parameter when opening the browser, so we parse the NID from it and add a nid != NID condition via hook_views_query_alter. Covers all fields that use the browser: child collections, items in collection, and related content on DH items, person records, places, and dictionary words.

## Test

- [ ] Create any content.
- [ ] Edit said content.
- [ ] Verify that the content doesn't show up in any entity reference fields on it's own form.

Pre-merge checklist:


christa-kavanagh-wsu (approved):
This is working for me!

michael-wynne-wsu:
## Accessibility Check: PASS

No frontend changes (no Twig, HTML, CSS, or JS modified). The change is purely server-side query filtering and has no impact on any WCAG 2.2 AA criteria.

---

## PR Review: 776 — Exclude current entity from content browser

**Blast Radius:** LOW — single `hook_views_query_alter()` block, scoped to `mukurtu_content_browser` view only, no cross-module dependencies affected.
**Breaking Changes:** None — this is an additive filter; existing behavior unchanged when `original_path` is absent.
**Tests:** No tests added (consistent with the rest of this module).
**Update hooks:** Not required — no schema, config, or data changes.

---

### LOOKS GOOD

- **SQL injection safety:** `original_path` is user-supplied (query string), but only the regex-captured digits are used, cast to `(int)` before passing to `addWhere()`. Drupal Views' `addWhere()` uses parameterized queries internally. No risk.
- **Graceful degradation:** If `original_path` is absent or doesn't match `/node/{id}/`, the condition is silently skipped and the browser returns full results. Correct behavior.
- **Scope is tight:** The `$view->id() === 'mukurtu_content_browser'` guard ensures this only fires for the one targeted view.
- **`addWhere(0, ...)` group:** Consistent with existing calls in the same function; group 0 is the default AND group in Views queries.
- **Docblock updated** to reflect the new behavior.
- **Comment explains the non-obvious:** The note that `original_path` comes from the entity browser Modal display plugin is valuable context that would otherwise require reading the contrib module source.

---

### SUGGESTIONS (non-blocking)

**1. Regex requires a trailing slash — `/node/42` won't match**

The pattern `#^/?node/(\d+)/#` requires a `/` after the node ID. `/node/42/edit` matches, but bare `/node/42` does not. This is almost certainly fine in practice (the content browser opens from edit forms, not view pages), but worth a brief comment so future readers don't wonder if it's a bug.

**2. `\Drupal::request()` — consider the service container**

Calling `\Drupal::request()` as a static in a hook is the standard Drupal procedural pattern, so this is not a blocker. Just noting that it limits unit testability if tests are ever added for this hook. No action needed.

---

### CHECKLIST SUMMARY

| Area | Status |
|---|---|
| Scope matches issue #776 | PASS |
| No unrelated changes | PASS |
| SQL injection | PASS |
| Hardcoded secrets | PASS |
| Input validation | PASS (int cast + parameterized) |
| Breaking changes | PASS |
| DB/config/update hooks | PASS (none needed) |
| Performance (N+1, unbounded loops) | PASS |
| Blast radius | LOW |

**Ready to merge** after updating from main.

michael-wynne-wsu:
Added the clarifying comment about the intentional trailing-slash requirement. The `\Drupal::request()` suggestion was "no action needed" so I left that as-is.

joelle-birano-wsu (approved):
Works for me!

nick-deer (approved):

joelle-birano-wsu:
Testing post-merge - this seems to be working as expected.

PR #1643, opened 2026-05-18
1642 updates to create protocol workflow

### LINKED ISSUE #1642: Updates to create protocol workflow
- [ ] `/protocols/protocol/add` 
   - [ ] Community field should be multi-value.
   - [ ] Community managers need access to this form.
- [ ] `/communities/community/1/add/protocol`
   - [ ] Expose community field (multi-value).

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1642

- Fix route access for /protocols/protocol/add: replace static Drupal permission check with _entity_create_access so community managers (who hold the OG create protocol protocol permission) can reach the form
- Make the community field multi-value (select multiple) on both the standalone and community-bound protocol creation forms
- Expose the community field on /communities/community/{id}/add/protocol, pre-selected with the URL community, allowing additional communities to be attached at creation time
- Aggregate member suggestions across all selected communities

Pre-merge checklist:


michael-wynne-wsu:
Joelle, I still need to make one small change, but this fixes the issues you flagged.

joelle-birano-wsu:
#### Create community and protocol workflow 

- We didn't need the add communities here right? It's forcing me to select a community which we don't have to do in this form.
- This "update member list" button is also displaying

Screenshot 2026-05-18 at 11 22 51 AM

#### Protocols/protocol/add

- after selecting communities, they don't show up in the form (it just looks like no communities were added), but they ARE selected, the form saves and they do show up on the community and protocol pages.
- Update member list button also displays

#### communities/community/ID/add/protocol
- Same problems as protocols/protocol/add
- URL community is not pre-selected
Screenshot 2026-05-18 at 12 20 26 PM

michael-wynne-wsu:
Oh that came out pretty different from my local. Will figure that out and re-ping when ready...

michael-wynne-wsu:
@joelle-birano-wsu try now. community should be pre-populated if coming from that route, selected communities are displayed, and instead of that active "update memberships" it should trigger automatically.

joelle-birano-wsu:
- The update membership button is no longer displaying, so that looks good.
- For community-tethered protocols forms, the community displays as pre-populated, but it isn't registering in the form -  I can't save the form because it doesn't see the community. If I were to select another community, the protocol would be added to that community and NOT the parent. It wouldn't be a shared community.

Screenshot 2026-05-18 at 3 59 30 PM

- For ALL protocol forms, adding a community still doesn't display them in form.

michael-wynne-wsu:
@joelle-birano-wsu I pushed an update here, let me know if it's better.

joelle-birano-wsu:
Getting better! 

- The preselected community is enough to save a new protocol form if I'm not adding any other communities. Good!

- If I AM adding other communities in a community-tethered protocol form the modal doesn't seem to be seeing the pre-selected community. If I want to make a shared protocol, I have to add the preselected community along with any other communities. 

- Selected communities are still not displaying in all forms.

michael-wynne-wsu:
OKAY new version. This took some doing but I think I've got it...

joelle-birano-wsu:
- All forms seem to be working well. Pre-selected communities are there and saving as expected. Selected communities are displaying, so all that is great!

- One weird thing that's happening now is the checkboxes in the community selection modal have disappeared. It's still perfectly functional - I can select multiple communities and they save just fine. Sometimes they appear for a second and then disappear. Other times they don't display at all. 
   - protocols/protocol/add - doesn't flash the checkboxes at all
   - community/ID/add/protocol AND protocols/protocol/add/community/ID- checkboxes appear then disappear

Screenshot 2026-05-20 124040
ear.

michael-wynne-wsu:
@joelle-birano-wsu I wasn't happy with the appearance and behaviour of having to click the tiny checkbox so I just made the whole "card" clickable. I'll see if I can sort out the checkbox appearing first...

joelle-birano-wsu:
> @joelle-birano-wsu I wasn't happy with the appearance and behaviour of having to click the tiny checkbox so I just made the whole "card" clickable. I'll see if I can sort out the checkbox appearing first...

I like it! It's nice to be able to click on the whole card.

michael-wynne-wsu:
Okay that flashed checkbox should be sorted.

joelle-birano-wsu (approved):
This looks good! I'm not seeing any other problems!

michael-wynne-wsu:
## WCAG 2.2 AA Accessibility Audit

### Blocking Issues

**Issue 1 - WCAG 2.1.1 Keyboard + 4.1.2 Name/Role/Value: Community browser cards are not keyboard operable**
`modules/mukurtu_gin_custom/js/community-browser-select.js`

The `.views-row` click handler only fires on mouse clicks. Keyboard users who tab to a card and press Enter/Space cannot select it. The underlying checkbox is hidden with `display: none`, removing it from the tab order entirely. No `tabindex`, `role`, or `aria-checked` is present on the card elements.

Fix needed:
```js
// After init, make rows focusable with an accessible role
$view.find('.views-row').each(function () {
  $(this).attr({ tabindex: '0', role: 'checkbox', 'aria-checked': 'false' });
});

// Handle both click and keyboard
$view.on('click keydown', '.views-row', function (e) {
  if (e.type === 'keydown' && e.key !== 'Enter' && e.key !== ' ') return;
  if (e.type === 'keydown') e.preventDefault();
  if ($(e.target).is('input')) return;
  var $cb = $(this).find('.views-field-entity-browser-select input');
  var checked = !$cb.prop('checked');
  $cb.prop('checked', checked);
  $(this).toggleClass('is-selected', checked).attr('aria-checked', String(checked));
});
```

**Issue 2 - WCAG 1.3.1 Info and Relationships: Checkbox hidden from accessibility tree with `display: none`**
`modules/mukurtu_gin_custom/css/mukurtu-gin-custom.css`

The `.views-field-entity-browser-select` column is hidden with `display: none`, which removes the checkbox from the accessibility tree before JS runs. After JS runs it remains hidden with no accessible equivalent. Use `visually-hidden` CSS instead (or add `aria-hidden="true"` on the checkbox from JS once the row becomes the accessible control):

```css
.view-mukurtu-community-select .views-field-entity-browser-select {
  position: absolute;
  width: 1px;
  height: 1px;
  overflow: hidden;
  clip: rect(0 0 0 0);
  white-space: nowrap;
}
```

**Issue 3 - WCAG 2.4.7 Focus Visible: No `:focus-visible` ring on unselected cards**
`modules/mukurtu_gin_custom/css/mukurtu-gin-custom.css`

The `.is-selected` outline only appears on selected cards. Keyboard users on an unselected card have no visible focus indicator. Add:

```css
.gin-entity-browser .entity-browser-form .views-row:focus-visible {
  outline: 3px solid var(--gin-color-primary, #0036b1);
  outline-offset: -3px;
}
```

**Issue 4 - WCAG 1.3.1 Info and Relationships: "Protocol members" heading has no semantic role**
`modules/mukurtu_protocol/src/Form/ProtocolAddForm.php`

The membership section heading is a `#type => 'item'` with `#title => 'Protocol members'`, which renders as a `<div>`, invisible to screen reader heading navigation. Use `#type => 'html_tag'` with `#tag => 'h3'` or add `role="group"` with `aria-labelledby` on the container.

### Non-Blocking Issues

**Issue 5 - WCAG 2.5.8 Target Size: Toggle button minimum size not enforced**
`modules/mukurtu_gin_custom/js/entity-browser-toggle.js` / CSS

The JS-created toggle buttons have no guaranteed 24x24px touch target. Ensure `.eb-view-toggle__btn` has `min-height: 24px; min-width: 24px` in CSS.

**Issue 6 - WCAG 4.1.2: Remove button name falls back poorly on empty display name**
`modules/mukurtu_protocol/src/Form/ProtocolAddForm.php`

If `$member->getDisplayName()` returns empty, the `aria-label` becomes `"Remove "`. Add a fallback:
```php
'aria-label' => t('Remove @name', ['@name' => $name ?: t('user #@uid', ['@uid' => $uid])]),
```

---

## PR Review: 1642 Updates to Create Protocol Workflow (#1643)

**Blast Radius:** MEDIUM - mukurtu_protocol and mukurtu_gin_custom modules; touches protocol creation form, community browser, routing, and query access control.

**CI Status:** FAILING - `Mukurtu v4 - PHP 8.3`

**Security:** No critical findings

**Tests:** No new test coverage added for new behavior

---

### MUST FIX (Blocking)

**1. CI: Protocol kernel tests are failing**

The CI run shows 5 failures in `ProtocolEntityAccessTest` and `AccessByProtocolTest`, and 4 failures in `ManageBulkRolesFormTest`. These are in `mukurtu_protocol`, the module this PR modifies. The `ImportListStringTest` failures are likely pre-existing (unrelated module). The protocol test failures need investigation - they test access scenarios with `community_manager` roles, which could be affected by the new query alter hook or form restructuring even indirectly. The PR should not merge with failing kernel tests in its own module.

**2. WCAG: Card selection widget is not keyboard operable (Issues 1-3 above)**

The community browser cards cannot be selected via keyboard. This is a blocking accessibility issue per CLAUDE.md workflow requirements.

**3. WCAG: Missing focus ring on unselected cards (Issue 3 above)**

---

### SHOULD FIX (Non-Blocking)

**4. Update hook 40030 replaces entire view config**

`modules/mukurtu_protocol/mukurtu_protocol.install`:
```php
\Drupal::configFactory()->getEditable('views.view.mukurtu_community_select')
  ->setData($view_data)->save(TRUE);
```
`setData()` replaces the full config object, which would overwrite any admin UI customizations to the view made since the last deployment. Consider using `set('display.default.display_options.query.options.query_tags', [...])` to surgically add the query tag only.

**5. Query alter hook silently no-ops if table alias is not found**

`modules/mukurtu_protocol/mukurtu_protocol.module` - the hook returns early if `$community_table` is empty, which means if the view structure changes and the alias lookup fails, the filter silently disappears and all community managers see all communities. A `\Drupal::logger()` warning would help catch this in non-prod.

**6. ProtocolAddForm: Member user list is no longer filtered by community**

The PR description says members are now "all active users" instead of filtering by community membership. This is a significant UX regression for sites with large user bases. Is this intentional, or should the member list still be scoped to users in the selected communities?

**7. No tests for the community manager filter**

The new `mukurtu_community_manager_select` query tag behavior (filtering the community browser to communities managed by the current user) has no test coverage. Given this is access control behavior, a kernel test asserting that community managers only see their own communities in the browser would prevent regressions.

---

### SUGGESTIONS

**8. `protocol-community-browser.js`: The `eb-target` input selector may be fragile**

```js
var entityIds = $btn.closest('form').find('input.eb-target').val();
```

The `eb-target` class is entity browser internal API. If entity_browser changes this class in a future update, the guard logic silently breaks. Consider adding a comment noting the dependency on this internal class.

**9. Routing: Verify `_entity_create_access` bundle resolution for community-bound route**

The `mukurtu_protocol.protocol_create` route change is correct for the standalone form. Verify the community-bound route (`/communities/.../add/protocol`) also uses entity-level access or inherits correctly, so community managers can reach both entry points.

---

### LOOKS GOOD

- The `_entity_create_access: 'protocol'` routing change is the correct Drupal pattern for OG-based create permissions. Well reasoned.
- `protocol-community-browser.js` handles the entity browser dual-event-fire edge case with the `eb-target` guard and `setTimeout(0)` deferral - this is non-obvious behavior handled correctly.
- `entity-browser-toggle.js` uses `aria-pressed`, `fieldset/legend`, `type="button"`, and `localStorage` keyed by field name - solid accessible toggle implementation.
- SQL joins in the query alter hook use Drupal's OOP query builder (no string interpolation), safe from injection.
- The `ProtocolAddForm` entity browser value parsing (`entity_type:id` format) is handled correctly in `buildEntity()`.
- Update hook numbering (40030) and structure follow project conventions.

---

**Summary:** Two blocking issues need resolution before merge - the failing kernel tests in `mukurtu_protocol` (CI) and the keyboard inaccessibility of the community browser cards. The `membership section heading` WCAG fix is also required per the pre-merge checklist. The update hook behavior (replacing full view config) and missing member filter are worth discussing before shipping.

michael-wynne-wsu:
Here's a summary of all fixes made:

| # | What | File | Status |
|---|------|------|--------|
| 1 | Keyboard operability for community browser cards (Enter/Space, tabindex, role, aria-checked) | community-browser-select.js | Done |
| 2 | Replace `display:none` with visually-hidden CSS on checkbox column | mukurtu-gin-custom.css | Done |
| 3 | Add `:focus-visible` ring for unselected cards | mukurtu-gin-custom.css | Done |
| 4 | Replace `#type => 'item'` heading with `<h3>` for "Protocol members" | ProtocolAddForm.php | Done |
| 5 | Remove button aria-label fallback for empty display names | ProtocolAddForm.php | Done |
| 6 | Update hook 40030: surgical query tag set instead of full config replace | mukurtu_protocol.install | Done |
| 7 | Restore community-filtered member suggestions (aggregated across all selected communities) | ProtocolAddForm.php | Done |
| 8 | Logger warning when query alter hook silently no-ops | mukurtu_protocol.module | Done |
| 9 | Comment explaining `eb-target` internal API dependency | protocol-community-browser.js | Done |
| CI | Test failures in ManageBulkRolesFormTest/ProtocolEntityAccessTest | Pre-existing, no files in this PR changed those test classes | Pre-existing |

christa-kavanagh-wsu (approved):
This works great! I like the whole card being clickable and being able to choose multiple communities at protocol creation is really useful.

nick-deer (changes requested):
Looks great, only thing I noticed is I can't tab communities on the select communities modal

michael-wynne-wsu:
Hopefully this helps

The hidden checkboxes inside each community card were still natural tab stops. Even though they were visually hidden with `clip`, they remained focusable — so Tab was landing on invisible inputs between cards, making it appear that communities couldn't be tabbed to.

The fix removes the hidden checkboxes from both the tab order (`tabindex="-1"`) and the accessibility tree (`aria-hidden="true"`), since the `.views-row` element with `role="checkbox"` is already the intended accessible target. Tab will now move cleanly between community cards.

nick-deer:
It's still not working for me... I can only tab if I click on a community card, and even then I can only tab to the end of the list and it doesn't ever loop back to the community cards at the start

michael-wynne-wsu:
@nick-deer AHA it's working now, I just tested on tugboat.

nick-deer (approved):
Nice it's working now!

PR #1652, opened 2026-05-21
Add site-level toggle for media download button (issue #120)

### LINKED ISSUE #120: Enable/disable "access [media]" buttons
Provide controls to enable or and disable the v3 access media buttons. Sometimes it's useful, but more often than not it confuses people.

michael-wynne-wsu:
Closes #120

# Summary

- Adds a 'Media Download Button' setting to the Mukurtu Media admin form (/admin/config/mukurtu/media-download-settings) allowing site admins to globally show or hide the download button on media assets.
- Defaults to enabled to preserve existing behavior for all current sites.

## Follow-up issues

- #1654 - Restructure /admin/config/mukurtu/settings
- #1655 - Extend media download toggle to additional content types

Pre-merge checklist:


christa-kavanagh-wsu (commented):
- The media download settings work for DH items. 
- I think we should consider adding them to other content types, as it doesn't make a lot of sense to have some media assets downloadable and some not depending on content type instead of permission.
- I am not sure what these settings do...I tried changing them and it didn't really seem to do anything? I also thought that at one point we discussed not having the right-hand related content teaser?

Image

michael-wynne-wsu:
> I think we should consider adding them to other content types, as it doesn't make a lot of sense to have some media assets downloadable and some not depending on content type instead of permission.

Agreed. The big question is which fields per content type?

> I am not sure what these settings do...I tried changing them and it didn't really seem to do anything? I also thought that at one point we discussed not having the right-hand related content teaser?

They're from earlier development that we decided against including for now, hence my note about restructuring that page.

joelle-birano-wsu:
>Adds a 'Media Download Button' setting to the Mukurtu Settings admin form (/admin/config/mukurtu/settings) allowing site admins to globally show or hide the download button on media assets.

This works well.

Would it be useful/feasible to have the global setting, but then allow for an override in a content item? Or even in the media asset itself? I could see myself potentially wanting certain media assets not to be downloadable, but others I might not care about as much...

>Defaults to enabled to preserve existing behavior for all current sites.

It's defaulting to "Hide download button"

> ## To do
> * [ ]  Decide if we need to restructure `/admin/config/mukurtu/settings` since it's a weird mix of settings.

If we can split these settings up I think it makes sense to put related content settings in its own separate page under Content Settings, and media download settings could go under Media.

> * [ ]  This seems to only be scope to DH item media assets fields. That's pre-existing in that only DH had the download button. Need to decide if we want to add this to all/select media-type fields across all/more content types.

I think it makes sense to expand the scope to more content types. Not collections/word lists, and probably only media uploaded to the site, and just core media assets, not thumbnails.

christa-kavanagh-wsu:
> > I think we should consider adding them to other content types, as it doesn't make a lot of sense to have some media assets downloadable and some not depending on content type instead of permission.
> 
> Agreed. The big question is which fields per content type?

I have mixed feelings about this, and I think it would come down to "what have folks asked for"? My gut reaction is to allow media assets and not metadata fields, but...I can definitely see reasons to allow users to download media assets and metadata. Specifically formatted text fields from DH items, the biography sketches and text sections from person and place records, and maybe mapping and birth/death info, but I think applying a universal setting to allow folks to do that might be a bit too open. 

I like Joelle's idea about being able to override downloads per content item on principle, but that could be a giant pain to manage and has the potential to expose things that users would prefer were not exposed.

michael-wynne-wsu:
- I'm not inclined to do more than a single site-wide on/off setting at this time. If there is request for more granular control later it can be expanded.
- Metadata download is out of scope for this. If needed, that would be a new feature request.
- I'll see if I can find another settings page to roll this into, otherwise I'll make it a standalone form.
- This does only apply to uploaded media, yes, so images, audio, documents, and video. Embed code, souncloud, vimeo, and youtube are out of scope.
- As for where this is available, I would say....
  - All content with a "media assets" field: DH, person, place, word.
  - Unique to words: recording, sample sentence.
  - None of the featured/thumbnail fields: collection, word list, word (thumbnail).
  - Only on full content pages, never previews, teasers, related content, search results, etc etc

christa-kavanagh-wsu:
> * I'm not inclined to do more than a single site-wide on/off setting at this time. If there is request for more granular control later it can be expanded.
> 
>     * Metadata download is out of scope for this. If needed, that would be a new feature request.
> 
>     * I'll see if I can find another settings page to roll this into, otherwise I'll make it a standalone form.
> 
>     * This does only apply to uploaded media, yes, so images, audio, documents, and video. Embed code, souncloud, vimeo, and youtube are out of scope.
> 
>     * As for where this is available, I would say....
>       
>       * All content with a "media assets" field: DH, person, place, word.
>       * Unique to words: recording, sample sentence.
>       * None of the featured/thumbnail fields: collection, word list, word (thumbnail).
>       * Only on full content pages, never previews, teasers, related content, search results, etc etc

All of that sounds good!

joelle-birano-wsu:
> * I'm not inclined to do more than a single site-wide on/off setting at this time. If there is request for more granular control later it can be expanded.
> * Metadata download is out of scope for this. If needed, that would be a new feature request.
> * I'll see if I can find another settings page to roll this into, otherwise I'll make it a standalone form.
> * This does only apply to uploaded media, yes, so images, audio, documents, and video. Embed code, souncloud, vimeo, and youtube are out of scope.
> * As for where this is available, I would say....
>   
>   * All content with a "media assets" field: DH, person, place, word.
>   * Unique to words: recording, sample sentence.
>   * None of the featured/thumbnail fields: collection, word list, word (thumbnail).
>   * Only on full content pages, never previews, teasers, related content, search results, etc etc

Yeah that makes sense!

michael-wynne-wsu:
- It is now a standalone "Media download settings" form, in the `Media` section of the dashboard.
  - Accessible by administrator and mukurtu manager roles.
- Download button now on:
  - All content with a "media assets" field: DH, person, place, word. Should all be identical to the DH item carousel/media. Added an icon to the button.
  - Unique to words: recording, sample sentence. This one is just an icon and inline with the audio.

joelle-birano-wsu:
Overall this is looking good. The only thing I spotted was the download video button isn't centered and doesn't have the same padding as the others. Applies to all content items.

Screenshot 2026-05-21 120927

Screenshot 2026-05-21 121410

michael-wynne-wsu:
Good catch @joelle-birano-wsu Try now

joelle-birano-wsu:
> Good catch @joelle-birano-wsu Try now

Looks good now!

joelle-birano-wsu (approved):

michael-wynne-wsu:
## PR Review: Add site-level toggle for media download button (#1652)

**Blast Radius:** LOW-MEDIUM - touches config, templates, theme preprocess, and a new admin form. No shared libraries or API contracts changed.
**Tests:** No new tests for new PHP class (pre-existing pattern in this repo). CI failures are pre-existing on main, not caused by this PR.
**Breaking Changes:** View mode change for Person/Place (`full` → `media_assets`) could affect sites that have existing content using those types.

---

### MUST FIX (Blocking)

**1. Cache invalidation will not work as written**

`mukurtu_v4.theme` reads `mukurtu.settings` config in the preprocess hook but never adds the config's cache tags to the render array. The `Cache::invalidateTags(['media_view'])` call in `MediaDownloadSettingsForm.php` can only work if media render arrays actually carry the `media_view` tag - they don't by default in Drupal core, and nothing in the preprocess hook adds it.

**The result:** Toggling the setting will not clear cached pages showing media. The download buttons will keep appearing (or not appearing) until cache expires on its own.

**Fix in `mukurtu_v4.theme`:** Add the config's own cache metadata to the render array:

```php
$config = \Drupal::config('mukurtu.settings');
$variables['media_download_enabled'] = $config->get('mukurtu_media_download_enabled') ?? TRUE;
$variables['#cache']['tags'] = Cache::mergeTags(
  $variables['#cache']['tags'] ?? [],
  $config->getCacheTags()
);
```

Then in `MediaDownloadSettingsForm::submitForm()`, remove the manual `Cache::invalidateTags(['media_view'])` - it becomes unnecessary because Drupal automatically invalidates `config:mukurtu.settings` when the config is saved via `->save()`.

---

### SHOULD FIX (Non-blocking)

**2. PR description route path is wrong**

The PR body says the setting is at `/admin/config/mukurtu/settings` but the actual route is `/admin/config/mukurtu/media-download-settings`. Minor, but should be corrected before merge so the description is accurate in git history.

**3. `mukurtu.settings` schema lives in the wrong module**

The config object `mukurtu.settings` is defined in `mukurtu_core/config/schema/mukurtu_core.schema.yml` but is read/written by `mukurtu_media`. This means any module can add keys to `mukurtu.settings` against the `mukurtu_core` schema, which will silently fail schema validation for new keys. Consider either:
- Moving the schema definition to `mukurtu_media` (if `mukurtu.settings` is media-specific), or
- Documenting this as an intentional shared settings object.

**4. Pre-merge checklist items remain open**

The PR has two unresolved TODO checkboxes that should be converted to separate GitHub issues before merge to avoid them going dark:
- Restructuring `/admin/config/mukurtu/settings`
- Extending the download toggle to other content types beyond DH items

---

### SUGGESTIONS

**5. `media--audio--default.html.twig` and `media--audio--audio-for-dictionary-teaser.html.twig` are identical**

Both new templates have the same content. If they're expected to stay in sync, consider whether one should extend/include the other, or at minimum note in a comment why they're duplicated.

**6. Video download link uses a different URL pattern than other types**

`media--video--media-assets.html.twig` still uses `file_url(media.field_media_video_file.entity.fileuri)` directly instead of the preprocessed `media_url` variable. This is pre-existing, but it's now visually inconsistent with the other four templates that all use `{{ media_url }}`. Not a bug, but worth standardizing.

---

### LOOKS GOOD

- Permission definition, routing, and role assignment are all correctly wired and consistent with each other
- `restrict access: true` on the permission is correct for an admin setting
- `?? TRUE` fallback in preprocess means existing sites with no config key continue showing download buttons (safe default)
- All 6 templates correctly use `aria-hidden="true"` on the SVG icons
- Icon-only buttons in dictionary templates use descriptive `aria-label="Download audio: {{ name }}"` with media item name
- Touch targets for `.button--download-icon` are 44x44px - meets WCAG 2.5.8
- The two pre-existing WCAG failures (hover contrast #159ec4 at 3.13:1, focus outline #66afe9 at 2.37:1 on white) are in the shared button mixin and outside this PR's scope - recommend filing a separate issue
- CI failures (7 errors, 6 failures in `mukurtu_protocol` and `mukurtu_import` kernel tests) are pre-existing on `main` and caused by a missing `image` field plugin in the test bootstrap - not related to this PR

---

**Summary:** One must-fix (cache metadata) and two should-fixes before merge. The core feature logic is correct and the permission model is solid.

michael-wynne-wsu:
Both changes are in. Here's a summary of what was implemented:

**Cache fix (`themes/mukurtu_v4/mukurtu_v4.theme`):**
- Added `use Drupal\Core\Cache\Cache;` import
- After reading the config, adds `config:mukurtu.settings` to the render array's cache tags via `Cache::mergeTags()` + `$config->getCacheTags()`. Now when the setting is saved, Drupal's built-in config cache invalidation automatically clears all media render arrays that carry that tag.

**Form cleanup (`modules/mukurtu_media/src/Form/MediaDownloadSettingsForm.php`):**
- Removed the `use Drupal\Core\Cache\Cache;` import (no longer needed)
- Removed `Cache::invalidateTags(['media_view'])` - the `->save()` call on the config object already triggers `config:mukurtu.settings` invalidation automatically, making the manual call redundant and incorrect

All should-fix items are done:

- **Video template** - download link now uses `{{ media_url }}` consistent with all other media-assets templates
- **Schema location** - `mukurtu.settings` moved from `mukurtu_core.schema.yml` to `mukurtu_media.schema.yml` where it belongs
- **Duplicate templates** - added comments to each explaining they're intentionally separate for independent styling of `field_recording` vs `field_sentence_recording` contexts
- **Open TODOs** - converted to GitHub issues #1654 (settings restructure) and #1655 (extend toggle scope)
- **PR description** - corrected route path from `/admin/config/mukurtu/settings` to `/admin/config/mukurtu/media-download-settings` and linked the new issues

christa-kavanagh-wsu:
If you can download the external resource there is that option with a black arrow, otherwise there is another download option with a white arrow. On the "Access xyz option, the text looks a little wonky, almost like it's not quite centered in the red box. The download option for external resources is left justified, while the download and access options for other media is centered.
image
image

michael-wynne-wsu:
weird I wonder why soundcloud shows up as downloadable. I will explore that!

christa-kavanagh-wsu:
That was a little weird. The external embed/ESRI storymap was also showing up as downloadable for me.

michael-wynne-wsu:
All external media download/access buttons are gone.

michael-wynne-wsu:
## PR Review: feature/1092-pdf-viewer-improvements

**Branch vs main — 21 files changed, ~583 lines**

**Blast Radius:** MEDIUM — media templates affect all content types that embed media; config permission affects Mukurtu Manager role on install  
**Security:** No findings  
**Breaking Changes:** None (the `and media_url` guard in generic template fixes a broken-button regression, not a break)  
**Tests:** No new tests (Twig templates — expected)

---

### MUST FIX (Blocking)

**1. `<object alt="…">` — invalid HTML, AT receives nothing**
media--document--media-assets.html.twig:43

The `alt` attribute is not valid on `<object>` and is silently ignored by browsers and screen readers. Add a `title` attribute and inner fallback content:
```twig
<object data="{{ media_url }}" title="Document: {{ name }}" type="application/pdf" style="width: 840px; height: 500px; display: block;">
  <p><a href="{{ media_url }}">{{ 'Open document'|t }}</a></p>
</object>
```

**2. Inline `style` on `<object>` — fails WCAG 1.4.10 Reflow**
media--document--media-assets.html.twig:43

`width: 840px` hard-codes a width that breaks reflow at 400% zoom. Move to SCSS using `inline-size: 100%` (responsive) and `block-size: min(500px, 70vh)`.

---

### SHOULD FIX (Non-blocking)

**3. `.button--download-icon` lives in `_dictionary-word.scss` but is used in audio media templates**
_dictionary-word.scss:226 / media--audio--audio-for-dictionary-teaser.html.twig

This class is applied outside the dictionary context (audio media in general), so it should live in `_media.scss` alongside `.button--download`. The current location will confuse anyone styling media components.

**4. Two new audio templates are identical in markup**
media--audio--audio-for-dictionary-teaser.html.twig vs media--audio--default.html.twig

The docstring for `default` says it's for `field_sentence_recording`, but `default` is a global view mode — it will match ALL audio media rendered in the default view mode, not just sentence recordings. If these two contexts are meant to look the same right now but diverge later, that's fine — but the docstrings need to accurately describe what each view mode is, not what field they appear in.

**5. Dictionary audio templates missing `and media_url` guard**
media--audio--audio-for-dictionary-teaser.html.twig:24

The generic fallback correctly has `{% if media_download_enabled and media_url %}` but the two new dictionary/default audio templates only check `{% if media_download_enabled %}`. Audio is always file-based so `media_url` will be set in practice — but for consistency and safety, the guard should match.

**6. No install config for `mukurtu.settings`**
mukurtu_media.schema.yml

There's no `config/install/mukurtu.settings.yml` to seed the default value. The code falls back to `?? TRUE` in both the form and the preprocess hook, so it works — but if the config object is ever wiped and re-imported from sync, the key won't exist until the form is saved. Consider adding `config/install/mukurtu.settings.yml` with `mukurtu_media_download_enabled: true`.

---

### SUGGESTIONS

**7. Trailing blank line in remote video template**
media--remote-video--media-assets.html.twig

The removed "Access video" line was replaced with a bare newline, leaving double blank lines before `</div>`. Minor — just clean it up.

**8. `mukurtu.settings` is a shared config object**
The schema key `mukurtu.settings` is defined in `mukurtu_media.schema.yml` but the namespace suggests it's a general Mukurtu-wide settings bag. If other modules also write to `mukurtu.settings`, the schema definition may conflict. Consider `mukurtu_media.settings` as the config name to stay within module scope, or confirm that `mukurtu.settings` is the intentional shared object.

---

### LOOKS GOOD

- All decorative SVGs correctly use `aria-hidden="true"`
- Icon-only download links use `aria-label="Download audio: {{ name }}"` — correct for WCAG 4.1.2
- `media_download_enabled` removes buttons from the DOM entirely (not just `display:none`) — correct, avoids hidden focusable elements
- Route is gated by `_permission` with `restrict access: true` — correct
- Cache tags are properly merged on config change — correct
- `media--video--media-assets` updated from raw Twig `file_url()` call to preprocessed `media_url` — more consistent, good cleanup
- Person/Place view display switch from `full` to `media_assets` — correct alignment with other content types
- `field_contributor` moved from hidden to visible in audio default display — correct, displayed with `entity_reference_label`

---

**Summary:** Two accessibility fixes are blocking (the `<object>` label and the responsive width). The `.button--download-icon` SCSS location and the install config gap are the most important non-blocking items. Everything else is clean and well-scoped.

michael-wynne-wsu:
All done. Here's a summary of everything fixed:

| # | Issue | Fix |
|---|-------|-----|
| 1 | `<object alt="…">` invalid HTML | Replaced with `title=` + inner `<a>` fallback in media--document--media-assets.html.twig |
| 2 | Inline `style` fixed 840px width | Moved to `.media--document__viewer` in _media.scss with `inline-size: 100%; block-size: min(500px, 70vh)` |
| 3 | `.button--download-icon` in wrong SCSS file | Moved from _dictionary-word.scss to _media.scss |
| 4 | Audio template docstrings misleading | Fixed in both media--audio--default.html.twig and media--audio--audio-for-dictionary-teaser.html.twig |
| 5 | Missing `and media_url` guard in audio templates | Added to both audio templates |
| 6 | No install config for settings | Created mukurtu_media.settings.yml |
| 7 | Trailing blank line in remote video template | Cleaned up |
| 8 | Config key `mukurtu.settings` too broad | Renamed to `mukurtu_media.settings` in schema, form, and theme preprocess |

SCSS compiled cleanly. Ready for a commit when you are.

christa-kavanagh-wsu (commented):
This PR is generally functioning as expected, but I ran into this issue https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1676 while testing.

nick-deer (approved):

christa-kavanagh-wsu:
Tested post-merge - the media download button toggle works, but the remote video "Access video" button is no longer showing (though the SoundCloud button is still there). It still has the "link is external" icon:
image

PR #1659, opened 2026-05-22
906 Stabilize media carousel height when navigating between media of …

### LINKED ISSUE #906: Media thumbnail and field sizing in carousel
Media types in the carousel appear with all different sized thumbnails. When you scroll through media assets, the media asset field resizes with each different media asset, causing you to have to navigate to the scrolling arrow each time you select it. The embed thumbnail is weirdly large and breaks the boundaries of the field. 

Image

Image

michael-wynne-wsu:
…different heights

Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/906

Pre-merge checklist:


michael-wynne-wsu:
## Issue #906 - Media Carousel Height Stability

**The problem:** Splide's fade-type main carousel dynamically resizes its track to match the active slide's height. When navigating between a tall image and a short audio player (which already gets padded to 300px via CSS), the entire carousel jumps vertically and drags the thumbnail strip with it.

**Why there's no perfect universal solution:**
- Remote video/embed iframes don't expose their natural height until they load - no amount of CSS or JS at mount time can know that height in advance
- Images have no shared aspect ratio
- Audio players are inherently very short

---

## Recommended: JS max-height equalization + CSS smooth transition

**Two-part fix:**

1. **JS - compute max height at mount** (`media-asset-carousel.js`): After Splide mounts, read `scrollHeight` on every slide (this reads through the `block-size: 0` CSS rule that hides inactive slides), find the tallest, and set that as `min-block-size` on the track. Wire to `resized` as well so it re-equalizes on viewport changes.

2. **CSS - smooth transition** (`_carousels.scss`): Add `transition: block-size 0.3s ease-in-out` to `.media-carousel .splide__track`. This handles the one case JS can't fix - async-loading remote embeds - by animating the height change instead of letting it snap.

**Residual limitation:** A remote video iframe that loads slowly will still cause a one-time height change after the carousel is already visible. The transition makes it look intentional rather than broken.

michael-wynne-wsu:
@christa-kavanagh-wsu This may be an improvement, but it's a little hard for me to tell. I think you had the strongest opinion on this behaviour so let me know what you think. The writeup above gives sueful context.

joelle-birano-wsu:
Seems like it's a little better. It still jumps around a bit but not as much. I think I'm ok with that.

joelle-birano-wsu (approved):

michael-wynne-wsu:
Yeah unless we limit the height it will be variable. So either taller stuff gets scaled small or small stuff has extra space somewhere above/below. 

OR we move the carousel. But if it's above I don't like it that close to CR tabs and/or MPI pages.

christa-kavanagh-wsu (commented):
This works much better on DH items and dictionary words. It is still having some problems on person/place records; I have to down scroll to find the carousel button for PDFs and embeds when I'm clicking through.

michael-wynne-wsu:
> This works much better on DH items and dictionary words. It is still having some problems on person/place records; I have to down scroll to find the carousel button for PDFs and embeds when I'm clicking through.

@christa-kavanagh-wsu I think the person/place record stuff is resolved by the glightbox PR at https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1672 since on this branch the lightbox isn't configured properly. I'm going to move this on to code review since there are no other issues and will check on it before final merge.

christa-kavanagh-wsu (approved):

nick-deer (approved):

PR #1660, opened 2026-05-22
320 audio thumbnails

### LINKED ISSUE #320: Audio media asset thumbnails
- [x] Provide a standard default thumbnail for audio files (to be used in MPI carousels, potentially elsewhere). 
   - [ ] Implement palette-specific thumbnails (these have been created)
- [ ] Allow site admins to replace the default audio thumbnail with their own custom thumbnail (site-wide).
- [x] If a user provides an thumbnail for a specific audio file (this is already an optional field), that should take precedence over the other two.

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/320

## Summary

- [ ] New config page at `/admin/config/system/mukurtu-media-thumbnail-settings` (in Media section of dashboard).
- [ ] A new thumbnail fallback scheme:
   - [ ] Uploaded or generated per-media. If none,
   - [ ] Admin provided media type (at that new config). If none,
   - [ ] Our default greyscale thumbnail (I prefer this to the theme coloured ones).
- [ ] Focused on audio originally, but SHOULD apply to all media types.

Pre-merge checklist:


michael-wynne-wsu:
I haven't done more than a superficial test. I won't be surprised if there are lots of gaps.

michael-wynne-wsu:
Done. Both types now appear on the settings page. Just to set expectations:

- **Remote video** — the carousel uses the oEmbed auto-thumbnail (from YouTube/Vimeo), so the admin setting won't show up there unless that view display is changed to use `field_thumbnail` instead of `thumbnail`.
- **Image** — images are their own content, no fallback display path.

The settings fields are there if either type ever needs to be wired up to a display fallback in the future.

michael-wynne-wsu:

Link:
- https://pr1660-zfzj54q4juepg23xtbfgsecu8iauqmzj.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a0fd43df15c56e310b4bc2c

christa-kavanagh-wsu (approved):
This appears to be working as expected.

joelle-birano-wsu:
Is the thumbnail supposed to show up on browse pages?

Screenshot 2026-05-22 at 4 09 46 PM

It looks great in DH. I'm not seeing it in person or place records.

Screenshot 2026-05-22 at 4 13 05 PM

michael-wynne-wsu:
> Is the thumbnail supposed to show up on browse pages?

As specced, no, but we seem to include it in grid view but not list. Personally I don't think having it adds anything for users, but I'm open to opinions either way.

> It looks great in DH. I'm not seeing it in person or place records.

Ah I bet they have some different config. I'll check dictionary > additional media field as well.

michael-wynne-wsu:
Okay, got that thumbnail showing on all "media assets" fields (added to person, place, and dictionary words).

Now, opinions on including it in all/some previews for DH/place/person? 
What  about dictionary word previews on pages other than `/dictionary`?

joelle-birano-wsu:
Download button is off to the side now for dictionary words - not sure if that's a problem for this PR. But the image displays nicely for all content now (when viewing the full item)

Screenshot 2026-05-27 at 9 40 51 AM

Grid view is inconsistent on /browse
Screenshot 2026-05-27 at 9 50 24 AM

I like a preview so people know there's an audio file in that content item. I don't mind if it's just the audio player since people can preview the audio without having to click into the item... The image is fine as well, just a little less useful.

michael-wynne-wsu:
> Download button is off to the side now for dictionary words - not sure if that's a problem for this PR. But the image displays nicely for all content now (when viewing the full item)

I actually don't want the image on the words page, will adjust both!

> Grid view is inconsistent on /browse

Those previews are all super inconsistent, I'm going to tackle them all in a future issue.

michael-wynne-wsu:
There is a little recording weirdness, but I'm going to stop fighting it for now, because it may be rendered moot by other earlier PRs. So this is a note to confirm recording display before merge.

michael-wynne-wsu:
## PR Review: 320-audio-thumbnails

**Blast Radius:** Medium — touches dictionary word display, media entity classes, theme preprocess, and update hooks. No shared API contracts or DB schema changes.
**Security:** Clean — no injection, no secrets, no XSS vectors.
**Tests:** None added (no PHP unit tests exist for this layer in the repo — consistent with existing practice).
**Breaking Changes:** None — all config changes are accompanied by update hooks.

---

### MUST FIX (Blocking)

**1. Upload location changed from private to public in `ThumbnailSettingsForm.php`**

`#upload_location` was changed from `"private://"` to `"public://thumbnail-settings"`. If admin-uploaded default thumbnails contain culturally sensitive images, moving them to the public filesystem makes them world-accessible without protocol checks. Confirm this is intentional and that these thumbnails are never expected to be restricted content.

**2. `excludedMediaBundles` emptied — audio thumbnail settings now exposed to all bundles**

`protected $excludedMediaBundles = ['audio', 'image', 'remote_video']` → `[]`. The removed comment says "we do not support thumbnail generation for audio, image, and remote video." Emptying this means the thumbnail settings form now applies to all media bundles including audio, which aligns with the feature goal — but verify that the audio, image, and remote_video paths through `ThumbnailSettingsForm` are actually exercised and tested before this ships, since the original exclusion implied they weren't ready.

---

### SHOULD FIX (Non-blocking)

**3. `has_additional_word_entries` only set for `dictionary_word` nodes**

In `mukurtu_v4_preprocess_node()`, `$variables['has_additional_word_entries']` is only set when `bundle() === 'dictionary_word'`. The template checks `has_additional_word_entries ? ... : 0` — if the variable is absent (other node types), Twig will treat it as falsy and `num_entries` will be `0`, which is the correct safe default. No bug, but worth noting: if the template is ever reused for another bundle the default behaviour is correct.

**4. Update hook 40029 only guards `isNew()` — not idempotent on repeat runs**

If `drush updb` is run twice (e.g., after a failed deployment), hook 40029 will call `$config->set()` and `->save()` again. Not harmful, but hooks 40024–40027 follow the same pattern consistently, so this is in line with project convention. Low risk.

---

### SUGGESTIONS

**5. `_mukurtu_v4_audio_fallback_thumbnail_url()` is a private helper — consider moving to a service**

The leading-underscore convention works, but as audio thumbnail logic grows (it already has a 3-step fallback chain), a small static helper class or a dedicated service would be easier to unit-test. Not a blocker for this PR.

**6. Tab loop uses `loop.index` for both button and pane IDs across two separate `{% for %}` loops**

Both loops iterate `node.field_additional_word_entries` with the same `{% if entry.entity %}` guard. Because Twig `loop.index` resets per loop, the IDs will match only if both loops skip the exact same items — which they will, since the condition is identical. This is correct but fragile: if the two guards ever diverge, `aria-controls` / `id` pairs will silently break. Consider extracting the valid entries list in the preprocess hook and passing it as a template variable alongside `has_additional_word_entries`.

---

### LOOKS GOOD

- All new update hooks guard with `isNew()` before writing — safe on fresh installs.
- `media--audio--browse.html.twig` correctly strips contributor and download from the browse card — minimal and intentional.
- `aria-label="Download audio: {{ name }}"` added to the dictionary word audio download button — matches the established pattern in `media-assets` template.
- The PHP-side empty-paragraph detection (`isEmpty()` per field) is the right tool for this; Twig field value checks are unreliable for this use case.
- SCSS scoping: `margin-block-start: 0` override is correctly scoped inside `.field--name-field-recording .media--audio`, not applied globally.
- Browse display config no longer depends on `audio_for_dictionary_teaser` — dependency array is clean.

michael-wynne-wsu:
Committed as `15564f9e`. All review findings addressed:

| Finding | Resolution |
|---|---|
| Public upload location | Intentional — no change |
| `excludedMediaBundles` emptied | Intentional — no change |
| `has_additional_word_entries` scoping | Already safe default (falsy when absent) — no change |
| Hook 40029 idempotency | Matches project convention — no change |
| `_mukurtu_v4_audio_fallback_thumbnail_url` | Moved to `MukurtuV4AudioHelper::fallbackThumbnailUrl()` in `includes/AudioHelper.php` |
| Tab loop `loop.index` fragility | Preprocess now exports `additional_word_entry_indices`; both loops iterate the same array, eliminating the divergence risk |

nick-deer (approved):

PR #1667, opened 2026-05-24
748 media library modal ux

### LINKED ISSUE #748: Add Media popup padding
The padding around the Add Media popup is large enough that it makes it challenging to see what media assets you have without knowing you need to to scroll down. I am not sure if that can be reduced, but reducing the size of the padding would alleviate that issue.

### LINKED ISSUE #770: Media Asset Upload Window Display Issues
This ended up being a few things.

1) After uploading and before inserting into a piece of content, the upload window looks like this with the "Remove" box floating by itself. It works ok but it wasn't immediately apparent to me that the upload was successful and the "Remove" box placement is a little weird. 

Image

2) When inserting an image, the "Alternative text" field is displayed above the "Name" and I wonder if that should be flipped.

Image

3) When inserting audio into a piece of content (after upload), thumbnails for selecting media look like this:

Image

The insert window looks the same whether you're uploading and then inserting, or inserting already uploaded audio and even if thumbnails are attached. 

4) Externally embedded video thumbnails in the insert window also look disproportionate - tall rather than wide. Uploaded video doesn't have that same issue. 

Image

In all cases, once the DH Item is saved, the media looks fine when viewing, except for uploaded video which is already a recorded issue.

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/748
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/770

- [ ] Moves and relabels the "remove" button present while creating media. It is now "cancel" and place beside the "Save" button at the bottom of the modal.
- [ ] Pins media type left sidebar so it doesn't scroll with the rest of the modal.
- [ ] Removes grid display (doesn't seem like a value add and removing the toggle saves space).
- [ ] Adjusts various margins and paddings to compact space.
- [ ] Improves specific button alignments.
- [ ] Adds an "all media" view.
- [ ] Enables full-row click to select media.
- [ ] Adds edit media action in modal.

Pre-merge checklist:


christa-kavanagh-wsu (commented):
Overall it looks much better and I like the cancel option. I love the all media option, that's really nice. Agree grid view is overrated.

It said it won't let me save media assets in the modal for some reason? All required fields were filled in and they all saved, I just got a weird error and had to X out of the modal.

Image

This is the error I got: 

Image

But the media assets ended up in the library. 

Image

michael-wynne-wsu:
@christa-kavanagh-wsu Error seems to be resolved, can save media in modal as expected.

christa-kavanagh-wsu (approved):
Looks good!

joelle-birano-wsu (approved):
Looks good!

michael-wynne-wsu:
## PR Review: 748 Media Library Modal UX Improvements

**Branch:** `748-media-library-ux` vs `main`
**Files Changed:** 7 (1 new JS, 1 new PHP class, 1 new service, CSS, .module, .libraries.yml, ExternalEmbedForm)
**Lines Added:** ~580
**Blast Radius:** HIGH — service override replaces `media_library.ui_builder` site-wide; affects every media library modal on every entity type

---

### MUST FIX (Blocking)

**1. N+1 query in `mukurtu_media_field_widget_single_element_form_alter`**
`mukurtu_media.module:319-332`

`$media_storage->load($media_id)` is called once per `$delta` inside a `foreach`. On a digital heritage node with 10 media items this fires 10 separate DB queries. Use `loadMultiple()` before the loop:

```php
// Collect all IDs first.
$media_ids = array_filter(array_map(
  fn($delta) => $element['selection'][$delta]['target_id']['#value'] ?? NULL,
  Element::children($element['selection'])
));
$media_items = $media_storage->loadMultiple($media_ids);

foreach (Element::children($element['selection']) as $delta) {
  $media_id = $element['selection'][$delta]['target_id']['#value'] ?? NULL;
  $media = $media_items[$media_id] ?? NULL;
  // ...rest unchanged
}
```

---

**2. Checkbox column `display:none` removes checkboxes from accessibility tree**
`mukurtu-gin-custom.css:361-364`

`display: none !important` removes elements from the accessibility tree entirely — screen readers cannot reach the checkboxes. Use the visually-hidden technique instead (already defined in the Gin theme):

```css
.media-library-widget-modal .views-table th.views-field-media-library-select-form,
.media-library-widget-modal .views-table td.views-field-media-library-select-form {
    position: absolute;
    width: 1px;
    height: 1px;
    overflow: hidden;
    clip: rect(0, 0, 0, 0);
    white-space: nowrap;
}
```

---

**3. Clickable table rows have no keyboard support**
`media-library-cancel.js:84-94`

The `<tr>` click handler has no corresponding `keydown` handler, and rows have no `tabindex`. Keyboard-only users cannot select media. Add to `syncEditLinks()` after appending cells:

```js
$table.find('tbody tr').each(function () {
  var $row = $(this);
  if (!$row.attr('tabindex')) {
    $row.attr('tabindex', '0');
  }
});

// In the delegated event block:
$(document).on('keydown', '.media-library-widget-modal .views-table tbody tr', function (e) {
  if (e.key === ' ' || e.key === 'Enter') {
    e.preventDefault();
    var $cb = $(this).find('input[type="checkbox"]');
    $cb.prop('checked', !$cb.prop('checked')).trigger('change');
  }
});
```

Also add a focus-visible ring to the CSS:
```css
.media-library-widget-modal .views-table tbody tr:focus-visible {
    outline: 3px solid var(--gin-color-primary, #0036b1);
    outline-offset: -3px;
}
```

---

**4. `ExternalEmbedForm` description severs `aria-describedby` association**
`ExternalEmbedForm.php:62-69`

Drupal's Form API automatically wires `aria-describedby` between a `#description` and its parent form element. Moving the description to a standalone `html_tag` outside `$form['container']` breaks that relationship — the textarea has no `aria-describedby` and screen readers won't announce the help text. Keep `#description` on the textarea:

```php
$form['container']['external_embed_code'] = [
  '#type' => 'textarea',
  '#title' => $this->t('Add External Embed'),
  '#rows' => 3,
  '#required' => TRUE,
  '#description' => $this->t('Embed code from an external website...'),
  '#description_display' => 'after',
];
// Remove the standalone $form['description'] html_tag entirely.
```

---

### SHOULD FIX (Non-blocking)

**5. `$(document).ajaxComplete` fires on every AJAX request site-wide**
`media-library-cancel.js:62-65`

`syncCancelProxy()` and `syncEditLinks()` run after every AJAX call on the page, not just media library ones. Add an early-return guard:

```js
$(document).ajaxComplete(function () {
  if (!$('.media-library-widget-modal').length && !$('.media-library-add-form').length) {
    return;
  }
  syncCancelProxy();
  syncEditLinks();
});
```

---

**6. `role="button"` on anchor conflicts with link semantics**
`MukurtuMediaLibraryUiBuilder.php:131`

An `<a href="...">` with `role="button"` is announced as "button" but only activates on Enter (links), not Space (buttons). Screen readers are misled about the interaction model. Remove `role="button"` — the link semantics are correct since it navigates to a URL via AJAX:

```php
'attributes' => [
  'data-title' => $label,
  // role="button" removed
],
```

---

**7. Cancel proxy triggers `mousedown` + `click` on a Drupal AJAX button**
`media-library-cancel.js:29-31`

Drupal's AJAX system attaches to `mousedown` on submit-type inputs, not plain `click`. Triggering both programmatically could double-fire or miss the AJAX handler in some browsers. Use `$source[0].click()` (native DOM click dispatches both) or directly trigger Drupal's AJAX handler:

```js
$proxy.on('click', function () {
  $source[0].click();
});
```

---

**8. Edit link `aria-label` missing for new-tab behavior**
`media-library-cancel.js:54`

```js
'<a href="/admin/media/' + mediaId + '/edit" target="_blank" rel="noopener">' +
Drupal.t('Edit') + '</a>'
```

Opens in a new tab with no announcement. Add visually-hidden text:

```js
Drupal.t('Edit') + '<span class="visually-hidden"> (' + Drupal.t('opens in new tab') + ')</span>'
```

---

### Suggestions

**9. Hard-coded `/admin/media/` path in JS**
`media-library-cancel.js:54`

If a site customizes the admin path prefix, these links silently 404. The server-side `hook_field_widget_single_element_form_alter` already generates correct edit URLs — consider passing the URL through a `data-edit-url` attribute on the checkbox or row, and reading it in JS instead of constructing the path client-side.

**10. No tests**

`MukurtuMediaLibraryUiBuilder` has significant branching logic (`isAllMediaMode`, tab injection, filter manipulation, display override). There are no test files in the diff. At minimum, a kernel test verifying that `buildMediaTypeMenu` prepends the "All media" tab and that `stateWithAllMedia`/`stateWithoutAllMedia` round-trip correctly would prevent regressions from future core `MediaLibraryUiBuilder` changes.

**11. Note core version dependency for protected method overrides**

`MukurtuMediaLibraryUiBuilder` overrides four protected methods (`buildUi`, `buildMediaTypeMenu`, `buildLibraryContent`, `buildMediaLibraryView`). These are not part of a stable API. A comment at the top of the class noting the Drupal core version this was validated against would help future maintainers catch breakage at upgrade time.

---

### Looks Good

- Service override pattern in `mukurtu_media.services.yml` is correct Drupal idiom
- HMAC-signed state via `opener_parameters` for `mukurtu_show_all` flag is the right approach - survives AJAX round-trips
- `$media->access('update')` check before rendering Edit button — no privilege escalation
- `rel="noopener"` on all `target="_blank"` links
- CSS scoped to `.media-library-widget-modal` — no global style pollution
- Allowed-type enforcement via hidden `bundle` filter is solid; avoids leaking out-of-scope media types on the "All media" tab
- `CloseModalDialogCommand` on clean media edit save is the correct AJAX response pattern

---

### Summary

| Priority | Count |
|---|---|
| MUST FIX | 4 |
| SHOULD FIX | 4 |
| Suggestions | 3 |

The core UX approach is well-conceived, but items 1-4 need to be addressed before merge: the N+1 query will hurt performance on content-heavy nodes, the two accessibility issues are WCAG 2.1 AA failures, and the `aria-describedby` breakage in `ExternalEmbedForm` is a regression from the current form behavior.

michael-wynne-wsu:
| Issue | Fix |
|---|---|
| N+1 query on media widget | Collect all IDs first, one `loadMultiple()` call |
| Checkbox column `display:none` | Replaced with visually-hidden CSS - stays in accessibility tree |
| No keyboard support on table rows | Added `tabindex="0"` per row + `keydown` handler for Space/Enter |
| No focus indicator on rows | Added `:focus-visible` outline rule |
| `ajaxComplete` fires site-wide | Early-return guard when no media library UI is present |
| Cancel proxy double-triggers AJAX | Changed to `$source[0].click()` (native DOM click) |
| Cancel button no aria-label | Added `aria-label: "Cancel media upload"` |
| Edit link no new-tab warning | Added `<span class="visually-hidden">(opens in new tab)</span>` |
| ExternalEmbed description severs aria-describedby | Added explicit `id` to description div + `aria-describedby` on textarea; description stays full-width below the row |
| `role="button"` on anchor | Removed |

michael-wynne-wsu:

Link:
- https://pr1667-zxcdqntjxrhdhqwzulimx6hzdv2fq5rh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a133178f15c56e310dff3c6

PR #1668, opened 2026-05-24
Add optional restricted and no-media placeholder images (#813)

### LINKED ISSUE #813: "No media available" messages
V3 included a "No media is available on this item" in the preview thumbnails for DH items when there was either no media asset provided, or the media asset(s) were under user-excluding protocols. This wasn't used for other content types or in other views that I can recall.

Do we want to maintain this behaviour in V4?

My immediate preference is no. Even with a better *looking* graphic, I find it can be confusing to users and may lead to people seeking access to content that doesn't exist/they don't have permission for.

A value-add would be easy end-user configuration of a placeholder graphic or message to be configured IF they want to.

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/813

- Adds a new Media Settings form at `/admin/config/mukurtu/media-settings` in the Media section of the dashboard (accessible to admins and Mukurtu Managers) with two optional file upload settings:
  - Restricted media placeholder: shown when a user lacks protocol access to media
  - No media placeholder: shown when content has no media assets at all
- This is optional. If the placeholders are not added, then there is simply no indicator of media displayed, same as the current behaviour.

Uses hook_entity_view_alter to inject placeholder render arrays after Drupal's formatter pipeline, and modifies RepresentativeMediaItemList to defer to the hook when placeholders are configured rather than falling back to the default image. Also fixes the list-view browse template to render placeholders injected via the computed field_representative_media.

Pre-merge checklist:


christa-kavanagh-wsu (commented):
Overall this works well. The only thing I noticed is that Word list is showing up with the no media asset thumbnail in the Browse and Dictionary browse pages but with the correct restricted media thumbnail on the word list itself. 

Image

Image

michael-wynne-wsu:
Okay I've got that fixed I think. Getting the three possible images set and showing as expected for me with word lists.

joelle-birano-wsu (approved):
This seems to be working well.

christa-kavanagh-wsu (approved):
This looks like it is working as intended.

michael-wynne-wsu:
## PR Review: Restricted Media Placeholder Images (#813)

**Branch:** `feature/813-restricted-media-placeholder`
**Files changed:** 10 (branch-only diff, excluding merge-in changes)
**Blast Radius:** MEDIUM — touches the computed `field_representative_media` field used site-wide on all content browse cards, plus a new `hook_entity_view_alter` that runs on every node render.

---

### MUST FIX (Blocking)

**1. Missing `#alt` on placeholder images — WCAG 1.1.1**

`modules/mukurtu_media/mukurtu_media.module` (Scenario 1 ~line 390, Scenario 2 ~line 415): both `#theme => 'image'` render arrays have no `#alt` key. Drupal renders `` with no alt, and screen readers will read the raw filename. These images are functional — they communicate restricted access or absent media — so they require descriptive alt text.

```php
// Scenario 1 (no-media or restricted):
'#alt' => $placeholder_file_to_use === $restricted_file
  ? (string) t('Media not available due to protocol restrictions')
  : (string) t('No media available'),

// Scenario 2 (restricted items in field):
'#alt' => (string) t('Media not available due to protocol restrictions'),
```

**2. Download button renders `href=""` when `media_url` is empty**

`themes/mukurtu_v4/templates/media/media--audio--media-assets.html.twig:57` uses `{% if media_download_enabled %}` without also checking `media_url`. The other two audio templates (`--default`, `--audio-for-dictionary-teaser`) correctly use `{% if media_download_enabled and media_url %}`. An empty `href` on a download anchor is an invalid link and a keyboard/screen reader trap.

```twig
{# Change line 57 from: #}
{% if media_download_enabled %}
{# To: #}
{% if media_download_enabled and media_url %}
```

---

### SHOULD FIX (Non-blocking)

**3. N+1 query pattern in `mukurtu_media_entity_view_alter`**

`modules/mukurtu_media/mukurtu_media.module`: the hook iterates every media reference field and calls `$media_storage->load($item->target_id)` individually inside nested loops. For a DH item with `field_media_assets` containing 5 items and `field_thumbnail`, that's 6+ separate queries per page render, multiplied by every node on a browse listing. The fix is to collect all `target_id` values first and call `$media_storage->loadMultiple($ids)` once per field.

**4. No file cleanup when placeholder is removed**

`modules/mukurtu_media/src/Form/MediaSettingsForm.php`: when a user removes a configured placeholder via the form, the old `File` entity has already been set permanent (`$file->setPermanent()`) and is never cleaned up. Drupal's file usage system won't garbage-collect it. Consider tracking file usage with `\Drupal::service('file.usage')->add/delete()` on save, or at minimum documenting that admins should manually delete orphaned files.

**5. New config keys missing from install YAML**

`modules/mukurtu_media/config/install/mukurtu_media.settings.yml` only contains `mukurtu_media_download_enabled: true` — the new `restricted_media_placeholder` and `no_media_placeholder` keys are absent. The code handles NULL safely via `!empty()`, but providing explicit defaults (`restricted_media_placeholder: []` / `no_media_placeholder: []`) is cleaner and consistent with the schema.

---

### No Tests

There are no tests for `mukurtu_media_entity_view_alter`, `_mukurtu_media_build_placeholder_field`, or the updated `RepresentativeMediaItemList` logic. Given the blast radius (this runs on every node render), at minimum a kernel test covering the three cases — accessible media, restricted media with placeholder configured, no media with placeholder configured — would prevent regressions.

---

### Looks Good

- **`RepresentativeMediaItemList.php` logic is correct.** The `return` on line 41 exits as soon as accessible media is found, so the early-return placeholder checks below are only reached when no accessible media was found. The three code paths are mutually exclusive.
- **Route properly guarded.** `mukurtu_media.media_settings_form` requires `administer mukurtu_media configuration`.
- **Update hook `mukurtu_media_update_40003`** correctly grants the new permission to existing `mukurtu_manager` roles.
- **`MediaSettingsForm` file upload** is validated to image types only (`png gif jpg jpeg webp`) and stored in `public://mukurtu/placeholder/`.
- **`node--digital-heritage--browse.html.twig`** change is minimal and correct — the `or content.field_representative_media` condition allows the placeholder to render without affecting any other template logic.
- **Permission gated with `restrict access: true`** in permissions.yml.
- **Config schema** covers both new settings keys correctly.

---

### Summary

| # | Issue | Severity |
|---|-------|----------|
| 1 | Placeholder images missing `#alt` (WCAG 1.1.1) | Must fix |
| 2 | `media--audio--media-assets.html.twig` download button missing `and media_url` guard | Must fix |
| 3 | N+1 media loads in `hook_entity_view_alter` | Should fix |
| 4 | No file cleanup on placeholder removal | Should fix |
| 5 | New config keys missing from install YAML | Should fix |
| 6 | No tests for new placeholder logic | Should fix |

michael-wynne-wsu:
| # | Fix | File(s) |
|---|-----|---------|
| 1 | `#alt` on both placeholder render arrays | `modules/mukurtu_media/mukurtu_media.module` |
| 2 | Added `and media_url` to download button condition | `themes/mukurtu_v4/templates/media/media--audio--media-assets.html.twig` |
| 3 | Batch `loadMultiple` in both Scenario 1 and 2 — eliminates 2× N+1 loops | `modules/mukurtu_media/mukurtu_media.module` |
| 4 | `submitForm` now diffs old vs new fids: marks added files permanent, removed files temporary for GC | `modules/mukurtu_media/src/Form/MediaSettingsForm.php` |
| 5 | Added `restricted_media_placeholder: []` and `no_media_placeholder: []` defaults | `modules/mukurtu_media/config/install/mukurtu_media.settings.yml` |
| 6 | Tests deferred — tracked separately | |

nick-deer (commented):
Looks good, but we are also giving the option to set a default image at /admin/config/mukurtu/settings (under citation formats and default image). I tried setting a default image here and at media settings and the media settings image was chosen, so it's not breaking anything at least. Is there any reason why we should allow setting a default image at /admin/config/mukurtu/settings?

michael-wynne-wsu:

Link:
- https://pr1668-rgoguj07fstzewgcvjj6rxrdfjylmxaj.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a13441331fbd5936744b0d7

michael-wynne-wsu:
> Looks good, but we are also giving the option to set a default image at /admin/config/mukurtu/settings (under citation formats and default image). I tried setting a default image here and at media settings and the media settings image was chosen, so it's not breaking anything at least. Is there any reason why we should allow setting a default image at /admin/config/mukurtu/settings?

Good catch. I thought I had addressed that but looking back I think I just forgot. I will see about cleaning that out.

michael-wynne-wsu:
@nick-deer I've removed that redundant default image and updated the citation formats link.

nick-deer (approved):

PR #1669, opened 2026-05-24
Improve media Found In page and fix sidebar null crash (#1057)

### LINKED ISSUE #1057: Media "found in" tool not working
From a media asset (viewed alone or in media library), there is a "found in" link to let you see which content references that media. It did work previously, unsure when there was a reversion.

Image

A mukurtu admin receives an access error....

Image

And UID 1 receives a PHP error...

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1057

- [ ] When in the media library (not the modal), or from a media asset page, select "found in" for a list of content and entities where the media is used.

Screenshot 2026-05-24 at 1 06 32 PM

- Fix null dereference in community/protocol sidebar preprocess hooks when route lacks the expected entity parameter
- Render Found In page with admin theme
- Rewrite MediaFoundInController to use a sortable table (Title, Type, Content Type) instead of collapsible browse-card sections
- Fix phantom community/protocol results caused by access handler corrupting OR queries; query with accessCheck(FALSE) and filter loaded entities by view access instead
- Move content (node) results to top of list

Pre-merge checklist:


christa-kavanagh-wsu (approved):
This appears to work well!

joelle-birano-wsu (approved):
Looks good to me!

michael-wynne-wsu:
## PR Review: Improve media Found In page and fix sidebar null crash (#1669)

**Blast Radius:** LOW - 3 files, admin-only controller + theme preprocess hooks. No shared library, API contract, or schema changes.
**Security:** 1 note (see below) - `accessCheck(FALSE)` is intentional and safe.
**Tests:** No new tests added; no test coverage delta (controller logic, not unit-testable without Drupal kernel).
**CI:** Failing, but identical failures exist on `main` (29 errors / 6 failures - pre-existing `og` module image plugin issue in the test runner).
**Breaking Changes:** None detected.

---

### LOOKS GOOD

**Correctness fixes:**
- `getType()` case fix - correct; aligns with the `FieldStorageDefinitionInterface` signature.
- `accessCheck(FALSE)` + post-load `$e->access('view', $this->currentUser())` filter - this is the right pattern to work around the known Drupal bug where access handlers corrupt OR queries. The comment in the PR description explains it well.
- Null guards in `mukurtu_v4_preprocess_field_group_html_element__group_community_sidebar` and the protocol equivalent - correct and necessary; these hooks fire on any route that includes a field group, not just community/protocol entity routes.
- `_admin_route: TRUE` on the Found In route - appropriate; the page is an admin utility and should render with the admin theme.
- Nodes moved to top of list (processing `node` before `community`/`protocol`) - the natural expected order.
- `array_unique(array_merge(...))` deduplication between reference-field and embedded-text results - correct.

---

### SHOULD FIX (Non-blocking)

**1. Missing table `#caption` - WCAG 1.3.1**
The `#type => 'table'` has no `#caption`. Screen reader users navigating by table will encounter a table with no announced purpose.

```php
// In MediaFoundInController::content(), add to the return array:
'#caption' => $this->t('Content using %label', ['%label' => $media->label()]),
```

**2. `$entity->label()` can return `null` - WCAG 4.1.2**
For `community` and `protocol` entity types, `label()` returning `null` would produce an `<a>` tag with no accessible name. Low probability but cheap to guard:

```php
'#title' => $entity->label() ?? (string) $entity->id(),
```

**3. No pagination on the table**
The old code had `->pager(10)`. The new table has no limit. For a media asset used in hundreds of nodes (e.g., a site logo), this page will load every entity into memory and render a very long table. Consider a note or a follow-up issue if an unbounded result set is acceptable for this use case.

---

### SUGGESTIONS

**4. `\Drupal::service()` static calls in a controller**
Lines 22 and 59 use `\Drupal::service('entity_field.manager')` and line 153 uses `\Drupal::service('entity_type.bundle.info')`. These were pre-existing except for line 153 which is new. These services should be injected via the constructor for testability. Not a blocker for this PR since the class is otherwise already using static calls, but worth noting since the class is being touched.

**5. `toUrl()` exception risk for community/protocol entities**
`$entity->toUrl()` on line 166 will throw a `\Drupal\Core\Entity\Exception\UndefinedLinkTemplateException` if the `community` or `protocol` entity type doesn't have a `canonical` link template defined. Worth verifying both entity types have this, or wrapping in a try/catch as a fallback.

---

### Checklist

| Item | Status |
|------|--------|
| PR title accurately describes the change | Pass |
| PR description explains WHY | Pass |
| Linked issue exists (#1057) | Pass |
| No scope creep | Pass - focused |
| No hardcoded secrets | Pass |
| No SQL injection / raw query interpolation | Pass |
| `accessCheck(FALSE)` is intentional and guarded by post-load filter | Pass |
| No XSS vectors | Pass |
| No N+1 queries | Pass - `loadMultiple()` called once per entity type outside the row loop |
| Table semantic structure (WCAG 1.3.1) | Pass (with `#caption` suggestion) |
| Link accessible names (WCAG 4.1.2) | Minor - label null risk |
| Keyboard access | Pass |
| CI failures are pre-existing | Confirmed - identical to `main` |
| Update hook check | N/A - no schema/config changes in these 3 files |
| SCSS recompile needed | No - no theme CSS changes |

**Overall verdict:** Ready to merge after addressing the `#caption` (accessibility) and label null-guard (defensive hardening). The pagination concern is worth a follow-up issue rather than a blocker. The CI failure is a pre-existing environment issue unrelated to this PR.

michael-wynne-wsu:
Both fixes applied:

- modules/mukurtu_core/src/Controller/MediaFoundInController.php line 103 - `label() ?? (string) $entity->id()` null guard
- modules/mukurtu_core/src/Controller/MediaFoundInController.php line 114 - `#caption` added with the media item's label

The two suggestions (service injection and `toUrl()` try/catch) I left alone - service injection would be a larger refactor outside this PR's scope, and `toUrl()` only throws if the entity type has no canonical link template, which both `community` and `protocol` almost certainly do. 

-----

Both `community` and `protocol` have `"canonical"` link templates defined, so `toUrl()` is safe. No try/catch needed.

The two "should fix" changes are sufficient and the suggestions can stay as-is. Ready to commit whenever you are.

nick-deer (approved):

christa-kavanagh-wsu:
Post merge test - looks good!

PR #1677, opened 2026-05-28
Update mukurtu_footer.install

michael-wynne-wsu:
Patches an incomplete update hook.

Pre-merge checklist:

### DIFF
--- a/modules/mukurtu_footer/mukurtu_footer.install
+++ b/modules/mukurtu_footer/mukurtu_footer.install
@@ -38,6 +38,12 @@ function mukurtu_footer_install() {
  * entity and migrates existing data into its fields.
  */
 function mukurtu_footer_update_40001(): void {
+  // Ensure all config/install configs (block content type, paragraph types,
+  // fields) exist before trying to create entities that use them. On existing
+  // sites, config/install YAMLs are only applied during a fresh module install,
+  // not during updates, so this call fills the gap.
+  \Drupal::service('config.installer')->installDefaultConfig('module', 'mukurtu_footer');
+
   // If a mukurtu_footer block_content entity already exists, nothing to do.
   $existing = \Drupal::entityTypeManager()
     ->getStorage('block_content')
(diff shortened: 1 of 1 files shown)


michael-wynne-wsu:
## PR Review: Fix field_footer_logos unknown on existing site updates

**Blast Radius:** LOW — single `.install` file, update-path only, no runtime code changed  
**Security:** No findings  
**Tests:** N/A (install hooks; manual verification via `drush updb`)  
**Breaking Changes:** None — `installDefaultConfig()` is additive only, never overwrites

---

### --- MUST FIX (Blocking) ---

None.

---

### --- SHOULD FIX (Non-blocking) ---

**1. Docblock on `mukurtu_footer_update_40001()` no longer fully describes what the function does**

The existing docblock only describes the data migration purpose. Now that the function also installs config, the description is incomplete and will confuse anyone looking at this hook later.

```php
/**
 * Install default configuration and migrate footer data to block_content.
 *
 * Installs block content type, paragraph types, and field configurations from
 * config/install (which is only applied on fresh installs, not updates). Then
 * migrates existing footer data from block plugin configuration into the new
 * mukurtu_footer block_content entity.
 */
```

---

### --- SUGGESTIONS ---

**2. Consider whether `installDefaultConfig()` in 40001 is still needed long-term**

Since 40002 exists specifically to backfill sites that already ran 40001, the call added to 40001 only benefits sites that haven't run 40001 yet. That's valid, but it means 40001 and 40002 do identical first steps. This is intentional and correct — just worth a comment noting the redundancy is deliberate so a future cleanup doesn't remove only one of them.

---

### --- LOOKS GOOD ---

- **Correct root cause fix.** `config/install` YAMLs are only applied on fresh installs; using `installDefaultConfig()` in an update hook is the standard Drupal pattern for this gap.
- **Idempotent.** `installDefaultConfig()` skips configs that already exist, so running it twice (40001 + 40002 on a broken site, or 40002 on a site with a working install) is safe.
- **Follows established Mukurtu pattern.** Same approach as `mukurtu_browse_update_40001()` and `mukurtu_browse_update_40003()`.
- **Correct hook numbering.** `40002` follows sequentially and doesn't conflict with existing hooks.
- **Docblock on 40002 is accurate** — clearly explains why the hook exists and what it does.
- **No scope creep.** The diff is exactly two logical changes: one line added to 40001, one new function. Nothing unrelated was touched.

---

**Summary:** The fix is correct and safe. The one non-blocking item is a docblock update on `mukurtu_footer_update_40001()` to reflect that it now does two things. Everything else is solid.

PR #1689, opened 2026-05-30
1398 site wide comment settings

### LINKED ISSUE #384: Site-wide comment settings

michael-wynne-wsu:
Closes: https://github.com/MukurtuCMS/Mukurtu-CMS/issues/384

Adds "require approval for comments", "allow visitors to view comments", and "allow visitors to leave comments" site-wide settings at `/admin/config/mukurtu/commenting/settings`

Pre-merge checklist:


michael-wynne-wsu:
Screenshot 2026-05-30 at 12 41 05 PM

christa-kavanagh-wsu (commented):
I tried logging out to view a DH item with neither of these selected 

Image

and the site broke. I got a PHP/Twig error:

Image

It works fine with one or both selected, but if you deselect both it gives you an error.

michael-wynne-wsu:
Potential fix being pushed. have not tested it myself yet so let me know.

christa-kavanagh-wsu (approved):
That seems to have fixed it!

michael-wynne-wsu:
## PR Review: `fix-pathauto-source-module-php8-attribute` (1398 site-wide comment settings)

**Blast Radius:** MEDIUM - changes comment approval flow and admin form; affects all comment submissions site-wide
**Security:** No issues found
**Tests:** No test files changed (no automated tests exist for this area in this codebase)
**Breaking Changes:** New config key added to existing config object - requires update hook

---

### MUST FIX (Blocking)

**1. Missing update hook for `site_comments_require_approval`**

A new key is added to the existing `mukurtu_protocol.comment_settings` config object, but no update hook sets an initial value for existing sites. Per `docs/update-hooks.md`, config additions to existing objects require an update hook.

Existing installs will get `null` (which evaluates to `false` = approval disabled), which changes behavior silently. The next available hook number is `mukurtu_protocol_update_40035`. Should set `site_comments_require_approval` to `FALSE` explicitly.

`modules/mukurtu_protocol/mukurtu_protocol.install`

---

### SHOULD FIX (Non-blocking)

**2. No validation preventing `post comments` without `access comments`**

A submitter can check "Allow visitors to leave comments" without checking "Allow visitors to view comments." This produces a logically inconsistent role: anonymous users can post but can't read. Consider adding a `#states` dependency so `anonymous_can_post_comments` is only enabled when `anonymous_can_access_comments` is checked, or add a `validateForm()` that catches this combination and sets an error.

`modules/mukurtu_protocol/src/Form/CommentSettingsForm.php:90`

**3. `site_comments_require_approval` default value is ambiguous**

The config key has no install-time default (`mukurtu_protocol.comment_settings.yml` doesn't exist), so a fresh install starts with `null`. The ternary `$commentsRequireApproval ? 1 : 0` silently defaults to `0` (disabled). This is probably the right default, but making it explicit in either an install config file or at least a `?? FALSE` in the ternary would be clearer.

---

### LOOKS GOOD

- **Twig null guards** - Clean, correct fix for the `field_media_assets` TypeError; pattern is consistent across all four templates.
- **Role permission management** - Correctly uses `grantPermission()`/`revokePermission()` + `$anonymous->save()` rather than writing to config directly. Drupal's role entity save triggers permission cache invalidation automatically.
- **Cache invalidation** - `Cache::invalidateTags(['node_view'])` is the right tag to bust after comment setting changes.
- **Approval logic** - `(!$this->anyProtocolsRequireCommentApproval($entity) && !$this->siteRequiresCommentApproval())` is semantically correct: a comment auto-publishes only when neither the protocols nor the site requires approval. The site setting stacks correctly on top of protocol settings.
- **DI pattern** - The `create()` factory override correctly chains `parent::create()` and injects `entity_type.manager` without breaking the config form base class injection.
- **No secrets, SQL injection, or XSS vectors found.**

---

The main blocker before merge is the **update hook**. The inconsistent anonymous permissions state is worth addressing but won't break existing behavior.

michael-wynne-wsu:
All three fixes applied:

1. **`?? FALSE` default** - `modules/mukurtu_protocol/src/Form/CommentSettingsForm.php:65` - explicit fallback instead of implicit null-to-false coercion.
2. **`#states` dependency** - `modules/mukurtu_protocol/src/Form/CommentSettingsForm.php:98` - "Allow visitors to leave comments" is only enabled when "Allow visitors to view comments" is checked.
3. **Update hook `mukurtu_protocol_update_40035`** - `modules/mukurtu_protocol/mukurtu_protocol.install` - sets `site_comments_require_approval` to `FALSE` for existing installs that don't have the key.

joelle-birano-wsu:
Disabling comment approval doesn't seem to do anything. The comment doesn't display and the success message says it's awaiting approval.

Screenshot 2026-06-03 at 11 55 05 AM


Screenshot 2026-06-03 at 11 46 07 AM

michael-wynne-wsu:
@joelle-birano-wsu protocol-level comment settings should take priority over site-wide settings. Can you see if that's why?

joelle-birano-wsu:
> @joelle-birano-wsu protocol-level comment settings should take priority over site-wide settings. Can you see if that's why?

Ah. That did it. Nevermind!

joelle-birano-wsu (approved):

nick-deer (approved):

PR #1690, opened 2026-05-30
Add per-protocol comment view and post access settings

### LINKED ISSUE #887: Community and Protocol level comment settings
Protocols

- [ ] Protocol level setting is at /admin/protocols/[protocol ID]/comment-settings
- [ ] Is the enabled/disabled binary of the "Require Approval for Comments" setting clear enough? Does it need to be more specific? Maybe specifying a role within the protocol?
- [ ] Protocol level setting should take precedence over site setting. Is that true? Ideal behaviour is if site-wide is disabled, protocols can't enable. If site-wide is enabled, protocols can disable.

Communities

- [ ] Do we want community-level controls?

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/887

Adds two new checkbox fields to /admin/protocols/{id}/comment-settings: who can view comments and who can leave comments. Options are visitors (anonymous), any site user, or protocol members (any role). Access checks enforce the most restrictive protocol wins across multi-protocol content.

- [ ] This will need to be tested after https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1689 is merged (or at least rebase it on that one) to make sure that the logic flow is accurate.

Pre-merge checklist:


michael-wynne-wsu:
Screenshot 2026-05-30 at 1 00 19 PM

joelle-birano-wsu:
I enabled view and commenting permissions for visitors. They can see comments but can't leave them.
Screenshot 2026-06-04 at 1 47 51 PM

If I disable viewing and commenting for visitors, the site breaks when a visitor tries to access a public item. This is the error I got:
Details-Mukurtu-CMS-06-04-2026_01_53_PM

I also run into problems if commenting is enabled but viewing is disabled, so if commenting is enabled, viewing should probably be auto-enabled.

michael-wynne-wsu:
@joelle-birano-wsu 

> If I disable viewing and commenting for visitors, the site breaks when a visitor tries to access a public item. This is the error I got:

that error message showed up in testing elsewhere and was fixed there, so it should be fine once this is merged in https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1689#pullrequestreview-4413101946

> I also run into problems if commenting is enabled but viewing is disabled, so if commenting is enabled, viewing should probably be auto-enabled.

I kind of implemented the inverse: if visitor view comment is disabled, visitor leave comment cannot be enabled. In progress.

> I enabled view and commenting permissions for visitors. They can see comments but can't leave them.

Okay that's because the site-wide default setting here doesn't allow visitors to comment - that is in one of the other PRs. But I manually changed that and it seems to work, so once they're all merged down I think it'll work.

Screenshot 2026-06-04 at 9 38 08 AM

Screenshot 2026-06-04 at 9 38 31 AM

PR #1702, opened 2026-06-02
Fix registration settings not triggering correct actions (#1695)

### LINKED ISSUE #1695: Updates to Registration Settings
Some settings at admin/config/people/accounts are not triggering the correct actions. This is a table of the settings that are triggering incorrect actions. I _think_ I've got them correct.

The descriptions of the settings also need updating - especially the settings that promise auto-generated passwords - currently those are not generating. Rather, reset links are being sent, which works just as well.

| Who can register accounts?  | Email verification  | User password entry  |  Current behavior |  Desired behavior |
|---|---|---|---|---|
|  Visitors |  Required | Password entry NOT allowed  |  Email sent: Welcome (new user created by administrator), user blocked when clicking on link | Welcome (no approval required) should send, user should not be blocked  |
| Visitors  |  Not required |  Password required | Email sent: An administrator created an account for you at Mukurtu, access denied  |  Does this require an email? Or just a success message that tells them they can login? User should not be blocked |
| Visitors | Not Required  | Password optional  |  The success message says that a password was generated. Welcome (new user created by administrator) email is also sent: An administrator created an account for you at Mukurtu, access denied |  Not sure what should happen here. If a password is generated, that should be communicated to the user either through success message or email. If the password is set by the user, they should get a success message that tells them they can login. The user should not be blocked. |
| Visitors  |  Not Required | Password entry NOT allowed  | Email sent: Welcome (new user created by administrator), user blocked when clicking on link  |  Welcome (no approval required) should send, user should not be blocked |
| Visitors, admin approval required  |  Required |  Password entry NOT allowed |  Email sent: Welcome (new user created by administrator), user blocked when clicking on link |  Welcome (awaiting appproval) should send, followed by Account activation once approved. |
| Visitors, admin approval required |  Not Required |  Password required |   Email sent: Welcome (new user created by administrator), user blocked when clicking on link|   Welcome (awaiting appproval) should send, followed by Account activation once approved.|
| Visitors, admin approval required  |  Not Required |  Password optional| Email sent: Welcome (new user created by administrator)  | Welcome (awaiting appproval) should send, followed by Account activation once approved.  |
|Visitors, admin approval required|Not Required|Password entry NOT allowed|Email sent: Welcome (new user created by administrator)|Welcome (awaiting appproval) should send, followed by Account activation once approved.|

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1695

## Summary

- [ ] Account settings form `/admin/config/people/accounts` has been completely reordered and regrouped to have a more logical flow and collapse low-use groups.
- [ ] Certain selections in one group/field will force-select or disable other selections further down the form so that invalid combinations can't be created. eg: only "visitors, no approval required" enables the email verification setting.
- [ ] Visitor registration with admin approval now properly records the pending status.
- [ ] Correct emails should be firing now (see below).
- [ ] Light theming adjustment to account request form and messages.

| Scenario | Email sent |
|---|---|
| Visitor registers, approval required | `register_pending_approval` (always) |
| Visitor registers, open registration, verify on | `register_no_approval_required` (has OTL) |
| Visitor registers, open registration, verify off | none (auto-logged in) |
| Admin creates account, active, notify checked | `register_admin_created` |
| Admin creates account, pending/blocked, or notify unchecked | none |
| Admin activates pending account | `status_activated` (core default, untouched) | 

Pre-merge checklist:


michael-wynne-wsu:
All five changes are done. Here's a summary of what was implemented:

- js/account-settings.js (new) - Attaches to the admin settings form on page load and on every change. Disables "May" and "Must" genpass radios when verify_mail is checked or admin-approval is selected; disables the verify_mail checkbox and admin-approval radio when May/Must genpass is active.
- mukurtu_core.libraries.yml - Added the account-settings library entry.
- mukurtu_core.module - Extended form_user_admin_settings_alter to attach the JS library and register a backend validation handler that sets form errors if an invalid combination is submitted.
- MailHooks.php - Extended mailAlter with a match block that suppresses: register_no_approval_required and register_pending_approval when verify_mail is off (user is auto-logged in, no OTL needed); register_admin_created when the account is not active (pending/blocked accounts get no email at creation).
- CommunityManagerUserCreationForm.php - Changed status_activated to register_admin_created so admin-created active accounts receive the correct welcome email with a one-time login link.

michael-wynne-wsu:
## PR Review: Branch `1695-registration-settings-fix`

**Blast Radius:** Medium -- changes touch `mukurtu_core` form/mail hooks (affect all user registration paths), `CommunityManagerUserCreationForm` (community-manager-created accounts), and the v4 theme (all forms with required fields). No DB schema changes, no shared API contracts broken.

**Security:** Clean -- no injection vectors, hardcoded secrets, or auth bypass patterns found.

**Update hooks:** None required. All changes are code/behavior, no stored config keys altered.

**Tests:** None added. Noted for the record; no existing test suite for these form alters.

---

### MUST FIX (Blocking)

**1. `setDisabled()` sets `disabled` but not `aria-disabled` -- screen readers can't announce why options are locked**
`account-settings.js:30-31`

Disabled form controls are removed from the accessibility tree. Users on screen readers can't discover the option exists or why it's unavailable. Fix:
```js
function setDisabled(el, state) {
  if (!el) return;
  el.disabled = state;
  el.setAttribute('aria-disabled', state ? 'true' : 'false');
}
```

**2. `syncState()` changes form state silently -- no live-region announcement for screen readers**
`account-settings.js:33-78`

When a user selects "Administrators only", options change with no announcement. WCAG 4.1.3 requires status changes to be programmatically determinable. Add a `role="status"` live region (injected by the JS behavior) and update it inside `syncState()` to describe the active constraint.

---

### SHOULD FIX (Non-blocking)

**3. `verifyMailEl.disabled` is set directly in two places instead of through `setDisabled()`**
`account-settings.js:44, 53`

Inconsistent with the helper -- once Issue 1 is fixed and `setDisabled` handles `aria-disabled`, these two direct assignments will be missed. Replace with `setDisabled(verifyMailEl, true/false)`.

**4. Validate function comment is now inaccurate**
`mukurtu_core.module` around line 306

The comment says: *"Genpass shows the generated password as a flash message on the registration page"* for the admin-approval mode. That's no longer true -- `visitorRegistrationPendingSubmit` clears all status messages (including genpass ones) for blocked/pending accounts. Update the comment to reflect the actual behavior.

**5. Silent `unset()` of genpass source groups could orphan future fields**
`mukurtu_core.module` -- `unset($form['genpass_config_generation'], $form['genpass_config_core_service'])`

The code moves specific known fields out of these groups and then deletes the groups entirely. If genpass adds a new field to either group in a future update, it would be silently dropped from the form with no warning. Consider a defensive log if the groups aren't empty after the moves.

---

### Informational

**6. `$isAdminCreated = \Drupal::currentUser()->isAuthenticated()` is correct for all current paths, but the assumption is implicit**
`MailHooks.php:50`

Works correctly for web form submissions. Edge case to be aware of: any programmatic account creation (e.g. migrations, Drush scripts running as anonymous) running while an admin is logged in elsewhere would not be affected, but any CLI-initiated creation would correctly have `isAuthenticated() = false`. Low risk, worth a brief comment explaining the assumption if not already done -- the existing comment does cover it well enough.

**7. Removed `formCommentCommentFormAlter` and `hideCommentBodyFormatHelp`**
`FormHooks.php` -- these two methods were deleted

Confirm this is intentional and the functionality is either no longer needed or was moved elsewhere.

---

### Looks Good

- Email gating matrix in `MailHooks` is thorough, well-documented, and correctly covers all six `register_*` keys
- `isNew()` guard in `formUserFormAlter` is the right fix for the root cause of the open-registration block bug -- clean and surgical
- Double-execution guards (`mukurtu_status_presave_ran` / `mukurtu_status_postsave_ran`) correctly handle the hook firing twice on `user_register_form`
- `CommunityManagerUserCreationForm` email type corrected from `status_activated` to `register_admin_created` -- those are meaningfully different emails and the fix is right
- `visitorRegistrationPendingSubmit` message replacement is clean -- clearing everything and showing one accurate message is better than trying to selectively filter
- Server-side validation in `mukurtu_core_user_admin_settings_validate` correctly mirrors the JS rules, so the JS is progressive enhancement with a real safety net behind it
- No security issues found across 560 lines of diff

---

**Summary:** Two accessibility blockers from the a11y audit need addressing before merge (Issues 1 and 2). Three non-blocking items are worth cleaning up. Core logic is sound.

michael-wynne-wsu:
All five issues addressed and committed. Quick summary of what changed:

- **`account-settings.js`**: `setDisabled()` now sets `aria-disabled` alongside `disabled`; live region injected at attach time and updated by `syncState()` with contextual announcements; direct `verifyMailEl.disabled` calls replaced with `setDisabled()`; `initializing` flag suppresses announcements on page load
- **`mukurtu_core.module`**: Validate function comment updated to reflect that genpass messages are suppressed for pending registrations; defensive `logger()->warning()` added before the `unset()` calls that delete genpass source groups

joelle-birano-wsu:
This flow makes sense to me. Do you think we need more in-form helper text on how Generate Password Settings interact with the other settings above? For example, if administrator approval is required, and visitors cannot set a password upon registration, and "Display when users create their own accounts" is selected, a password won't display. The success message when a user creates an account is very helpful, so that's fine. Just not sure if we want to add a little more for the admin choosing those settings.

michael-wynne-wsu:
Oh good catch. I chose to suppress some messages because if users requested an account that had to be reviewed they got a "a password has been generated for you"  message which is just not needed.

If anything I might remove the "Display when users create their own accounts" option entirely... 

Any other weirdness/more helper text needed?

joelle-birano-wsu:
> Oh good catch. I chose to suppress some messages because if users requested an account that had to be reviewed they got a "a password has been generated for you" message which is just not needed.

Agreed - that wasn't useful.

> If anything I might remove the "Display when users create their own accounts" option entirely...

I think that should work.

> Any other weirdness/more helper text needed?

I just noticed that if visitor, no approval required is selected, and they can't enter password, and email verification is not enabled, and generated password display is set to "Do not display," there's no way the user can get their password. They'll be logged in but if they log out without setting their password, they'll have to go through password reset to get back in, so if we're going to allow that I think users should immediately be routed to their account manage page with a message to set their password. 
Screenshot 2026-06-12 at 12 47 35 PM

michael-wynne-wsu:
Good finds. I've addressed them both (and resolved an issue where when directed to set a password it was still requiring you to provide the existing password).

Screenshot 2026-06-12 at 2 44 53 PM

Screenshot 2026-06-12 at 2 45 10 PM

joelle-birano-wsu (approved):
Those edits look great! That takes care of all the issues I could spot without being able to test email, so I think this is good.

nick-deer (approved):

PR #1703, opened 2026-06-02
Fix collection organization tree not showing on full collection view

### LINKED ISSUE #1701: Subcollections not showing up in collections
Sub-collections are not showing up in their top level collection pages after saving. 
They are staying assigned on the form, so this seems to be a regression.
First noticed on PR 1693

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1701

Collection trees should now render again.

Pre-merge checklist:


michael-wynne-wsu:
@christa-kavanagh-wsu when you test this, if it's working, make sure to test how private/inaccessible sub-collections display.

christa-kavanagh-wsu (changes requested):
Subcollections still aren't showing up. I added public and private subcollections under the top level collection and added a subcollection directly from the top level collection.

Image

Image

michael-wynne-wsu:
Good grief that took forever to hunt down a tiny change. It's working for me now on tugboat @christa-kavanagh-wsu

christa-kavanagh-wsu:
This works! 
image
But...now not all my collections are showing up in my Collections browse. They are showing up in regular browse fine.
image
image

michael-wynne-wsu:
What's odd there is that "collection 1" is showing up on /collections at all. Hmm.

Were collection 1 and 2 created the same way?

Oh, I created a new collection and added it to the top level collection and collection 1 correctly disappeared from the /collections page. Maybe an edit/save triggered something. See if you can replicate it, but if not, we can consider it a weird one off.

christa-kavanagh-wsu:
- Ok, so I tested this and now it's showing all of the subcollections on the /collections page.

image

- I tried adding a new collection to replicate what happened on yours yesterday, and that particular collection does not show up on the /collections page but all of the other subcollections do.

michael-wynne-wsu:
what workflow/order of operations did you use to create the collection/subcollections?

if you go in to the manage collection and organization settings and save without making any changes does it resolve?

christa-kavanagh-wsu:
Ok, so I think it might have to do with the workflow. When I create a new collection on its own and then add it to a top-level collection, it shows up as a unique collection on the /collections page. When I create a subcollection from the top-level collection it applies it properly as a subcollection.

The problem resolves when I go into collections management and hit save; it's like it doesn't properly assign them as subcollections until then.

michael-wynne-wsu:
Ok great, that is a specific workflow I can address, TBD!

michael-wynne-wsu:
Untested fix uploading now, give it a go.

michael-wynne-wsu:
Okay seems to be working for me!

michael-wynne-wsu:
## PR Review: `fix/collection-organization-tree-regression`

**Blast Radius:** MEDIUM ‚Äî touches collection hook registration, Search API indexing, and a theme template  
**Security:** No issues found  
**Tests:** No new tests (pre-existing pattern for this module)  
**Breaking Changes:** None ‚Äî no API or schema changes

---

### MUST FIX (Blocking)

**1. Double menu rebuild on every collection update ‚Äî negates the smart diff logic**

`CollectionEntityHooks::entityTypeUpdate()` (OOP hook via `#[Hook('node_update')]`) does a careful diff of `field_child_collections` and only calls `markRebuildNeeded()` when something meaningful changed. But `mukurtu_collection_entity_update()` in `.module` was also given `markRebuildNeeded()` in this PR and calls it **unconditionally** for every collection save.

Both hooks fire for the same event. The procedural one cancels out all the careful logic in the OOP one ‚Äî every collection save now triggers a menu rebuild regardless of what changed. Same issue applies to insert and delete (redundant but harmless since those always rebuilded anyway).

The `.module` file's `entity_insert/update/delete` hook bodies should **not** call `markRebuildNeeded()` ‚Äî that responsibility was migrated to `CollectionEntityHooks`. Remove those three additions.

---

### SHOULD FIX (Non-blocking)

**2. `\Drupal::entityTypeManager()` static call in an injected service**

`CollectionEntityHooks.php` uses the static container to load nodes and the index. The class already uses constructor injection for `MenuRebuildProcessor` and is registered as a service. `EntityTypeManagerInterface` should be injected for consistency and testability.

**3. `CollectionEntityHooks` defined twice ‚Äî attribute autodiscovery + explicit services.yml entry**

`mukurtu_collection.services.yml` explicitly registers `Drupal\mukurtu_collection\Hook\CollectionEntityHooks` as a service. Drupal 11 OOP hooks with `#[Hook]` are also auto-discovered. Verify this doesn't cause duplicate instantiation or double-firing. If explicit service registration is required for injection, that's fine ‚Äî just confirm it doesn't produce two instances.

---

### LOOKS GOOD

- **The regression fix itself** is clean and targeted. Diffing added/removed children and calling `trackItemsUpdated` with `index_directly: true` on the index is the right approach.
- **`CollectionPreprocessHooks` refactor** ‚Äî removing `#[Hook('preprocess_node')]` and delegating via a `.module` procedural hook is a valid workaround for the auth user regression. The comment explains the reason clearly.
- **Twig + SCSS changes** ‚Äî `<nav aria-labelledby>`, `<h3>` promotion, and the `:not(.collection__menu)` exclusion are all correct and validated clean by the accessibility audit.
- **No hardcoded secrets, no SQL injection vectors, no XSS surface.**

---

**Summary:** The regression fix (Search API reindex) is solid. Before merging, remove the `markRebuildNeeded()` calls added to the three procedural hooks in `.module` ‚Äî they conflict with the OOP hook's diff-based logic and cause unnecessary menu rebuilds on every collection save.

michael-wynne-wsu:
All three fixes applied. Summary of changes:

- **mukurtu_collection.module** ‚Äî Removed `markRebuildNeeded()` from `entity_insert`, `entity_update`, and `entity_delete` hooks. Menu rebuild is now handled exclusively by `CollectionEntityHooks` with its smart diff logic.
- **CollectionEntityHooks.php** ‚Äî Added `EntityTypeManagerInterface` constructor injection; replaced both `\Drupal::entityTypeManager()` static calls with `$this->entityTypeManager`.
- **mukurtu_collection.services.yml** ‚Äî Added `@entity_type.manager` as the second argument for `CollectionEntityHooks`.

christa-kavanagh-wsu:
When I use the same workflow all of the collections I added to the top level collection as subcollections are still showing up on the /collections page.

michael-wynne-wsu:
Then I don't understand your workflow properly. What I did was:

1. Create multiple collections, fully independently.
2. On the target top-level collection, click "manage collection organization".
3. Search for and add the target sub-collections > save.

christa-kavanagh-wsu:
Ah, understood. I created Collection 1 and Collection 2 completely independently. Then I created a collection I called Top-level Collection and added Collection 1 and Collection 2 as subcollections in the Top-level Collections form.

michael-wynne-wsu:
Gotcha, I will target that!

michael-wynne-wsu:
@christa-kavanagh-wsu confirmed this is fixed for all workflows now! sending to @nick-deer for code review

christa-kavanagh-wsu (approved):
Works for me!

nick-deer (approved):

PR #1707, opened 2026-06-04
Add patches to fix implicit nullable parameter deprecations in contri…

michael-wynne-wsu:
Patches admin_theme, message_digest, message_ui, and search_api_glossary to use explicit ?Type nullable syntax, eliminating PHP 8.x deprecation notices. Also upgrades composer/installers from ^1.9 to ^2.0, which natively resolves the deprecations in that package.

Pre-merge checklist:


michael-wynne-wsu:
@nick-deer @brian-wsu so I still see one deprecation warning in the build logs but it might be a tugboat thing and I don't think it's an issue and doesn't seem to surface on the front end. All the other ones are patched with this.

michael-wynne-wsu:
## PR Review: Add patches to fix implicit nullable parameter deprecations (#1707)

Blast Radius: LOW | Security: No findings | Breaking Changes: None

### MUST FIX
None.

### SHOULD FIX
1. ~~Orphaned patch file `patches/composer_installers-nullable-param-deprecation.patch`~~ — RESOLVED (deleted)

### SUGGESTIONS
2. ~~PR description missing message_ui~~ — RESOLVED (PR updated)
3. Tugboat init-phase pin (~line 44 in .tugboat/config.yml) is now redundant since composer.json already requires ^2.0. Low priority — consider removing or adding an explanatory comment.

### LOOKS GOOD
- All five patches are minimal and correct (only adds ? prefix to nullable type hints)
- composer/installers upgrade to ^2.0 is the right approach
- No update hooks, schema changes, or accessibility impact
- message_digest patch covers all three affected locations

michael-wynne-wsu:
Both fixes applied:

Deleted [patches/composer_installers-nullable-param-deprecation.patch](vscode-webview://0221t97r27mcm5751n6kf139ul9pnbh5fhrou3bcr85pcs3mp69j/patches/composer_installers-nullable-param-deprecation.patch)
Updated PR #1707 description to include message_ui

nick-deer:
This cleared up the old warnings, but a new one cropped up for me:

PHP Deprecated:  Drupal\message_digest\Plugin\Notifier\DigestBase::__construct(): Optional parameter $message declared before required parameter $time is implicitly treated as a required parameter in /var/www/html/web/modules/contrib/message_digest/src/Plugin/Notifier/DigestBase.php on line 76

michael-wynne-wsu:
Interesting. I see one about line 136, but not that one. I will explore!

nick-deer:
I've cleared up some PHP deprecation warnings in the past and it seems like they only show warnings one at a time for each file... strange...

michael-wynne-wsu:
@nick-deer see if that sorts it out for you. These are kind of whackamole as they pop up...

nick-deer (approved):
That was the last of them!

michael-wynne-wsu:
Yeehaw!

PR #1687, opened 2026-05-30
983 Grant protocol stewards comment edit/delete/review access

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/983

Protocol stewards should be able to:

- [ ] Edit their own and other users comments within their protocol
- [ ] Delete their own and other users comments within their protocol
- [ ] Review comments within their protocol
- [ ] Approve comments within their protocol

Protocol stewards already hold administer comments at the OG level, but MukurtuCommentAccessControlHandler only overrode the approve operation, letting edit/delete fall through to core's site-level-only check. Added update/delete handling that checks protocol-level permission via the commented entity. Also relaxed force_require_all in loadThread from TRUE to FALSE so stewards in any of the content's protocols can see unpublished comments.

Pre-merge checklist:

- [ ] I have checked for and if required, created update hooks.
- [ ] I have run an accessibility check, and resolved any issues.
- [ ] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [ ] I have updated from `main` and resolved any merge conflicts.
- [ ] I have verified that all expected checks pass.
- [ ] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1686, opened 2026-05-30
Comments styling

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/262

- [ ] Comments now have styling.
- [ ] Test comment, reply, and admin review.

Pre-merge checklist:

- [ ] I have checked for and if required, created update hooks.
- [ ] I have run an accessibility check, and resolved any issues.
- [ ] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [ ] I have updated from `main` and resolved any merge conflicts.
- [ ] I have verified that all expected checks pass.
- [ ] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1710, opened 2026-06-05
Comments super branch

### LINKED ISSUE #262: Comment subject
Can this field be removed? I feel like it will be confusing.

### LINKED ISSUE #384: Site-wide comment settings

### LINKED ISSUE #887: Community and Protocol level comment settings

### LINKED ISSUE #983: Comment management permissions
Protocol stewards should be able to do the following to their own and other users comments (currently only Mukurtu managers can take these actions): 

- [ ] Edit their own and other users comments
- [ ] Delete their own and other users comments
- [ ] Review comments within their protocol
- [ ] Approve comments within their protocol

~~From @christa-kavanagh-wsu on #550~~ 

~~I ran into issues with editing and deleting comments: none of the protocol user roles could edit or delete their own comments. Protocol stewards couldn't review, approve, edit, or delete any comments either, including their own. I wonder if the Protocol steward access is tied to their lack of access to the "Manage Content" link in the Dashboard; that is where I batch approve comments as an admin.
I updated my findings on the [sheet](https://emailwsu.sharepoint.com/:x:/r/teams/libraries.mukurtudevelopment/_layouts/15/Doc.aspx?sourcedoc=%7B0a972509-d87f-4cbf-9cb6-dd63c487b355%7D&action=edit&activeCell=%27Protocol%20roles%27!B20&wdinitialsession=abc2121b-c933-88b9-3de0-222b7100863e&wdrldsc=6&wdrldc=1&wdrldr=AccessTokenExpiredWarningUnauthenticated%2CRefreshin) as well.~~

michael-wynne-wsu:
Combines https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1690, https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1689, https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1687, and https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1686

Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/887, https://github.com/MukurtuCMS/Mukurtu-CMS/issues/384, https://github.com/MukurtuCMS/Mukurtu-CMS/issues/983, and https://github.com/MukurtuCMS/Mukurtu-CMS/issues/262

Pre-merge checklist:


michael-wynne-wsu:
## PR Review: `comments-super-branch`

**Blast Radius:** HIGH — touches comment access control, protocol entity base fields, filter formats, and field formatter used on all Digital Heritage nodes.
**Security:** 1 finding (critical)
**Tests:** No new tests (pre-existing gap; Mukurtu test coverage is limited)
**Breaking Changes:** None for external API consumers; update hooks required for existing sites.

---

### MUST FIX (Blocking)

**1. `CulturalProtocolControlledInterface` used but not imported**
`MukurtuCommentAccessControlHandler.php:90`

The `use` statement for `CulturalProtocolControlledInterface` was removed when the old `checkAccess` logic was rewritten, but `checkCreateAccess` still uses it. PHP will throw a fatal error whenever comment create access is checked for a protocol-controlled entity.

```php
// Add this to the use block:
use Drupal\mukurtu_protocol\CulturalProtocolControlledInterface;
```

✅ Fixed in commit `bb3ba4a9`

---

### SHOULD FIX (Non-blocking)

**2. Post access enforcement is one-directional**
`ProtocolCommentSettingsForm.php:733-737`

The form only strips `anonymous` from post access if `anonymous` can't view. But if only `protocol_member` can view, `authenticated` users can still be saved as post access values - even though they can't see the comments. Consider enforcing post access must be a subset of view access (not just the anonymous case).

**3. `mukurtu_core_update_40034` sets an incorrect format key**
`mukurtu_core.install:180`

Sets `allowed_formats` to `['basic_html' => 'basic_html']` — an existing but unrelated format. Hook `_40035` immediately overwrites this with the correct `comment_basic_html` key. The intermediate state is harmless since both run in the same `drush updb` batch, but if `_40035` ever fails, sites are left with `basic_html` restricting comments instead of the new format. Consider setting the correct value directly in `_40034`, or collapsing both hooks into one.

**4. `filter_html_nofollow: false` in `comment_basic_html` format**
`filter.format.comment_basic_html.yml` / `mukurtu_core.install:210`

Links posted in comments will not receive `rel="nofollow"`. For a CMS used by Indigenous communities that may have public-facing content, this could encourage comment spam. Confirm this is intentional — if not, set `filter_html_nofollow: true`.

**5. `hasSiteOrProtocolPermission` `$force_require_all` changed from `TRUE` → `FALSE` in three places**
`MukurtuCommentStorage.php:96`, `MukurtuCommentDefaultFormatter.php:48,51`

Changing `$force_require_all` to `FALSE` means on a content item with multiple protocols using the "any" sharing setting, a user who is a comment admin on just **one** of those protocols can now see and manage all pending comments. This is a deliberate permission relaxation — confirm it matches the intended design for protocol steward comment moderation scope.

---

### LOOKS GOOD

- **`MukurtuCommentFieldItemList::accountMatchesAccessList`** is well-factored as a public static helper and is reused correctly in the access handler.
- **`#after_build` for format help link** correctly defers hiding until after `filter_process_format()` runs — this is the right pattern.
- **Cache tags on access results** are correctly applied (`user:{id}`, entity, and commented entity dependencies).
- **Update hook numbering** is consistent and non-conflicting across all three modules (40034/40035 in core, 40035/40036/40037 in protocol, 40005 in digital_heritage).
- **`comment_aria_label` preprocess variable** avoids Twig sandbox issues by computing the label server-side.
- **`#title_display = 'invisible'`** keeps the comment body label in the DOM for screen readers while hiding it visually — correct approach.
- **Config schema** updated for `site_comments_require_approval` in `mukurtu_protocol.schema.yml`.

michael-wynne-wsu:
## Session Summary — `comments-super-branch`

### Merge Conflict Resolution
Resolved two conflicts from merging `feature/per-protocol-comment-access`:
- `style.css.map` — auto-generated file; resolved by recompiling SCSS with `npx gulp buildSass`
- `mukurtu_protocol.install` — both branches had written a `_40035` hook; resolved by keeping HEAD's `_40035` (site_comments_require_approval default) and renumbering the incoming hooks to `_40036` (install field storage definitions) and `_40037` (remove old fields from form display)

### Accessibility Check (WCAG 2.2 AA)
Two issues found and fixed:
- **WCAG 1.3.1 / 2.4.6** — Added `aria-label="Comment by @author on @date"` to each comment `<article>` via a computed `comment_aria_label` preprocess variable in `mukurtu_v4.theme`
- **WCAG 2.4.7** — Added explicit `focus-visible` outline to comment action links in `_comment.scss`

### PR Review
One PHP fatal found and fixed; four non-blocking issues flagged:
- **Fixed** — Missing `use Drupal\mukurtu_protocol\CulturalProtocolControlledInterface` in `MukurtuCommentAccessControlHandler.php` (would have caused a fatal error on any comment create access check)
- **Flag** — Post access enforcement only strips `anonymous` when view is restricted, not `authenticated` when view is limited to `protocol_member` only
- **Flag** — `mukurtu_core_update_40034` sets wrong format key (`basic_html` instead of `comment_basic_html`), superseded by `_40035` but fragile
- **Confirm** — `filter_html_nofollow: false` means comment links won't get `nofollow` — intentional?
- **Confirm** — `hasSiteOrProtocolPermission($force_require_all)` changed from `TRUE` to `FALSE` in storage and formatter — more permissive for protocol stewards

### Commits
- `c8a12d7b` Merge branch 'feature/per-protocol-comment-access' into comments-super-branch
- `bb3ba4a9` Fix missing use import, add WCAG focus style, and comment aria-label

michael-wynne-wsu:
Done. Three changes committed (`784a76d5`):

1. **`filter_html_nofollow: true`** — comment links now get `rel="nofollow"` in both the config file and the update hook
2. **Post access strict subset** — `array_intersect($postAccess, $viewAccess)` replaces the anonymous-only strip; any role unchecked from view access is also removed from post access
3. **Client-side `#states` updated** — all three checkboxes in the post access group are now individually disabled when their corresponding view access checkbox is unchecked

christa-kavanagh-wsu (changes requested):
- Protocol stewards don't get a confirmation message when they change the comment settings in the protocol. Comments are enabled across protocols and the site

Image

- Visitors can't comment or reply to comments on content when visitor comments have been enabled. They enter text in the Name field, but can't save anything because the comment field can't be null.

Image

- Protocol stewards can't edit comments in their protocol.

Image

- Users who are protocol members can't comment on content in their protocol.

Image

- Protocol stewards who create content can't comment on the content or any content in their protocol.

Image

- As a protocol steward if the protocol is set to comments needing approval, it doesn't give me the option to approve comments. 
- I can't test approving comments at the site level because I can't comment as anything other than an admin.
- I can delete comments from the admin.
- As a Mukurtu Manager, I can technically comment but I have this weird text format box that doesn't let me do anything but says I have to select an item in the list? 

Image

I think that's all I've run across; there is a lot here that I think overlaps.

michael-wynne-wsu:
Tackling two main areas first:

1. Protocol steward edit/approve permissions.
2. Going to change the text format from basic html to plain text. That's the source of not being able to enter comment text, and I don't think I want to expose any text formatting tools.

michael-wynne-wsu:
Pushed updates but I have not tested them, gotta go drop off the kids. Planning to test locally once I'm back. You can test, just know I haven't vetted it yet

christa-kavanagh-wsu:
- Protocol steward edit/approve permissions are now working! 
- Plain text is working, but the helper text is probably not necessary.
image

michael-wynne-wsu:
Great, I removed that helper text.

So I'm able to leave comments anonymously now, haven't tested as more generic users, but the format change should probably have resolved that.

Can you test again and let me know which issues are still going on? Some additional thoughts as you do that:

- [ ] I think we should add a required email field or something for visitor comments.
- [ ] We need to at least internally map out how site-wide and protocol settings interact so we can provide better helper text. I can probably start that, unless you have a good grasp on it.
   - [ ] eg: right now I think if site anonymous comments are enabled, and protocol allows anonymous view but not add comments, visitors CAN leave comments which is not right, protocol should take priority.
- [ ] I also want to add more explicit enforcement > if site-wide commenting is disabled, prevent accessing those settings in protocol.

christa-kavanagh-wsu:
> So I'm able to leave comments anonymously now, haven't tested as more generic users, but the format change should probably have resolved that.

Generic users can leave comments.
 
>     * [x]  I think we should add a required email field or something for visitor comments.

Agreed, something should be there.

>     * [ ]  We need to at least internally map out how site-wide and protocol settings interact so we can provide better helper text. I can probably start that, unless you have a good grasp on it.
>       
>       * [ ]  eg: right now I think if site anonymous comments are enabled, and protocol allows anonymous view but not add comments, visitors CAN leave comments which is not right, protocol should take priority.

That is not what I am seeing right now; if site-wide comments are enabled and protocol comments are disabled, visitors can't leave comments. 

>     * [ ]  I also want to add more explicit enforcement > if site-wide commenting is disabled, prevent accessing those settings in protocol.

That is a good idea.

Overall the changes you made seem to have fixed the problems I was having with things earlier. I can go through and explore the interactions between site-wide and protocol comment settings a little more, but it seems fairly straightforward from what I have come across.

michael-wynne-wsu:
Adding a "require email for anonymous comments" site-wide setting

[CommentSettingsForm.php](vscode-webview://0221t97r27mcm5751n6kf139ul9pnbh5fhrou3bcr85pcs3mp69j/modules/mukurtu_protocol/src/Form/CommentSettingsForm.php) - New checkbox "Require email address from visitors who leave comments" added after the existing "Allow visitors to leave comments" toggle. It's disabled (via #states) when anonymous posting is unchecked. The value is saved to mukurtu_protocol.comment_settings.

[MukurtuCommentForm.php](vscode-webview://0221t97r27mcm5751n6kf139ul9pnbh5fhrou3bcr85pcs3mp69j/modules/mukurtu_protocol/src/MukurtuCommentForm.php) - After reading the field-level anonymous setting (line 50), it now checks the site config. If anonymous_comments_require_email is true and the current user is anonymous, $anonymous_contact is forced to ANONYMOUS_MUST_CONTACT - which makes the existing email field required and visible (lines 151/155 already handle this correctly).

[mukurtu_protocol.install](vscode-webview://0221t97r27mcm5751n6kf139ul9pnbh5fhrou3bcr85pcs3mp69j/modules/mukurtu_protocol/mukurtu_protocol.install) - Added mukurtu_protocol_update_40038() at the end, which initializes the new config key to FALSE for existing sites.

To test: go to /admin/config/mukurtu/commenting/settings, enable anonymous posting, check the new "Require email" box, then visit a node as a logged-out user and try to comment.

Screenshot 2026-06-05 at 10 26 17 PM

Screenshot 2026-06-05 at 10 40 09 PM

michael-wynne-wsu:
Visually enforcing site-level restrictions at protocol levels (so people can't make setting changes at the protocol level that are actually not doing anything).

- If site comments are disabled, protocols can't enable them.
- Similarly, if site disallows visitor view/comment, protocols can't allow that.

Screenshot 2026-06-05 at 10 38 25 PM
Screenshot 2026-06-05 at 10 38 57 PM

michael-wynne-wsu:
## PR Review: comments-super-branch

**Blast Radius:** HIGH - touches access control handlers, entity field definitions, config schema, and Twig templates across 30 files.
**Security:** 0 blocking findings, 1 access-behavior change to confirm.
**Breaking Changes:** 1 confirmed (schema gap); 1 intentional behavior change to verify.
**Tests:** 0 test files changed.

---

### MUST FIX (Blocking)

**1. Config schema missing `anonymous_comments_require_email`**

`modules/mukurtu_protocol/config/schema/mukurtu_protocol.schema.yml` has `site_comments_require_approval` added, but `anonymous_comments_require_email` is absent. Drupal will log config schema validation errors for every form submit and config import on this key.

```yaml
# Add to mukurtu_protocol.comment_settings mapping:
anonymous_comments_require_email:
  type: boolean
```

---

### SHOULD FIX (Non-blocking)

**2. `$site_check` flipped from `TRUE` to `FALSE` in storage and formatter - site-wide admins lose unpublished comment visibility**

In `MukurtuCommentStorage.php` and `MukurtuCommentDefaultFormatter.php`, two calls to `CulturalProtocols::hasSiteOrProtocolPermission(...)` changed their `$site_check` argument from `TRUE` to `FALSE`. This means users with site-wide `administer comments` permission who are NOT protocol members will:

- No longer see unpublished comments in the thread (storage)
- No longer see the comment section on content with zero published comments (formatter)

The intent appears to be confining comment administration to protocol members only, but this could break the workflow of site-wide admins (e.g., `mukurtu_manager` role) who need cross-protocol comment oversight without being a member of every protocol. Confirm this is intentional and document it in the PR body.

**3. `unapprovedComments()` page has no cache tags - stale after approvals**

`ProtocolCommentSettingsController::unapprovedComments()` returns a render array with no `#cache` metadata. After a comment is approved via the operations links on that page, the listing stays stale until caches expire. Add cache tags so approval invalidates the display:

```php
return [
  '#type' => 'table',
  ...
  '#cache' => ['tags' => ['comment_list']],
];
```

---

### SUGGESTIONS

**4. `mukurtu_core_update_40034` docblock doesn't match what the hook actually does**

The docblock says "restrict comment body to Basic HTML only" but the final released format is `plain_text`. The hook sets `basic_html`, which is then overwritten by 40035 and 40036. The docblock should say what the hook does (hides subject field, sets basic_html as intermediate step) or the function body should be simplified since later hooks supersede it. This creates confusion when reading the install file.

**5. Protocol-level update/delete access expansion is undocumented**

`MukurtuCommentAccessControlHandler::checkAccess()` now grants protocol-level `administer comments` users the ability to `update` and `delete` any comment on their protocol's content - including published comments and comments by other users. This is likely correct and intentional, but it's a meaningful permission expansion that isn't called out in the PR description. Worth a sentence in the PR body so reviewers know it's deliberate.

**6. `unapprovedComments()` missing batch protection on entity loads**

```php
$comments = $this->entityTypeManager()->getStorage('comment')->loadMultiple($comment_ids);
```

If a protocol has thousands of unapproved comments (spam scenario), this loads all of them into memory at once. The admin comment queue at `/admin/content/comment` uses Views with pagination. Consider adding a `->range()` limit or deferring to a View for the listing.

---

### LOOKS GOOD

- **Twig null-guard fix** (`field_media_assets ? ...|length : 0`) across four node templates - catches a crash when the field has no value. Clean one-liner fix.
- **`MukurtuCommentForm` anonymous email logic** - reading `anonymous_comments_require_email` from config to override `$anonymous_contact` is minimal and correctly placed before the field renders.
- **`ProtocolCommentSettingsForm` site ceiling enforcement** - the `#after_build` pattern for disabling the 'anonymous' checkbox mirrors the existing `addPostAccessStates()` pattern. The `submitForm()` stripping of 'anonymous' is a solid defense against JS bypass.
- **Update hook chain** - the sequence 40034 through 40038 handles the upgrade path correctly. 40036 properly deletes the editor config before the filter format to respect config dependencies.
- **`MukurtuCommentFieldItemList::accountMatchesAccessList()`** - clean, readable, correct. The `public static` visibility is the right choice since it's reused in the access handler.
- **Comment Twig template** - the `aria-label` on the article element and the `visually-hidden` permalink span are good accessibility practice. The `content|without('links')` separation into a `<footer>` is semantically appropriate.

---

**Summary:** One schema gap is a definite bug to fix before merging. The `$site_check = FALSE` changes in the storage/formatter are the most consequential behavioral shift - they may be exactly right for the protocol-first model, but they need an explicit acknowledgment in the PR. Everything else is clean.

michael-wynne-wsu:
Committed as `a8f29a47`. Here's what was done:

| Finding | Fix |
|---|---|
| Schema missing `anonymous_comments_require_email` | Added `type: boolean` entry to `mukurtu_protocol.schema.yml` |
| `$site_check=FALSE` locked out site-wide admins | Reverted to `TRUE` in `MukurtuCommentStorage` and `MukurtuCommentDefaultFormatter` (both occurrences) |
| Unapproved comments page stays stale after approval | Added `#cache => ['tags' => ['comment_list']]` to render array |
| Unbounded `loadMultiple` on comment query | Added `->range(0, 200)` to the comment query |
| `_update_40034` docblock said "Basic HTML only" (misleading) | Clarified that hooks 40035/40036 supersede the format change |
| Update/delete permission expansion undocumented | Added explanatory comment in `MukurtuCommentAccessControlHandler` |

michael-wynne-wsu:
Okay I have implemented a change to requiring approval on comments that mirrors other settings.
- If site-wide comment settings require comment approval, then protocol level settings MUST ALSO require comment approval.
- If site-wide comment settings do not require comment approval, then protocol level settings can either require comment approval or not.

nick-deer (approved):

PR #1653, opened 2026-05-21
823 grid browse remote video display

michael-wynne-wsu:
Closes #823 

Grid and list view teasers now both use the same YouTube player.
I don't think this needs any manual documentation updates?

Pre-merge checklist:

- [x] I have checked for and if required, created update hooks.
- [x] I have run an accessibility check, and resolved any issues.
- [x] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [x] I have updated from `main` and resolved any merge conflicts.
- [x] I have verified that all expected checks pass.
- [x] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1715, opened 2026-06-05
Fix remote video teaser display oddities (#1653)

michael-wynne-wsu:
Addresses regressions in remote video display from https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1653

_vertical-card.scss Added aspect-ratio: 16/9 on remote video iframes inside the grid card's media container. This fixes the Vimeo black letterboxing in the grid view.

_horizontal-card.scss Same aspect-ratio: 16/9 fix applied to the list card media container, for consistency.

mukurtu_media.module Added hook_oembed_resource_data_alter() that appends rel=0&modestbranding=1 to YouTube embed URLs. This removes the YouTube logo from the control bar and suppresses related videos. It applies globally (all YouTube embeds), not just browse cards - which is appropriate since these are site-wide quality improvements.

Remaining limitations (YouTube platform restrictions, not fixable from our side):

- The title/channel info bar that appears on hover/pause will still show
- "Watch on YouTube" link may still appear in some player states
- Pixelated thumbnails for specific videos are YouTube's CDN responsibility

Screenshot 2026-06-05 at 3 24 03 PM

Screenshot 2026-06-05 at 3 24 09 PM

Pre-merge checklist:


michael-wynne-wsu:
## PR Review: Fix remote video teaser display oddities

**Blast Radius:** LOW - scoped to browse card layout and YouTube oEmbed rendering. The PHP hook does apply globally (all YouTube embeds site-wide, not just browse cards), which is intentional and beneficial.
**Security:** No findings
**Tests:** None expected - standard for Drupal hook + SCSS changes
**Breaking Changes:** None

---

### MUST FIX (Blocking)

None.

---

### SHOULD FIX (Non-blocking)

**1. `_horizontal-card.scss` - potential mobile conflict between `block-size: revert-layer` and `aspect-ratio`**

The existing rule sets `block-size: revert-layer` on mobile for all `.media iframe` elements. Combined with the new `aspect-ratio: 16/9` on `.media--remote_video iframe`, behavior on mobile is implementation-defined - `revert-layer` could reinstate the iframe's HTML `height` attribute, which would then fight `aspect-ratio`. On desktop `block-size: auto` + `aspect-ratio` is fine.

Low risk since remote video teasers are less common on mobile, but worth testing on a small screen. If issues appear, add an explicit `height: auto` to the remote video rule on mobile.

---

### SUGGESTIONS

**2. `mukurtu_media.module` - `$url` parameter is unused**

The function signature accepts `string $url` but the body only checks `$data['provider_name']`. PHP won't error on this, but a reader might wonder if the URL check was meant to be there. Either drop the type hint and use `$_url` to signal intentional non-use, or add an inline comment explaining why `$provider_name` is the preferred discriminator. (Either way is fine for a procedural hook.)

**3. `mukurtu_media.module` - hook applies to all YouTube embeds, not just browse cards**

The hook fires globally, so full-page Digital Heritage views and any other YouTube embed site-wide will also get `rel=0&modestbranding=1`. This is almost certainly desirable behavior, but worth a one-line note in the doc comment so future developers don't inadvertently scope it.

---

### LOOKS GOOD

- `preg_replace_callback` is correctly used; no user input reaches the replacement string, so no injection risk
- `str_contains($src, '?')` correctly handles both fresh and already-parameterized oEmbed URLs
- `provider_name === 'YouTube'` is the right discriminator - matches exactly what YouTube's oEmbed API returns
- The vertical-card fix (constrain the container, not just the iframe) is the correct approach - `width/height: 100%` on the iframe inside a sized flex container is reliable across browsers
- `overflow: hidden` on `.media--remote_video` is appropriate insurance against any player chrome that might bleed past the container boundary
- Compiled `style.css` and `style.css.map` are included, consistent with project workflow
- WCAG 2.2 AA: no accessibility regressions (verified separately)

---

**Verdict:** Ready to merge after verifying mobile horizontal-card rendering and confirming hard-refresh resolves the YouTube chrome cache issue observed in the grid view screenshot.

michael-wynne-wsu:
All three fixes committed to `fix/remote-video-teaser-display`:

1. **Mobile height conflict** - Added `block-size: auto` to the horizontal card's `.media--remote_video iframe` rule so it explicitly overrides `revert-layer` before `aspect-ratio` does its work.
2. **Unused `$url` param** - Renamed to `$_url` to signal intentional non-use.
3. **Doc comment** - Clarified that the hook applies to all YouTube embeds site-wide, not just browse cards.

nick-deer (approved):

PR #1723, opened 2026-06-11
Fix #1722: Gin selected content actions dropdown z-index

### LINKED ISSUE #1722: Gin selected content z-index issue
in Gin, when content is selected and then the actions dropdown is expanded, they expand under the lower bulk actions, obscuring the selections.

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1722

Give .views-bulk-actions[data-drupal-sticky-vbo=true] an explicit z-index: 3 so it sits below the sticky edit-header stacking context (z-index: 4). This prevents the lower bulk-actions bar from painting over the open Actions dropdown.

Pre-merge checklist:


michael-wynne-wsu:
## PR Review: Fix #1722 - Gin selected content z-index

**Blast Radius:** LOW - CSS-only, scoped to Gin admin views pages
**Security:** Clean
**Tests:** N/A (CSS-only)
**Breaking Changes:** None detected

---

### MUST FIX (Blocking)

None.

---

### SHOULD FIX (Non-blocking)

None.

---

### SUGGESTIONS

**1. Stacking context side-effect is worth noting**

Assigning `z-index: 3` to `.views-bulk-actions` promotes it from a non-stacking element to a stacking context. Any absolutely-positioned descendants inside `.views-bulk-actions` that currently "escape" their container (e.g. a future tooltip or popover) would be clamped to that level-3 context. This is unlikely to matter today since the bar only contains a select + buttons, but it's a property that can bite later. Documenting it in the comment would be helpful.

Suggested comment addition:
```css
/* Note: this intentionally creates a stacking context on .views-bulk-actions
   at level 3. Any absolutely-positioned descendants will be confined to that
   context — currently not an issue since the bar contains only flat elements. */
```

**2. Consider whether non-sticky state needs the same guard**

The `[data-drupal-sticky-vbo=true]` attribute is applied by VBO's JS once the bar enters sticky mode. The stacking conflict exists as soon as the bar is in the DOM, not only when it's sticky. In practice the overlap only surfaces when scrolled to the sticky position, so the current scope is fine — but worth knowing if the bug is ever reported in a non-scrolled state.

---

### LOOKS GOOD

- `z-index: 3` is the exact right value: one below Gin's `z-index: 4` on the sticky header, not an arbitrary large number
- Selector mirrors Gin's own pattern from `tableselect.css` - consistent specificity, no `!important`
- Comment references the issue number and explains the root cause
- Zero risk of regressions: the rule only applies inside `.block-system > .views-form` and `.view-content .views-form` admin contexts
- WCAG audit from the accessibility check found no issues with this change

The change is minimal, correct, and ready to merge.

nick-deer (approved):

PR #1725, opened 2026-06-12
Taxonomy export and import

### LINKED ISSUE #1483: Import taxonomy terms
Do we need this to be custom, or can we get away with https://www.drupal.org/project/taxonomy_manager ?

### LINKED ISSUE #1698: Taxonomy export tools
- We want to extend the import and export tools to cover taxonomy terms as well (currently they are focused on content).
- This could make use of existing drupal modules, or be custom work.
- In most cases users would be exporting one or more complete taxonomies. But we don't necessarily want to omit the option to export a selection of terms.
- Users may be importing wholly complete/new taxonomies, or updating/extending existing ones.
- Permissions would likely be tied to the roundtrip manager, mukurtu manager, and administrator roles.

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1698

- [ ] Adds a vocabulary picker form at `/admin/export/taxonomy` (linked on dashboard) that lets Roundtrip Managers and Mukurtu Managers select one or more vocabularies and export all their terms to CSV using the existing batch export pipeline.
- [ ] Import uses the existing Import path and process. There are now taxonomy import settings that probably need to be brought in line with the others.
- [ ] Updated helper/description text and ensure the category importer and exporter use the correct thumbnail structure.

Pre-merge checklist:


michael-wynne-wsu:
I haven't tested this on tugboat but it's working well on my local. I think it's pretty self-explanatory given the existing import/export work, but let me know if you need more before testing.

joelle-birano-wsu (approved):
Works well for me!

christa-kavanagh-wsu (approved):
This works well for me! Are the taxonomy import formats going to be included in import format information? They're already in import templates.

michael-wynne-wsu:
> This works well for me! Are the taxonomy import formats going to be included in import format information? They're already in import templates.

Good catch, I will make sure to write that.

michael-wynne-wsu:
@christa-kavanagh-wsu  Okay that's been added - one for categories (because of the images) and one template for all other taxonomies. Probably need to update the descriptions on those two pages though.

michael-wynne-wsu:
Updated helper/description text and ensure the category importer and exporter use the correct thumbnail structure.

michael-wynne-wsu:
## PR Review: feature/taxonomy-export-import

**Blast Radius:** MEDIUM - extends two existing modules (mukurtu_export, mukurtu_import) with new entity type support; the migration destination hook change (`entity:taxonomy_term`) affects all taxonomy term imports going forward.

**Security:** No findings.

**Tests:** No new tests. No existing test files changed. Follows existing module pattern.

**Breaking Changes:** None detected. All changes are additive.

---

### MUST FIX (Blocking)

**1. Static `\Drupal::entityTypeManager()` in an injected service class**

`ProtocolAwareEntityContent.php:199`

`applyMediaEntityAltText()` calls `\Drupal::entityTypeManager()` statically despite `$this->storage` already being injected. The storage object is typed as `EntityStorageInterface` which doesn't expose `getEntityTypeManager()`, but the entity type manager is available on the container. The cleanest fix is to inject it in `__construct` / `create()` the same way `$current_user` was added.

```php
// In __construct parameter list, add:
protected EntityTypeManagerInterface $entityTypeManager,

// In create(), add:
$container->get('entity_type.manager'),

// Then replace the static call:
$media_storage = $this->entityTypeManager->getStorage('media');
```

---

### SHOULD FIX (Non-blocking)

**2. N+1 query in the `TaxonomyTermImportStatus` Views field**

`TaxonomyTermImportStatus.php:55-63`

`render()` fires one `COUNT(*)` query against `taxonomy_term_revision` per row. For a 500-term import this is 500 queries on the results page. The standard Views approach is to override `preRender(array $values)` to batch-load all tids at once, cache results in a property, then read from cache in `render()`.

```php
protected array $revisionCounts = [];

public function preRender(array $values): void {
  $tids = array_filter(array_map(fn($v) => (int) $this->getValue($v), $values));
  if (empty($tids)) return;
  $result = $this->database->select('taxonomy_term_revision', 'r')
    ->fields('r', ['tid'])
    ->condition('r.tid', $tids, 'IN')
    ->groupBy('r.tid')
    ->addExpression('COUNT(*)', 'cnt');
  // ... store into $this->revisionCounts[$tid]
}

public function render(ResultRow $values): string {
  $tid = (int) $this->getValue($values);
  return ($this->revisionCounts[$tid] ?? 1) === 1 ? $this->t('New') : $this->t('Updated');
}
```

**3. Gap in export update hook numbering**

`mukurtu_export.install` has hooks 40001, 40002, 40004, 40005, **40009** - skipping 40006, 40007, 40008. If those numbers are reserved for parallel branches that haven't merged yet, fine - but if this is unintentional the hook should be renumbered to 40006 (next available) to avoid confusion when future developers look at the sequence.

**4. `update_40029` unconditionally overwrites config**

`mukurtu_import.install` - hooks 40028 and 40029 both target `views.view.mukurtu_import_results_taxonomy_terms`. Hook 40028 creates it if new; hook 40029 unconditionally overwrites it. Any site that customized the view will have their changes silently reverted on update. If the intent is to always reset to the shipped version, add a comment explaining that. If not, add a `$config->isNew()` guard consistent with the pattern in every other strategy-create hook.

---

### SUGGESTIONS

**5. `exportEntityReference` loads entities one-by-one inside a loop**

`CsvEntityFieldExportEventSubscriber.php` - for `taxonomy_term` and `user` reference options, `getStorage()->load($id)` is called per reference value inside `foreach ($field->getValue() ...)`. For a multi-value term field with many values this is fine in practice (term fields rarely have hundreds of values), but worth noting if batch export of large content sets is a goal.

**6. `EntityReference::isApplicable` uses an inline list without `multipage_item` in the label**

`EntityReference.php:130` - the `isApplicable` whitelist includes `'multipage_item'` but the `getProcess` switch only has explicit handling for `taxonomy_term`, `user`, and the `community/media/node/protocol/multipage_item` group. It works correctly (falls through to the generic lookup), but a comment noting that `multipage_item` intentionally shares the generic path would help future readers.

---

### LOOKS GOOD

- `ProtocolAwareEntityContent` access control design is solid: user-1 bypass is appropriately scoped, access is checked against the pre-modification entity in `updateEntity()` to prevent protocol-escalation attacks.
- The `extractAndClearMediaAltUpdates` / `applyMediaEntityAltText` two-pass approach for setting image alt text on referenced media is clean and handles the edge case that entity_reference fields have no native `alt` sub-property.
- All 16 taxonomy import strategies follow a consistent YAML pattern and are covered by a single batched update hook (40026) rather than 16 individual hooks.
- `accessCheck(TRUE)` is correctly set on the entity query in `TaxonomyExportStartForm`.
- Permission grants in update hooks use `hasPermission()` guards before `grantPermission()` - safe to re-run.
- The migrate_source_csv patch addresses a real deprecation and is scoped correctly.
- Config role YAMLs match what the update hooks grant.

---

**Summary:** Two issues worth fixing before merge - the static service call in `ProtocolAwareEntityContent` (straightforward injection change) and the N+1 in `TaxonomyTermImportStatus`. The update hook gap and the `update_40029` overwrite behavior are worth a quick decision/comment. Everything else is in good shape.

michael-wynne-wsu:
All 6 issues addressed. Here's what was done:

1. **`ProtocolAwareEntityContent`** - Added `EntityTypeManagerInterface` to the `use` block, declared it as an injected property, added it to `__construct` and `create()`, and replaced `\Drupal::entityTypeManager()` with `$this->entityTypeManager`.

2. **`TaxonomyTermImportStatus`** - Added `preRender()` that batch-loads revision counts for all tids in a single grouped query, and updated `render()` to read from the `$revisionCounts` cache instead of firing a per-row query.

3. **`mukurtu_export.install`** - Renumbered `update_40009` to `update_40006` (next available after 40005).

4. **`mukurtu_import.install`** - Added a comment to `update_40029` explaining the intentional unconditional overwrite.

5. **`CsvEntityFieldExportEventSubscriber`** - Added a comment to `exportEntityReference` noting the per-value entity loading is intentional.

6. **`EntityReference.php`** - Fixed `isApplicable` to explicitly include `multipage_item` in the list (it was missing from the array) and added a comment explaining it shares the generic lookup path.

michael-wynne-wsu:
@nick-deer Export PR 1. This one is pretty isolated from the next ones.

nick-deer:
The patch isn't applying for me... here's the error:

In RemoteFilesystem.php line 105:
                                                                                                    
  [TypeError]                                                                                       
  Composer\Util\RemoteFilesystem::copy(): Argument # 1 ($originUrl) must be of type string, null given, called in /var/www/html/vendor/cweagans/composer-patches/src/Patches.php on line 388

michael-wynne-wsu:
@nick-deer funnily enough I was just working on that separately - this SHOULD sort out the patch and deprecation warnings https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1737 (and then come back to update this one).

michael-wynne-wsu:
@nick-deer okay now this has the message digest patch applied. I'm not seeing any build errors on my end, but let me know if you run into something new...

michael-wynne-wsu:

Link:
- https://pr1725-km6ehn1rpkirxzkw7dwkmx4wfrqj03ud.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a2b63dee73ea4de46541b12

nick-deer (approved):

PR #1737, opened 2026-06-16
Fix/remove obsolete message digest patches

michael-wynne-wsu:
Pre-merge checklist:


michael-wynne-wsu:
## PR Review: Fix/remove-obsolete-message-digest-patches

**Blast Radius:** LOW - composer.json patch config and patch files only; no Mukurtu PHP changed
**Security:** Clean - no secrets, no SQL, no XSS
**Tests:** N/A - no testable logic in the Mukurtu profile changed
**Breaking Changes:** 1 confirmed (see below)

---

### MUST FIX (Blocking)

**1. Two deleted patches contained fixes that are NOT in message_digest 1.5.0**

The exploration that concluded "all patches already applied in 1.5.0" checked the DDEV installation after the patches had been applied - not the upstream 1.5.0 source. The actual 1.5.0 `DigestInterval.php` is missing three things the deleted patches provided:

**a) Null guard before `$this->flag->actionAccess()`**
The current 1.5.0 `access()` method (no null check):
```php
public function access($object, AccountInterface $account = NULL, $return_as_object = FALSE) {
    $result = $this->flag->actionAccess('flag', $account, $object);  // fatal if $this->flag is null
    return $return_as_object ? $result : $result->isAllowed();
}
```
If `$this->flag` is null, this throws a fatal error. The deleted patch added:
```php
if (!$this->flag) {
    return $return_as_object ? AccessResult::forbidden() : FALSE;
}
```

**b) Plugin ID fallback in constructor**
The 1.5.0 constructor only sets `$this->flag` when `configuration['flag_id']` is present. When the action is loaded via plugin discovery (derivative ID format `message_digest_interval:email_node:daily`), that key may not be in configuration. The deleted patch extracted the flag ID from the derivative plugin ID as a fallback. Without it, `$this->flag` stays null and triggers the fatal above.

**c) Implicit nullable deprecation in `access()`**
`AccountInterface $account = NULL` is still in 1.5.0 (not `?AccountInterface`). The deleted `message_digest-nullable-param-deprecation.patch` fixed this line - and it's still needed.

**Fix needed:** Write updated versions of both patches with context lines that match the 1.5.0 constructor signature (which has more parameters than the version the old patches were written against). The patch application was failing because the context didn't match, not because the fixes were already present.

---

### LOOKS GOOD

- `message_digest-module-nullable-param-deprecation.patch` is correct: targets the right file (`message_digest.module` line 137), applies cleanly, silences the `PHP Deprecated` warning confirmed in `drush status`
- Patch format (git unified diff with `a/`/`b/` prefixes) is correct for `cweagans/composer-patches`
- Commit messages are clear and accurate
- The `DigestIntervalActionDeriver.php` changes from the deleted patch are genuinely NOT needed: 1.5.0 deliberately hardcoded the flag IDs with the comment "the flag service is not completely ready during installation" - so the `FlagServiceInterface` injection and existence check in the deriver can stay removed

---

### Summary of Required Changes

1. Create `patches/message_digest_ui-digest-interval-access-null-guard.patch` targeting `message_digest_ui/src/Plugin/Action/DigestInterval.php` in 1.5.0 - adding the `AccessResult` import, the `?AccountInterface` type hint, and the `if (!$this->flag)` guard before `actionAccess()`
2. Wire it up in `composer.json` under `drupal/message_digest`

The deriver patch can stay deleted - that logic changed in 1.5.0 intentionally.

michael-wynne-wsu:
Done. The branch now has three commits:

1. Remove two stale patches (already-in-1.5.0 claim was based on patched DDEV state, not upstream)
2. Add `message_digest-module-nullable-param-deprecation.patch` for `message_digest.module` line 137
3. Add `message_digest_ui-digest-interval-null-guard.patch` - the consolidated replacement for the two deleted patches, rewritten against the 1.5.0 constructor signature

The new `patches/message_digest_ui-digest-interval-null-guard.patch` applies cleanly (`patch -p1` confirmed), all three fixes are present in the installed file, and `drush status` shows a clean bootstrap with no deprecation warnings.

nick-deer:
Got the patch applying, but two deprecation warnings slipped through:

Deprecated: Drupal\message_digest\Plugin\Notifier\DigestBase::__construct(): Implicitly marking parameter $message as nullable is deprecated, the explicit nullable type must be used instead in /var/www/html/web/modules/contrib/message_digest/src/Plugin/Notifier/DigestBase.php on line 76

Deprecated: Drupal\message_digest\Plugin\Notifier\DigestBase::create(): Implicitly marking parameter $message as nullable is deprecated, the explicit nullable type must be used instead in /var/www/html/web/modules/contrib/message_digest/src/Plugin/Notifier/DigestBase.php on line 93

nick-deer (approved):

michael-wynne-wsu:
Thanks I'll see if I can address them before merging, shortly.

PR #1743, opened 2026-06-18
Add mukurtu_protocol_sync module for media-to-content protocol sync

### LINKED ISSUE #1730: Exploring content and media protocol sync
# Protocol Sync Between Media and Content - Options Exploration

## Context

In Mukurtu CMS, both media entities and content nodes independently require a `field_cultural_protocols` value (protocols + sharing setting). Media is added to content via `field_media_assets` (or `field_recording`, `field_thumbnail`, etc.) using the Drupal Media Library modal widget. Both entities must have protocols set, but in the vast majority of cases, a user wants the same protocols on the media as on the content it belongs to.

The current workflow forces users to select protocols twice: once on the content form, and once on the media upload form inside the media library modal. This is a known friction point.

There is already a `field_mukurtu_protocol_inherit` node-to-node inheritance field on some bundles (collection, dictionary_word, word_list), but no media-to-content inheritance exists.

---

## Key Architectural Facts

- `field_cultural_protocols` is a custom compound field (protocols list + sharing_setting) defined on both `MukurtuNode` and `MukurtuMedia` via `CulturalProtocolControlledTrait`
- Media access control is enforced via `MukurtuProtocolMediaAccessControlHandler` and the `mukurtu_protocol_access` table
- Media is added to content via `media_library_widget` - a separate AJAX modal; media is saved as its own entity (not inline)
- Media can be referenced by more than one content item
- Media can also be orphaned (not attached to any content)
- `field_mukurtu_protocol_inherit` (node-to-node) provides a precedent for the inheritance pattern

---

### Option A: Pre-populate on New Upload (Creation-time Copy, No Persistence)

**What it does:** When a user uploads new media within the media library modal while editing a content form, the content's current protocols are pre-populated as defaults into the media protocol field. The user can change the value before saving. After save, media and content are fully independent - no ongoing link.

**UX:** Minimal change to workflow. User sees protocol field already filled in and can accept or override.

**Implementation path:** Pass the parent content's protocol values as a query parameter into the media library modal URL. A `hook_form_alter` or custom media library opener sets default values on the upload form. No new database fields required, no changes to access control logic.

**Per-media opt-out:** User just edits the pre-populated value - no flag needed.

**Risks and challenges:**
- The media library modal runs as a separate AJAX request; passing parent form state into it is non-trivial. The parent form's protocol field value may not yet be saved (user is mid-edit)
- If a user adds media to content that has no protocols yet selected, there is nothing to pre-populate
- No protection against drift: if content protocols change later, media is not updated
- Selecting existing media (not uploading new) - no mechanism to warn about a protocol mismatch

**Verdict:** Low implementation complexity, low persistence value. Good as a convenience but not a solution to the drift/maintenance problem.

---

### Option B: "Inherit from content" Persistent Flag on Media (Dynamic at Access-Check Time)

**What it does:** Add a flag to media (`field_protocol_source`) with values: `"own"` (default) or `"inherit"`. When set to `"inherit"`, the media access handler dynamically looks up all content that references this media and evaluates access using those content items' protocols instead of the media's own protocol field.

**UX:** Media form has a toggle "Use own protocols / Inherit from content". If inherit is selected, the protocol field is hidden/disabled.

**Implementation path:** New field `field_protocol_source` on all media bundles. Modify `MukurtuProtocolMediaAccessControlHandler` to check the flag and, if set, query `node__field_media_assets` to find parent content, then resolve protocols from there. Modify `mukurtu_protocol_access` grant rebuild to handle inherited media.

**Per-media opt-out:** The flag itself. Default is `"own"` (opt-in to inherit).

**Risks and challenges:**
- **Multiple parents problem:** A media item attached to 5 content items with 5 different protocol sets - whose protocols apply? Most permissive (union)? Most restrictive (intersection)? This requires a policy decision
- **Orphaned media:** If `inherit` is set but media has no parent content, access is undefined - need a fallback (deny? warn?)
- **Performance:** Access check requires a query to find all referencing content, then resolve their protocol sets. The current `mukurtu_protocol_access` table is designed for efficient lookups and this bypasses it
- **Grant rebuild complexity:** The `mukurtu_protocol_access` table would need to be rebuilt for media whenever any referencing content's protocols change - requires `hook_entity_postsave` on nodes to trigger media grant rebuilds

**Verdict:** Semantically clean (no duplicate data) but operationally complex. The multiple-parent problem requires a policy decision. Best suited if a media item has at most one "canonical" parent.

---

### Option C: "Sync from content" Flag - Content Pushes Protocols to Media on Save

**What it does:** When content is saved, for each media item in `field_media_assets` that has a "sync" flag set, overwrite the media's `field_cultural_protocols` to match the content's protocols.

**Two sub-variants:**
- **C1:** Flag is on the **media entity** - "Keep protocols in sync with content" checkbox on media form
- **C2:** Flag is on the **content-media relationship** (per-reference, not per-entity) - a checkbox next to each media item in the content form

**Implementation path (C1):** New boolean field `field_sync_protocols` on media. `hook_entity_postsave` on node iterates `field_media_assets`, checks each media's sync flag, copies protocols and saves. C2 requires a custom compound field type wrapping the media reference - more complex but architecturally cleaner.

**Risks and challenges:**
- **Multi-parent conflict (C1):** If media is attached to two content items with different protocols and both have sync enabled, whichever saves last wins - a data integrity risk
- **C2 avoids this** by making sync a property of the relationship, not the media entity - but custom compound field types are significant build effort
- **No creation-time convenience standalone:** New media still needs protocols manually at upload time; sync only fires on content save
- **Cascade saves:** Saving a node triggers media saves and grant rebuilds - potentially slow for content with many media items

**Verdict:** C1 is pragmatic but has a meaningful multi-parent hazard. C2 is architecturally superior but is a significant build.

---

### Option D: Hybrid - Pre-populate at Creation + Optional Persistent Sync (Recommended Starting Point)

**What it does:** Combines Option A (creation-time convenience) with Option C1 (optional persistent sync), and adds a "canonical parent" concept to handle the multi-parent problem.

**Specific design:**
1. When a user opens the media upload form inside the content form modal, the content's protocol field value is pre-populated as the default
2. A new checkbox on the media form: "Keep protocols in sync with this content" (shown when media is being added from a content form context)
3. If checked at creation time, the media stores both the sync flag AND a reference to the "canonical parent" content item
4. On content save: if media has sync enabled AND this content is the canonical parent, overwrite media protocols
5. If media is later attached to a second content item, a notice appears: "This media's protocols are managed by [original content]"

**New fields on media:**
- `field_sync_protocols` (boolean)
- `field_protocol_parent` (entity_reference to node, optional, single value)

**Per-media opt-out:** Unchecking the flag (default unchecked). Admin can also force-clear from media edit form.

**Risks and challenges:**
- More complex than any single option alone
- Creation-time pre-population still has the "parent form state not yet available" AJAX challenge
- The "canonical parent" concept means the first content to reference media "claims" it - may surprise users
- If canonical parent content is deleted, orphaned media needs handling (clear sync flag? preserve last-synced protocols?)

**Verdict:** Most user-friendly combined approach. The canonical parent concept solves the multi-parent problem with a clear policy.

---

### Option E: Dedicated "Create media with these protocols" Flow

**What it does:** Instead of modifying the media library modal, add a dedicated action on the content form: "Create and attach media" that opens a simplified upload form where the protocol field is pre-set and read-only (with a note "Protocols inherited from content"). Separate from the standard "Browse media library" path.

**Risks and challenges:**
- Adding new media while keeping the content form open and injecting the reference back is architecturally complex - would likely require a custom opener plugin for the media library
- Users learn a parallel workflow alongside the standard Media Library UX

**Verdict:** High UX value for the "create new" case but high implementation complexity. Not recommended as a first approach.

---

## Summary Comparison

| | Creation convenience | Persistent sync | Multi-parent safety | Implementation effort |
|---|---|---|---|---|
| A: Pre-populate on upload | Yes (if AJAX challenge solved) | No | N/A | Medium |
| B: Dynamic inheritance | No | Yes | Risky (policy needed) | High |
| C1: Sync flag on media | No | Yes | Risky (last-save-wins) | Medium |
| C2: Sync flag on relationship | No | Yes | Safe | High |
| D: Hybrid A + C1 + canonical parent | Yes | Yes | Medium (canonical parent policy) | Medium-High |
| E: Dedicated creation flow | Yes (opinionated) | No | N/A | High |

---

## Open Questions Before Implementation

1. **Multi-parent policy:** When media is referenced by two content items with different protocols, what should happen? (Block sync? Use first/canonical parent? Union of protocols?)
2. **Default behavior:** Should the sync flag default to ON or OFF for new media created via the content form?
3. **Existing media:** Should there be a bulk tool to retroactively set sync flags on existing media where protocols already match?
4. **Scope:** Is creation-time convenience (Option A alone) sufficient for v1, or is persistent sync required?
5. **UI for mismatch warning:** When a user selects existing media whose protocols don't match the content, should they see a warning?

michael-wynne-wsu:
Resolves https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1730

## Summary

- [ ] When you create/edit content and select protocol/sharing settings BEFORE opening a media modal, and then create a new media asset in-modal...
   - [ ] A "Sync protocols with content." toggle is active. 
   - [ ] Under the toggle, the media will be showing the same selections as the content.
   - [ ] If you turn off the sync, you can set media protocols independently as before.
- [ ] For synced media, if the content's protocols change, so does the media.
- [ ] When creating media singly or in bulk NOT through the content modal, the sync option is still present.
   - [ ] You DO have to assign a protocol in these cases.
   - [ ] But if sync is enabled, once the media is added to content, the media protocols will sync. 
   - [ ] This is a little annoying, but I think it's probably acceptable for the time being at least...
- [ ] Covers all relevant fields in content:

Field | Used by content type(s)
-- | --
field_media_assets | Digital Heritage, Person, Place, Dictionary Word
field_recording | Dictionary Word
field_thumbnail | Dictionary Word
field_collection_image | Collection
field_word_list_image | Word List

## To do

- [x] Assess roundtrip functionality. Likely needs to add a sync setting, and I anticipate we'll need that same set protocol first, but sync later requirement.
- [x] Review conflict experience, eg: adding media to multiple content.
- [x] Review experience when synced content is deleted.

Introduces the mukurtu_protocol_sync sub-module, which allows media assets to optionally keep their cultural protocols in sync with a canonical parent content item. Includes creation-time pre-population in the media library modal, persistent sync on node save/update, canonical parent claiming, orphan handling on parent delete, and UI warnings for multi-parent conflicts.

Pre-merge checklist:


michael-wynne-wsu:
@joelle-birano-wsu @christa-kavanagh-wsu I finally have this ready for testing. It worked on my local...

joelle-birano-wsu (changes requested):
This is so awesome! Overall this is working really well.

I think we're supposed to be able to see the protocols even if sync is turned on? Right now the protocols disappear and you have to turn sync off to see them. It would be nice to see them without having to turn sync off. Maybe grey them out when sync is on?

I see this in the modal and the library:

Image

I also think the sync toggle should go under the protocols so it's clear that protocols need to be selected first. 

I tested these out and didn't notice any issues (but I kept things pretty basic). This is what happened:

**Assess roundtrip functionality**
- Importing protocol changes to content worked nicely - media stayed in sync.
- I was able to export the sync protocol field
- I created a DH item with image from scratch, including sync protocol field
- Imported DH and media asset (different protocols but sync turned on) and the media asset synced with the DH item.

 **Review conflict experience**
- Media respected protocols of original content. It displays if it's accessible, and editing the 2nd content form warns you that it's synced with other content.

Image

 **Review experience when synced content is deleted.**
- Content is deleted, media is automatically set to "unsync." It does not sync up with any other content that it appears in. If you want it to sync with something else, you need to go into that form, remove it, and re-add it.

michael-wynne-wsu:
@joelle-birano-wsu I disagree about the order of operations. In my planning, the whole point of syncing media to content protocols is that you don't have to touch the media protocols unless needed. So, my intended workflow is...

1. Start creating content > set content protocols
2. Open media library > create/select media assets
3. If sync is desired, nothing is needed except ensuring it's selected.
4. If not desired, set as needed
5. Save and move on

Exposing protocols before sync controls in media means users will first think they need to assign the protocols, which they don't, if they're syncing.

Does that make more sense?


Everything else seems as intended for me...

joelle-birano-wsu:
Yeah that makes sense for the modal. I was getting confused in the media library since protocols need to be assigned but the sync toggle comes first. Maybe add helper text that explains that you need to select protocols but when sync is on, adding the media asset to content will change its protocols to match the content.

michael-wynne-wsu:
Gotcha, yeah that makes more sense then. Inheriting protocols really does only work in the content > media workflow. I didn't want to try and allow creation of media with sync allowed but no pre-set protocols since that's risky. So media created independently in-UI or in batch import will always need a protocol set, even if syncing is intended.

I can see if we can have different behaviour depending on page context, but that might get messy...

michael-wynne-wsu:
Not quite right, wanna help me edit?

> Sync protocols with content.
> When enabled, this media asset's protocols will always match those of it's parent content to keep content and media access in sync. Media created within a content form will automatically inherit content protocols; media created elsewhere requires protocol selection. Disable sync to set media protocols independent of parent content.

christa-kavanagh-wsu:
> Not quite right, wanna help me edit?
> 
> > Sync protocols with content.
> > When enabled, this media asset's protocols will always match those of it's parent content to keep content and media access in sync. Media created within a content form will automatically inherit content protocols; media created elsewhere requires protocol selection. Disable sync to set media protocols independent of parent content.

The middle sentence is important for documentation and elsewhere, but seems like just too much information here. Does this work? 

"Optionally sync protocols with content.
When enabled, the media asset's protocols are automatically synced with the parent content to help streamline workflows and keep content and media access in sync.
Disable sync to set media asset protocols independent of parent content."

michael-wynne-wsu:
I think that middle sentence is exactly what Joelle was asking about though.

christa-kavanagh-wsu:
Maybe add that sentence last? Because I feel like helper text should tell you what to do first, then maybe why you would do it?

christa-kavanagh-wsu (approved):
This seems to work as expected! The only thing I had trouble wrapping my head around was that I needed to add cultural protocols to the media import spreadsheet, which would then get overwritten by the sync settings.

michael-wynne-wsu:
Yeah, it's really intended to improve user experience in-page. As noted, I don't want to allow media to be created without protocols, and the import workflow is media > content (because content needs to reference existing media)

christa-kavanagh-wsu:
Makes sense, I just need to add a note about it in the roundtrip documentation so it doesn't trip other people up too.

joelle-birano-wsu:
Ha. I made it a bit longer

Sync protocols with content.
When enabled, this media asset's protocols will always match those of it's parent content to keep content and media access in sync. Media created within a content form will automatically inherit content protocols; media created elsewhere require protocol selection, and will update once added to content. Disable sync to set media protocols independent of parent content.

michael-wynne-wsu:
Okay so other than updating helper text, y'all are both happy with the functionality?

christa-kavanagh-wsu:
Yes, the functionality is great!

joelle-birano-wsu:
Yep! Love the functionality!

michael-wynne-wsu:
## Accessibility Review (WCAG 2.2 AA) — protocol-sync.js + mukurtu_protocol_sync.module

High
1. Capture-phase click interception likely loses the race to Drupal core's AJAX handler. protocol-sync.js:112-130 assumes addEventListener(..., true) guarantees it runs before Drupal's own mousedown/keypress-driven AJAX handler on the same .js-media-library-open-button. Per DOM spec, capture only orders ancestor listeners ahead of target-phase ones — for two listeners on the same element, registration order wins. mukurtu_protocol_sync.libraries.yml doesn't depend on core/drupal.ajax, so there's no guarantee of attach order. If core's handler wins, the media dialog can open even when protocols aren't selected (bypasses the gate for mouse and keyboard users via keypressResponse()'s synthetic mousedown). Needs a real-browser check; if confirmed, move the guard to an ancestor's capture listener instead of the button itself.

Medium
2. Multiple blocked "Add media" buttons across fields can leave stale role="alert" messages and duplicate/misfiring Escape-dismiss handlers (protocol-sync.js:135-178) if a keyboard user tabs between blocked buttons without an intervening click/Escape.
3. aria-live="polite" on the sync message (protocol-sync.js:50-55) relies on toggling style.display, not content mutation — an unreliable announcement pattern on some AT/browser combos (notably Safari/VoiceOver). Worth a real screen-reader spot check.
4. aria-label="Warning" paired with real text content (both server-rendered containers and the JS alert) risks the accessible name replacing the actual message on some AT, so only "Warning" gets announced instead of the full text.

Low / informational — static role="alert"/role="status" on full-page load won't auto-announce on many AT (pre-existing, not from this diff); redundant duplicate messaging on already-synced media forms.

Confirmed sound: label/checkbox association survives the DOM move, display:none correctly removes fields from the a11y tree, and keyboard click activation does work correctly if finding #1 doesn't occur.

## PR Review — full branch diff (main...HEAD)

Should Fix
1. Stale TempStore protocol data can leak into unrelated media pre-population (mukurtu_protocol_sync.module, mukurtu_protocol_sync_media_library_open_ajax). The AJAX wrapper only writes tempstore when the submitted protocol/sharing values are non-empty (if (!empty($protocol_ids) && !empty($sharing))). If a user selects protocols, opens the media dialog (storing them), closes it, then clears their protocol selection and reopens the dialog, the stale tempstore entry is never overwritten or cleared — the next media item gets pre-populated with protocols the user has since removed.
2. Unbounded save cascade on node insert/update (_mukurtu_protocol_sync_sync_media_from_node). Runs on every node with field_cultural_protocols, iterating 5 media fields and calling ->save() on every sync-enabled referenced media item, with no batching. For nodes with large media libraries, or bulk operations (migrations, VBO, revision bulk-edit), this means N synchronous entity saves per node save — worth a scale check against real content.

Suggestions
3. Confirm the interaction between finding #1 in the a11y review (AJAX race) and the intended "protocols required" gate — if core wins the race, this is a functional bug, not just an a11y one.
4. Minor style: strpos($form_id, 'media_library_add_form_') !== 0 could use str_starts_with() (PHP 8+, matches the rest of the file's typed-declaration style) — non-blocking nit.

Checked, no issue found (matches existing codebase conventions):
- Procedural hook_entity_insert/update/delete, hook_form_alter, hook_entity_field_storage_info/hook_entity_bundle_field_info in a .module file — sibling modules (mukurtu_protocol, mukurtu_collection, mukurtu_dictionary, etc.) do the same; this isn't a deviation from the OOP #[Hook] convention, which here is reserved for other hook types.
- accessCheck(FALSE) in mukurtu_protocol_sync_entity_delete's entityQuery — idiomatic for this codebase's system-level cleanup queries.
- Declaring fields via hook_entity_field_storage_info/hook_entity_bundle_field_info and installing them via entityDefinitionUpdateManager in hook_install() — this mirrors mukurtu_protocol.install:658's own pattern (hooks declare the field for discovery; update/install manager creates the actual schema).
- No hook_uninstall() cleaning up the fields — no sibling module does this either, so it's consistent, not a gap.

michael-wynne-wsu:
Both files pass linting. Here's a summary of what changed:

protocol-sync.js
1. Moved the "protocols required" guard from listeners on the button itself to capture-phase listeners on the ancestor nodeForm — this is now guaranteed to run before Drupal core's own AJAX listener on the button, since same-element listeners fire in registration order regardless of capture, but ancestor capture listeners always run first.
2. Added an activeProtocolMessage tracker so a new warning always removes the previous one (and its dismiss listeners) before showing, regardless of which field's button triggered it — fixes stacking/stale-listener issue.
3. The sync-status message now mutates textContent (not just style.display) when toggling, so assistive tech sees a real content change and reliably announces it.
4. Removed aria-label="Warning" from the JS-created alert so its own text content is what gets announced, instead of being overridden by the label.

mukurtu_protocol_sync.module
5. mukurtu_protocol_sync_media_library_open_ajax now clears the TempStore entry when the current selection is empty, instead of leaving stale data in place — fixes the leak where a cleared protocol selection could still pre-populate later media.
6. _mukurtu_protocol_sync_sync_media_from_node now skips the ->save() call when protocols already match, avoiding an unnecessary media revision on every node save that doesn't touch protocols (the main contributor to the flagged save-cascade concern).
7. strpos(...) !== 0 replaced with str_starts_with(...).
8. Removed aria-label from both server-rendered role="alert"/role="status" containers, matching the same fix as #4 (their visually-hidden <h2> + message text is read directly instead).

michael-wynne-wsu:
Real browser test results

Ran this in DDEV (installed the module fresh, seeded a test community/protocol, drove it with Playwright), since the previous review was code-only.

Found and fixed a real bug the code review missed: keyboard activation (Enter/Space) completely bypassed the protocol gate. Drupal core's ajax.js calls its AJAX handler directly from its own keypress listener rather than dispatching a click/mousedown event — our guard only listened for those two, so a keyboard user pressing Enter on the "Add media" button opened the media library even with no protocols selected. Fixed by adding a keypress listener to the same ancestor-capture guard, filtering to Enter/Space (modules/mukurtu_protocol_sync/js/protocol-sync.js).

All 4 tests pass after that fix:

1. Blocked without protocols (mouse) — PASS: No modal, no AJAX request fired, warning shown with role="alert" and received focus
2. Unblocked after selecting protocols — PASS: Modal opens normally (mouse and keyboard)
3. Keyboard access (Tab + Enter) — PASS: Blocked identically to mouse, message shown and focused
4. Dismiss behavior — PASS: Escape removes message and returns focus to the button; click-outside also dismisses cleanly

No console or page errors in any run. I also spot-checked the media-edit-form sync toggle (the other JS path we touched): the checkbox correctly relocates into the protocols fieldset, and the aria-live message's textContent now populates/clears correctly on toggle (confirmed via DOM inspection, screenshot).

One methodology note: my first pass showed false failures — my own test harness used offsetParent !== null to check dialog visibility, which is unreliable for position: fixed elements (jQuery UI dialogs use it). Switched to getBoundingClientRect() and the false failures disappeared. Also had to isolate each test in its own browser context, since Chromium restores form field state across goto() calls to the same URL within one context, which was making a later test start with a stale, pre-checked checkbox.

nick-deer (approved):

PR #1744, opened 2026-06-18
Chore/drupal core security update

michael-wynne-wsu:
Drupal core and tagify security updates


michael-wynne-wsu:
@nick-deer okay I've got this all running cleanly now.

nick-deer (approved):

PR #1748, opened 2026-06-19
Remove drupal/message_digest and message_digest_ui modules

michael-wynne-wsu:
## Summary

- Removes \`drupal/message_digest\` and \`drupal/message_digest_ui\` from composer requirements and deletes their three PHP 8.4 compatibility patches
- Strips all config references to message_digest (field definitions, display configs, role dependency, info.yml dependencies)
- Adds \`mukurtu_notifications_update_40005()\` to unregister the modules from core.extension and delete their field storage from existing installations

## Deployment note for sites with message_digest currently installed

The update hook writes directly to \`config.storage\` (bypassing \`configFactory()->save()\`) so it does **not** require the module code to be present on disk when it runs. However, \`drush updb\` still needs to bootstrap Drupal first -- and bootstrap may fail if message_digest is still listed in \`core.extension\` but missing from vendor.

Safe upgrade order:

1. Pull this branch
2. \`drush updb\` (bootstraps while message_digest code is still on disk, then update 40005 removes it from core.extension)
3. \`composer update\` (removes the packages)
4. \`drush cr\`

If \`composer update\` was run first and Drupal can no longer bootstrap, manually remove message_digest from core.extension via \`drush sql-query\` before re-running \`drush updb\`.

## Test plan

- [x] Fresh install: verify message_digest is not installed, no field storage errors
- [x] Existing install (correct order): run \`drush updb\`, confirm update 40005 runs cleanly, then \`composer update\`, then \`drush cr\` -- no errors
- [x] \`drush pm:list | grep message_digest\` returns nothing after upgrade
- [x] User edit form has no message digest interval field
- [x] Admin user bulk-action form has no message_digest_interval actions

## Pre-merge checklist:


michael-wynne-wsu:
Here's a summary of what was completed:

Committed to feature/remove-message-digest (commit ac46d20c):

- composer.json -- removed drupal/message_digest and drupal/message_digest_ui from require and their 3 patch entries
- 3 patch files deleted
- mukurtu.info.yml + mukurtu_notifications.info.yml -- removed the dependency declaration
- 3 field config files deleted (storage and instance for user + flagging entities)
- core.entity_form_display and core.entity_view_display YAMLs -- removed hidden message_digest components
- user.role.mukurtu_manager.yml -- removed from module list
- FormHooks.php -- removed the 3 message_digest action IDs from the bulk action filter (the method itself stays; it still  serves other purposes)
- mukurtu_notifications.install -- added update_40005() which removes the modules from core.extension, calls module_installer->uninstall() if code is present, and deletes the field storage entities

michael-wynne-wsu:
## PR Review: Remove drupal/message_digest and message_digest_ui modules (#1748)

**Blast Radius:** MEDIUM -- removes a composer dependency and DB field storage across two entity types (user, flagging). Affects all users on upgrade path. Fresh installs also affected (hook_install() changed).

**Security:** Clean -- no hardcoded secrets, no injection vectors, no auth bypasses introduced.

**Tests:** 0 test files changed (none existed for this code path; acceptable given the procedural nature of the removal).

**Breaking Changes:** Intentional -- field storage deleted, permissions removed, modules unregistered.

---

### MUST FIX (Blocking)

None remaining. The `mukurtu.install` fix (removing the explicit `install(['message_digest_ui'])` call) that just landed was the last blocker.

---

### SHOULD FIX (Non-blocking)

**1. Deployment note is misleading about WHY the order matters**

The PR description says `drush updb` must run "while the module code is still in vendor." That was true when the hook used `module_installer->uninstall()`. Now that the hook uses `config.storage->write()`, the hook itself doesn't care whether the code is on disk. The order is still required, but for a different reason: Drupal must be able to bootstrap to run `drush updb`, and bootstrap may fail if message_digest is in `core.extension` but missing from vendor. The explanation should be updated so deployers don't accidentally think this is a hook-level restriction.

**2. `administer message digest` permission in `mukurtu.install`**

Both occurrences are already commented out with `#`, which is fine. But the stale comments could confuse a future reader who doesn't know why they exist. Consider deleting the lines entirely rather than leaving them as dead comments.

---

### SUGGESTIONS

**3. `formAlterRemoveNotificationBulkActions` docblock**

The updated docstring says "Removes unwanted actions from the user admin bulk form and OG membership form." This is accurate -- the function still strips OG membership and block-user actions. No code change needed, but worth noting the docblock change is an improvement.

**4. Field deletion ordering in update hook**

The hook deletes `field_storage_config` for both `user.message_digest` and `flagging.message_digest`. Drupal's `FieldStorageConfig::delete()` internally removes field instances first, then the storage -- so this is safe. The `if ($storage)` guard correctly handles fresh installs where the storage was never created.

---

### LOOKS GOOD

- `config.storage->write()` instead of `configFactory()->save()` is the correct approach for unregistering modules whose code no longer exists -- avoids the `ConfigCrudEvent::SAVE` subscriber chain that triggers `ModuleInstaller::install()`.
- `keyValue('system.schema')->deleteMultiple()` correctly cleans up the orphaned schema version entries that bypass `module_installer->uninstall()` would leave behind.
- All three config/install YAML files for message_digest field storage properly deleted.
- All three composer patches properly deleted alongside the composer.json entries.
- `$changed` flag guard prevents an unnecessary `config.storage->write()` on fresh installs -- idempotent.
- `mukurtu.install` hook_install() fix resolves the Tugboat fresh-install failure that was the root cause of the "build error."

---

The PR is ready to merge pending a CI pass on the latest commit. The pre-existing PHPUnit failures (`og` module / image plugin) and Playwright flakiness both appear on `main` too -- not regressions from this PR.

michael-wynne-wsu:
Done:

- **Item 1** -- PR description deployment note updated: explains the order is needed so Drupal can bootstrap, not because the hook requires module code on disk.
- **Item 2** -- Both `#'administer message digest',` dead comments removed from `mukurtu.install` and committed.
- **Items 3 & 4** -- Already correct, no changes needed.

michael-wynne-wsu:
@nick-deer so all those failing patches were for a module that as far as I can tell we don't actually use (message_digest). So I've tried stripping it out of the codebase here... Let me know what you think.

nick-deer (approved):

michael-wynne-wsu:
Ugh, this might come up in https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1752, let's not do this one yet...

michael-wynne-wsu:
@nick-deer I've implemented an alternate message system (being tested here https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1760) so YES we can remove this module that has been a problem.

nick-deer (approved):

PR #1749, opened 2026-06-19
Collapse Gin admin toolbar and sidebar by default for new users

michael-wynne-wsu:
Adds early-loading JS to mukurtu_gin_custom that sets the Gin localStorage defaults to collapsed before gin_init reads them. Only applies when no user preference has been stored, so existing user choices are preserved.

Pre-merge checklist:


michael-wynne-wsu:
## PR Review: Collapse Gin Admin Toolbar and Sidebar by Default

**Blast Radius:** LOW - client-side localStorage only, no PHP APIs, config, or schema touched
**Security:** Clean - no user input, no DOM manipulation, no server-side implications
**Tests:** N/A - pure localStorage initialization; no Drupal JS test infrastructure in this codebase
**Breaking Changes:** None - `!localStorage.getItem()` guard preserves existing user preferences

---

### LOOKS GOOD

- Guard logic is correct: only sets defaults when the key is absent, so existing user preferences survive deployment
- Setting both `Drupal.gin.toolbarExpanded` and `Drupal.navigation.sidebarExpanded` is good defensive coverage - handles whichever toolbar variant is active
- `weight: -100` correctly ensures this runs before Gin's `gin_init` library (`weight: -99`), preventing any layout flash
- `header: true` + `preprocess: false` is the right pattern to match Gin's own init script

---

### SUGGESTIONS (non-blocking)

**1. Library loads for anonymous users**
The library is now attached outside `isAdminRoute()`, which means it runs for every visitor - including anonymous users who never see the admin toolbar. The overhead is negligible (12 lines of localStorage reads), but a tighter scope would be to gate it on the user having an active session or the admin theme being active:

```php
// Only attach when a user is logged in
if (\Drupal::currentUser()->isAuthenticated()) {
  $attachments['#attached']['library'][] = 'mukurtu_gin_custom/gin-collapsed-defaults';
}
```

**2. Fragile dependency on Gin internal key names**
`Drupal.navigation.sidebarExpanded` and `Drupal.gin.toolbarExpanded` are undocumented Gin internals. If Gin renames them in a future release, the code will silently stop working (new users will see the expanded state again). Worth a comment in the JS noting the key source:

```js
// Keys used by gin/js/navigation/navigation.js and gin/js/toolbar.js respectively.
```

---

**Verdict: Ready to merge** once you decide whether to apply suggestion #1. Suggestion #2 is a minor documentation note. No blocking issues.

michael-wynne-wsu:
Done and pushed. Both fixes are in - the library now only loads for authenticated users, and the JS has a comment pointing back to the Gin source files for those localStorage keys.

michael-wynne-wsu:
@nick-deer this is just a small UI change, setting some admin/gin menus to default collapsed instead of expanded to clean up user windows.

nick-deer (approved):

PR #1742, opened 2026-06-17
Feature/editorial workflow

michael-wynne-wsu:
## Summary

- [ ] Adds an optional editorial workflow to extend the [default content workflow](https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1738) (not discussed more here, but that functionality should all be in here).
- [ ] Enabled at `/admin/config/mukurtu/workflow-settings`, linked on Dashboard as "publishing workflows".
- [ ] Users can duplicate and create new workflows. I have prevented them from editing our core workflows to avoid breaking things.
- [ ] Content now has these states:
   - [ ] Draft
   - [ ] Awaiting review
   - [ ] Needs revision
   - [ ] Published 
   - [ ] Draft
- [ ] When active:
   - [ ] All content creators except protocol stewards can only save as draft/awaiting review.
   - [ ] Protocol stewards and content creators can leave review notes and send back and forth in different states.
   - [ ] Only protocol stewards can publish, edit published content, and move content in and out of published/archived states.
- [ ] There is a status banner on all content for content that is not Published.
- [ ] Status banner includes the expandable review notes and review actions.
- [ ] New "my content" and "review queue" links on dashboard (my account).

## To do

- [ ] De-link content types at `/my-content` and `/review-queue`
- [ ] Remove non-steward access to `/review-queue`
- [ ] Expand scope so that language stewards have same review capacity over words and word lists.
- [ ] Assess multi-protocol content access and process.

## Pre-merge checklist:

- [ ] I have checked for and if required, created update hooks.
- [ ] I have run an accessibility check, and resolved any issues.
- [ ] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [ ] I have updated from `main` and resolved any merge conflicts.
- [ ] I have verified that all expected checks pass.
- [ ] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1752, opened 2026-06-23
Notifications phase 1: re-enable hooks, add /notifications page (#270)

### LINKED ISSUE #270: Events and notifications
- User-configurable notifications for set events.
- Should be able to control email notifications (on and off, ideally by type), and digest/frequency.
- Should have an in-dashboard (?) feed.
- Different events for different roles

Events to include:
- publishing workflow notifications
- comment notifications
- edits to published content
- being added to groups
- group membership requests

michael-wynne-wsu:
## Summary

- Initial implementation of a notifications center. This is going to be expanded later, but I want to keep code reviews manageable.
- `/notifications` is linked in "my account" section of dashboard. Interface is intentionally sparse now.
- Collects an assortment of basic notifications, to be expanded later.

## Code notes

Re-activates the disabled notification hooks in mukurtu_notifications, deploys missing field attachment configs that prevented hook execution, and adds a /notifications page showing the current user's activity log.

Key changes:
- Re-enable hook_node_insert, hook_node_update, hook_node_delete, hook_user_insert; remove dead early returns
- Fix broken API calls: replace nonexistent mukurtu_protocol.protocol_manager service with current $node->hasField/getProtocols() API; remove undefined protocol constants
- hook_node_update notifies node owner + protocol stewards when a different user edits published content; uses User::loadMultiple to avoid N+1 queries
- hook_node_delete now cleans up both insert and update message records
- Add /notifications Views page display at /notifications path with uid contextual filter (current user default) and user validator to prevent IDOR (reject requests for other users' feeds with 403)
- Add 13 missing field attachment config/install YAML files
- Update hook 40005: deploy views config
- Update hook 40006: deploy field attachment configs
- Update hook 40007: create field_protocol storage+attachment via Field API (raw config writes skip DB schema creation; FieldStorageConfig::create() handles this correctly)
- Update hook 40008: fix "Details" link text to "View batch import report" for WCAG 2.4.4 compliance
- Stub mukurtu_notifications_notify_protocol/community for Phase 2

Pre-merge checklist:


christa-kavanagh-wsu (approved):
This works as expected.

michael-wynne-wsu:
## PR Review: Notifications Phase 2 - `feature/270-notifications-phase-2`

**Blast Radius:** MEDIUM -- changes are self-contained within `mukurtu_notifications`. The `FormHooks.php` change touches the `CommunityManagerUserCreationForm` notify flow. No shared library or API contract changes.

**Security:** No findings. IDOR is mitigated by `current_user` contextual filter on the notifications view.

**Tests:** No test files added or removed. No automated tests for notification hooks exist in this module (pre-existing gap, not introduced here).

**Breaking Changes:** None detected. Update hooks handle all existing-site migrations.

---

### SHOULD FIX (Non-blocking)

**1. `getProtocols()` call in `hook_node_update` is unverified**

```php
$protocol_ids = $node->getProtocols();
```

This is called on a raw `Node` object. `getProtocols()` is not a standard Drupal core method -- it must come from a Mukurtu trait or base class. If the node type doesn't include that trait, this will fatal. Should add `method_exists` guard or type-hint to the appropriate base class.

**2. Seven update hooks all write the same view config**

Hooks `40005`, `40009`, `40013`, `40014`, `40015`, `40016`, `40017` all do the identical config write:

```php
$config_storage->write('views.view.mukurtu_message_log', $source->read('views.view.mukurtu_message_log'));
```

This is correct and harmless -- each fixed an in-flight bug during iterative development. But the docblock for `40017` says "Add accessible caption (WCAG 1.3.1)" when the caption was already present in the config being written by 40016 and earlier. The docblocks diverge from reality. Consider updating 40017's comment to read: "Re-deploy view to ensure caption and defaults block are active on all environments."

**3. `mukurtu_new_user_registration` is in view dependencies but not the filter**

The master display depends on `mukurtu_new_user_registration` but the page display filter excludes it. The intent appears to be that admin-facing registration messages should not appear on the user-facing `/notifications` page -- this is correct. But the absence is easy to misread as an oversight. A YAML comment on the filter block would make the intent clear:

```yaml
# mukurtu_new_user_registration is intentionally excluded: it is an
# admin-facing message, not a user-facing activity feed entry.
```

---

### SUGGESTIONS

**4. `field_item` on collection messages is cardinality 1, set with an array**

```php
$message->set('field_item', array_values($newItems));
```

`field.field.message.mukurtu_new_item_in_collection.field_item` has no explicit cardinality setting, which defaults to 1. Setting an array stores only the first item. This was the same behavior before this PR (`'field_item' => $newItems` in `create()`), so it is not a regression -- but it means only one new item is recorded per collection-update notification. Worth a comment or a future ticket if multi-item is intended.

**5. `notify_protocol` and `notify_community` stubs have no tracking marker**

```php
function mukurtu_notifications_notify_protocol(Node $node, $subscribers): void {
  // Phase 2 implementation needed...
}
```

These functions are called nowhere in the active code (the call sites are commented out), so leaving them as stubs is fine. But adding a `@todo` with a ticket reference would prevent them from being silently forgotten.

---

### LOOKS GOOD

- `aria-live` correctly moved to the stable `<details>` parent in both `FormHooks.php` and `CommunityManagerUserCreationForm.php` -- this was the right fix and the inline comment explains the rationale clearly.
- `email.validator` service used consistently across all `sendMessage` call sites -- good defensive improvement.
- `notify message owner => FALSE` correctly prevents the registering/created user from receiving their own admin-notification email.
- All `sendMessage` calls are now wrapped in `try/catch` with `\Drupal::logger()` warnings -- this keeps entity saves from failing due to notification errors.
- Role-based access (`authenticated`) on the view instead of `perm: access content` -- correct for a user-specific page.
- `field_protocol` created via `FieldStorageConfig` / `FieldConfig` API (not raw config write) so the database schema is properly created. The pattern used in `hook_install` and `hook_update_40007` is correct.
- `drush updb -y` added to Tugboat's update step -- this ensures notification update hooks run in preview environments without manual intervention.
- Dead code removed cleanly -- the commented-out blocks in the old module were large and confusing. The stubs with phase 2 markers are much cleaner.
- IDOR risk on `/notifications` is addressed: page requires `authenticated` role and the `uid` contextual filter defaults to `current_user`.

---

**Items to address before merging:** items 1 (verify `getProtocols()` existence) and 2 (fix the 40017 docblock) are the most important. Items 3-5 are low-stakes but worth a note.

michael-wynne-wsu:
All 5 fixes applied. Here's a summary of what changed:

1. **`getProtocols()` guard** -- Added `use Drupal\mukurtu_protocol\CulturalProtocolControlledInterface;` and changed the check from `$node->hasField('field_cultural_protocols')` to `$node instanceof CulturalProtocolControlledInterface` before calling `$node->getProtocols()`. This prevents a fatal if a node has the field but doesn't implement the interface.

2. **`40017` docblock** -- Replaced the misleading "Add accessible caption (WCAG 1.3.1)" with an accurate description: the caption was already in 40016's config, and 40017 ensures it lands on environments where 40016 ran before the caption was committed. Also fixed the cross-reference in 40016's docblock.

3. **YAML comment** -- Added a comment in the page display filter block of `views.view.mukurtu_message_log.yml` explaining that `mukurtu_new_user_registration` is intentionally excluded (admin-facing, not for the user activity feed).

4. **`field_item` cardinality comment** -- Added a comment in `mukurtu_notifications_notify_collection()` noting that `field_item` is cardinality 1 and only the first new item is stored, with a reference to issue #270 if multi-item recording is ever needed.

5. **`@todo` markers** -- Added `@todo Implement in issue #270...` to both `mukurtu_notifications_notify_protocol()` and `mukurtu_notifications_notify_community()` with enough context to act on them later.

michael-wynne-wsu:
Passing this on for review so we can expand it more with https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1760

I do think we need to merge https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1742 BEFORE this one.

joelle-birano-wsu (approved):
Looks good!

nick-deer (approved):

PR #1753, opened 2026-06-23
Add mukurtu_setup module with site setup checklist (issue #366)

### LINKED ISSUE #366: 3Cs setup workflow - add categories?
We have a semi-guided community/protocol workflow. We should include categories in some way.

michael-wynne-wsu:
Adds a new mukurtu_setup module providing a setup checklist for administrators and Mukurtu Managers on new Mukurtu sites. 

Pre-merge checklist:


michael-wynne-wsu:
@joelle-birano-wsu @christa-kavanagh-wsu I know I need to do a pass on the text (feel free to make suggestions though). But is there anything else you think is missing or should work differently?

The system can automatically mark some tasks as complete but not others, which is why only some have "mark as done".

Screenshot 2026-06-24 at 10 50 38 AM

Screenshot 2026-06-24 at 10 50 29 AM

christa-kavanagh-wsu (commented):
I like that it can automatically check off at least some of the tasks. 

Is there a way to redirect back to the checklist when you navigate away from it to complete a task?

michael-wynne-wsu:
Mmmm, maybe. But some things, like community > protocol or landing page are more involved and multi-step, so I don't know if that will play nice. I'll look but I think it's low priority.

michael-wynne-wsu:
Yeah exactly what I thought:

- The front_page task links to / -- not a form, so no redirect applies
- The navigation_menu task (/admin/structure/menu/manage/main) is a menu listing page, not an add form -- redirecting back would depend on what the user does on that page
- The site_footer task links to /admin/content/block-content (a listing), same situation

So is it worth adding redirects on links where possible and having an inconsistent experience?

christa-kavanagh-wsu:
Eh, probably not. Some kind of flag or other visual cue to get folks back to the checklist would be nice, but not necessary.

michael-wynne-wsu:
Alright we'll leave it for now, see if users ask for something more structured in the future.

joelle-birano-wsu (commented):
I really like this! 

- RE: the redirect, is it possible to open the links in a separate window? 

- Is it possible to add "mark as done" to "Set Your Site Name and Email"?

michael-wynne-wsu:
@joelle-birano-wsu 

> RE: the redirect, is it possible to open the links in a separate window?

Yes, but tasks won't show as complete until the checklist page is refreshed, so there is still a manual action needed. Happy to do that though if you want it.

> Is it possible to add "mark as done" to "Set Your Site Name and Email"?

Is that necessary? It marked as complete when I changed it from the default tugboat values.

joelle-birano-wsu:
> Yes, but tasks won't show as complete until the checklist page is refreshed, so there is still a manual action needed. Happy to do that though if you want it.

Ahhh right. Umm...maybe it's fine as is. It easy enough to get back to. 
> 
> > Is it possible to add "mark as done" to "Set Your Site Name and Email"?
> 
> Is that necessary? It marked as complete when I changed it from the default tugboat values.

Oh really? It wasn't working for me. Nothing was happening when I changed stuff.

michael-wynne-wsu:
> Oh really? It wasn't working for me. Nothing was happening when I changed stuff.

When i opened this tugboat build it was still showing the default values... Were you testing it here?

joelle-birano-wsu:
> > Oh really? It wasn't working for me. Nothing was happening when I changed stuff.
> 
> When i opened this tugboat build it was still showing the default values... Were you testing it here?

I see what you mean...maybe it was cache thing, because when I was testing earlier, it was not marking as complete. If you want to rebuilt it I can test again.

michael-wynne-wsu:
Rebuilding now, give it a go

joelle-birano-wsu:
> * https://pr1753-xb56xkrakz0d3tkkutdrccfm23ueavn3.tugboatqa.com

Ok it didn't work the first time, but then I did it again and it worked. Probably caching? I think it's ok....

michael-wynne-wsu:
Okay, we'll keep an eye on it just in case.

So for now the only action items I'm seeing is to update the text on this page... Chime in if anyone wants other tasks.

michael-wynne-wsu:
## Accessibility Audit Results

**Required fix before merge:**

**Issue: Ambiguous action link text** (WCAG 2.4.6 / 4.1.2 - Critical)

The "Mark as done", "Undo", "Dismiss", and "Restore" links in mukurtu-setup-checklist.html.twig have no context for screen readers. A user tabbing through links would hear "Dismiss... Dismiss... Dismiss..." with no way to know which task each refers to.

Fix requires adding aria-label attributes with the task name. Proposed strings needing your approval:
- Dismiss @label (e.g. "Dismiss Create a community")
- Mark @label as done
- Undo @label
- Restore @label

Do you approve these aria-label strings?

**Other findings (lower severity):**

- **Moderate:** Completion state is indicated by color only (green box vs gray border). Color-blind users may not distinguish complete from incomplete. Consider using color: #767676 for the incomplete checkmark so the glyph is always visible -- a gray check for incomplete, white check on green for complete.
- **Low:** Verify a h1 page title is rendered by Drupal's page template above the h2 group headings.
- **Low (optional):** The li for complete tasks has no programmatic completion announcement. Adding a visually-hidden (Complete) span would make this explicit for screen readers.

---

## PR Review

**Blast Radius:** LOW-MEDIUM -- new module with no external dependents yet; touches mukurtu_protocol media access and node revision access (both existing systems).

**Security:** PASS -- CSRF tokens on all state-changing routes, access restricted to administrator/mukurtu_manager, SQL in mukurtu_protocol.module uses named placeholders, Markup::create() uses only hardcoded doc URLs.

**Tests:** None added (this module has no test suite yet -- acceptable for a new admin-only UI feature at this stage).

---

### MUST FIX

**1. Ambiguous aria-label on action links** (WCAG 2.4.6 -- see accessibility section above)
Awaiting your approval of the string wording before I apply this fix.

**2. Today's text changes are uncommitted**
The diff against main still shows the old task strings and the old string $description type in SiteSetupTask.php. The edits I made this session need to be committed before the PR is opened.

---

### SHOULD FIX

**3. SiteSetupChecklistBlock missing block_content_list cache tag**
SiteSetupChecklistBlock.php caches with community_list, protocol_list, and taxonomy_term_list but not block_content_list. The full page controller includes it. The block's task count won't invalidate when a footer block_content entity is updated.

---

### SUGGESTIONS

**4. isSiteNameSet() heuristic**
SiteSetupTaskManager.php marks the site name/email task complete if the email doesn't contain example.com. A real site email like admin@mycollection.org set at install time would correctly pass, but a site still using Drupal's default admin@example.com would be correctly flagged. This is reasonable and works in practice.

**5. aboutPageExists() LIKE %about%**
An About page named "Our Story" or "About the Collection" (without the standalone word "about") would not be detected. Low real-world impact since this is a heuristic -- the "Mark as done" manual override exists precisely for this case.

---

### LOOKS GOOD

- CSRF tokens on all 4 state-mutation routes (dismiss, restore, complete, uncomplete)
- mukurtu_protocol.module EXISTS-vs-JOIN fix is correct -- eliminates duplicate rows from media queries when a user has multiple protocol grants; named placeholders are properly parameterized
- MukurtuProtocolNodeAccessControlHandler.php revision operation additions (view revision, revert revision, delete revision) correctly pass through to parent handler
- Update hook 40061 guards against double-install and missing dashboard config
- DashboardBlockAccessSubscriber correctly gates the new block to administrator/mukurtu_manager
- Dismissal state stored in state (not config) -- correct for per-site runtime state

---

**Before merging I need to:**
1. Get your approval on the aria-label strings so I can apply the WCAG fix
2. Add block_content_list cache tag to the block
3. Commit today's text changes

michael-wynne-wsu:
All 5 issues fixed and committed. Here's what was done:

1. **Aria-labels** -- "Dismiss", "Mark as done", "Undo", and "Restore" links now carry aria-label attributes like "Dismiss Create a community" using Twig's %label placeholder (HTML-escaped).

2. **Uncommitted text changes** -- The SiteSetupTask.php type widening, all 11 updated task strings with Markup::create() doc links, and the 6 label changes are now committed.

3. **block_content_list cache tag** -- Added to SiteSetupChecklistBlock so the dashboard count refreshes when footer block content is updated.

4. **isSiteNameSet() heuristic** -- Now checks both the site name (must not be "Drupal", "My Drupal Site", or empty) and the email (must not contain example.com). Both conditions must pass.

5. **aboutPageExists() LIKE filter removed** -- Now detects any basic_page node, which is more reliable than hoping the title contains the word "about".

christa-kavanagh-wsu:
- [x] The "Create a Community" link directs you to `/admin/content/communities/add` instead of `/communities/community/add`
   - Done.

image

- [x] The "Create a Protocol" link directs you to `/admin/content/protocols/add` instead of `/protocols/protocol/add`.
   - Removed, see below.

image

- [x] I am not really sure how useful having this cultural protocol creation link is; if users create a community the site automatically directs them to create a cultural protocol, and the link is deactivated once an initial protocol has been created. It might be better to combine "Create a Community and Cultural Protocol" as one link/task.
   - Done.

- [x] Even though technically it isn't required to create a Mukurtu Manager account, I feel like that should probably be the first step and then the rest of the steps should be taken by the Mukurtu Manager.
   - Done.

- [ ] Is there a way to enable the check box to check off things that are not automatically checked off by the system?

- [x] The "Create an About Page" link directs you to `/node/add/basic_page` instead of `/node/add/page`.
   - Done.

- [ ] Why do some tasks have "Mark and Done" and "Dismiss" and some only have "Dismiss"?

image

- [ ] The "Configure site footer" link takes you to `/admin/content/block-content` instead of `/admin/content/block/1`.
   - Done, but this is potentially fragile if the block ID is different.

christa-kavanagh-wsu:
- [x] These are all the things that checked off for me after I completed them; site name and email were not checked off for me either even after rebuilding the cache:
   - Done, set to manual mark as done instead.

image

michael-wynne-wsu:
I'll check out the bad routes.

> It might be better to combine "Create a Community and Cultural Protocol" as one link/task.

Yeah that's easy enough to do, makes sense.

> Even though technically it isn't required to create a Mukurtu Manager account, I feel like that should probably be the first step and then the rest of the steps should be taken by the Mukurtu Manager.

I can see that.

> Is there a way to enable the check box to check off things that are not automatically checked off by the system?

That is the "Mark as done" actions. That's on things the site can't really tell are completed.

> Why do some tasks have "Mark and Done" and "Dismiss" and some only have "Dismiss"?

See above.

> Site name and email were not checked off for me either even after rebuilding the cache:

Okay will dig back into that one.

michael-wynne-wsu:
Okay I've addressed all actionable comments, and also changed to front page redirect (but that's potentially fragile).

christa-kavanagh-wsu:
- All of the links work, thank you!
- I like creating a Mukurtu Manager from the start.
- The combined community/protocol workflow works well.
- As Mukurtu Manager, I don't have access to configure the Landing Page.

image

- Could you please add a "Mark as done" option to the "Create an about page" link?
- Is there a way to automatically remove the Site Setup link from the left-hand admin menu after all of the tasks have been checked off or dismissed?

michael-wynne-wsu:
> As Mukurtu Manager, I don't have access to configure the Landing Page.

Separate issue for that please

> Could you please add a "Mark as done" option to the "Create an about page" link?

Will do

> Is there a way to automatically remove the Site Setup link from the left-hand admin menu after all of the tasks have been checked off or dismissed?

We'll want it removed from the dashboard too. And something similar with the migration block too...

michael-wynne-wsu:
About page now has mark as done, and the whole site setup menu auto hides when all tasks are completed or dismissed.

michael-wynne-wsu:
## PR Review: Site Setup Checklist (feature/366-site-setup-checklist)

**Blast Radius:** LOW-MEDIUM -- new self-contained module; mukurtu_core touched for update hook and block access; mukurtu_protocol touched for media dedup fix.

**Security:** PASS -- CSRF on all mutation routes, role access on all controllers, parameterized entity queries, Markup::create() on hardcoded URLs only.

**Breaking Changes:** None -- new module, additive config and update hook.

**Tests:** None (admin-only UI feature, acceptable at this stage).

---

### MUST FIX

**1. Dead code: create_protocol case in isComplete() match**
SiteSetupTaskManager.php line 688

The create_protocol task no longer exists in getTasks() -- it was merged into create_community. The match arm is never reachable. Remove it to avoid confusing future maintainers.

// Remove this line:
'create_protocol' => $this->entityExists('protocol'),

---

### SHOULD FIX

**2. SiteSetupController::page() missing mukurtu_setup:tasks cache tag**
SiteSetupController.php lines 186-194

The block and menu link both include mukurtu_setup:tasks in their cache tags so they invalidate on state changes (dismiss/complete). The full page at /admin/mukurtu/setup does not. After dismissing a task, Cache::invalidateTags(['mukurtu_setup:tasks']) fires, but the page render cache won't be busted -- a subsequent visit could serve stale content until another tag (like community_list) happens to invalidate.

Fix: add 'mukurtu_setup:tasks' to the page controller's #cache['tags'].

**3. create_community auto-detection only checks for a community, not a protocol**
SiteSetupTaskManager.php line 687

The task is now labeled "Create a community and cultural protocol" (canAutoDetect: TRUE), but isComplete() only calls $this->entityExists('community'). A site could have a community with no protocols and the task would still show as complete. Either also check $this->entityExists('protocol'), or accept the community-only proxy and note it in a comment.

---

### LOOKS GOOD

- CSRF tokens on all 4 mutation routes
- Cache::invalidateTags(['mukurtu_setup:tasks']) fires correctly in all 4 state-mutation methods
- SetupMenuLink::getCacheTags() returns entity list tags so menu cache invalidates on entity changes
- Block returns empty cache-only array when all done -- correct Drupal pattern to suppress a block without an access denied error
- SiteSetupDismissController validates task IDs against the known task list before acting -- no arbitrary state injection
- DashboardBlockAccessSubscriber correctly gates the new block to administrator/mukurtu_manager
- mukurtu_core_update_40061() guards against double-install and missing dashboard config
- mukurtu_protocol.module EXISTS-vs-JOIN fix correctly eliminates duplicate rows with named placeholders
- MukurtuProtocolNodeAccessControlHandler revision operation additions are correct pass-throughs
- @label placeholder (not %label) used in aria-labels -- correct, avoids em tags leaking into attributes

---

Two things to fix before merging: remove the dead create_protocol match arm, add mukurtu_setup:tasks to the page controller's cache tags.

michael-wynne-wsu:
Done. Both fixes applied and committed:

1. Removed the dead 'create_protocol' => $this->entityExists('protocol') match arm from isComplete().
2. Added 'mukurtu_setup:tasks' to the page controller's cache tags so the full checklist page invalidates on dismiss/complete/restore, matching the block and menu link behavior.

michael-wynne-wsu:
@nick-deer sorry for jumping the gun on this, should be review-ready now.

nick-deer (approved):

PR #1757, opened 2026-06-24
Fix/1756 media library duplicate entries

### LINKED ISSUE #1756: Media modal showing multiple entries
When creating a media asset through the modal, if you select more than one protocol and the the Any sharing setting, the media library modal shows multiple entries for that media asset.

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1756

Pre-merge checklist:


michael-wynne-wsu:
## PR Review: Fix duplicate media library entries (#1756)

**Blast Radius:** LOW -- single internal query hook in `mukurtu_protocol.module`, no shared API contracts, no schema changes.

**Security:** No findings -- named PDO placeholders (`:grant_0`, `:grant_N`) used correctly; `{$mediaTable}` is a Drupal query-builder alias, not user input.

**Tests:** No new tests added for the specific regression scenario; pre-existing tests pass.

**Breaking Changes:** None -- behavior change is intentional and correct. "all" sharing path (single compound row) is unaffected.

---

### MUST FIX
None.

---

### SHOULD FIX

**1. No regression test for the bug scenario.**

The bug -- N rows returned for a media entity with N protocols under "any" sharing -- has no kernel test. A test case in `AccessByProtocolTest.php` that creates a media entity with 2 protocols under "any" sharing, queries it as a non-admin member, and asserts the result count is exactly 1 would prevent regressing this in future.

Not blocking since the fix was verified manually and the root cause is fully understood, but it's worth filing a follow-up ticket.

---

### LOOKS GOOD

- EXISTS subquery is semantically correct: returns exactly one row per entity regardless of how many grant rows exist in `mukurtu_protocol_access`.
- UID 1 early-return is preserved -- admins are unaffected.
- Placeholder generation is safe: `getAccountGrantIds()` always returns at least the deny grant (id 0), so the IN clause is never empty.
- "all" sharing path (inserts one compound row) was never affected by the bug and is unchanged.
- The fix is minimal -- no collateral changes, no refactoring.
- CSS compilation committed separately and cleanly.

michael-wynne-wsu:
@christa-kavanagh-wsu I did test and verify on tugboat, I don't think you need to.

christa-kavanagh-wsu (approved):
That did it!

nick-deer (approved):

PR #1762, opened 2026-06-25
Add header nav search box (issue #1331)

### LINKED ISSUE #1331: Add a search box to the navigation bar
- [ ] Add a general search box to the main menu or header (site-wide).
 - [ ] Should search and return results at /browse (or /digital-heritage).
 - [ ] Optional, allow admins to add/remove the box and to configure which search page it indexes/returns.

Example 1

Image

Example 2

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1331

Renders a search form inside the main menu nav bar. Submitting sends the user to /browse (or /digital-heritage) with the search_api_fulltext query parameter. Enabled by default; admins can toggle it and configure the destination path at admin/config/mukurtu/search/settings.

Pre-merge checklist:


michael-wynne-wsu:
@joelle-birano-wsu @christa-kavanagh-wsu ready to test. A lightly-configurable nav menu search box.

Screenshot 2026-06-25 at 12 40 32 PM

Screenshot 2026-06-25 at 12 53 55 PM

joelle-birano-wsu (approved):
Works as expected for me!

christa-kavanagh-wsu (commented):
So far as I can tell it's working as a full text search everywhere except the Person Records Biography formatted text field and the Place Records text sections formatted text field.

michael-wynne-wsu:
> So far as I can tell it's working as a full text search everywhere except the Person Records Biography formatted text field and the Place Records text sections formatted text field.

Is that different from searching at /browse directly?

christa-kavanagh-wsu:
> > So far as I can tell it's working as a full text search everywhere except the Person Records Biography formatted text field and the Place Records text sections formatted text field.
> 
> Is that different from searching at /browse directly?

Neither of them are pulling from the biography/text sections.

michael-wynne-wsu:
> > > So far as I can tell it's working as a full text search everywhere except the Person Records Biography formatted text field and the Place Records text sections formatted text field.
> > 
> > 
> > Is that different from searching at /browse directly?
> 
> Neither of them are pulling from the biography/text sections.

Okay that is a separate issue then, log that.

michael-wynne-wsu:
FYI I adjusted it so when the menu collapses on small screens the search box is still exposed.

michael-wynne-wsu:
## PR Review: Add Header Navigation Search Box (#1331)

**Blast Radius:** LOW -- changes are confined to the `mukurtu_search` module (config/settings form/update hook) and the `mukurtu_v4` theme (one block template, new SCSS partial). No shared libraries, API contracts, or DB schema involved.

**Security:** Clean -- no hardcoded secrets, no XSS vectors, no SQL interpolation. Config values are admin-controlled and Twig-escaped on output.

**Tests:** No automated tests added or changed. Appropriate for a theming/settings-form change with no business logic.

**Breaking Changes:** None. Update hook 40004 uses NULL-guards so existing installs get conservative defaults without clobbering any overrides.

---

### FIXED (applied in this session)

**1. Duplicate `$form['header_search']` definition -- SearchSettingsForm.php**
The fieldset was assigned twice, with the second assignment appearing after the `backend` radios and silently overwriting the first. The dead first block was removed and the single remaining definition reordered so the form renders: Browse display -> Header search -> Search Backend.

**2. Accessibility: input focus ring fails 1.4.11 Non-text Contrast -- _header-search.scss**
`outline-offset: 0` on `.header-search__input:focus` placed the ring directly against the `#767676` border (~1.4:1 contrast against the focus color). Changed to `outline-offset: 2px` so the ring sits against the white background.

**3. Accessibility: touch target too small (2.5.8) -- _header-search.scss**
Submit button was ~34px block-size. Added `min-block-size: 2.75rem; min-inline-size: 2.75rem` to meet 44px touch target recommendation.

**4. Accessibility: missing `aria-label` on `role="search"` landmark -- block--mukurtu-v4-main-menu.html.twig**
Added `aria-label="{{ 'Search'|t }}"` to both `<form>` elements so the landmark has a name for screen reader users.

**5. Spurious `.node--status-banner` in compiled style.css**
Prior commits compiled `style.css` from DDEV where `_full-node.scss` has been updated ahead of the repo. Recompiling from the repo restores `.node--draft-banner` and removes the unintentional diff.

---

### SHOULD FIX before merge

**6. `header_search_path` has no input validation on save**
`submitForm` stores the raw `header_search_path` value without checking it against the allowed option list. An admin can submit an arbitrary path if the form is bypassed (e.g., via REST). Low risk given it's admin-only, but worth a one-line allowlist check:
```php
$allowed = ['/browse', '/digital-heritage'];
$path = in_array($form_state->getValue('header_search_path'), $allowed, true)
  ? $form_state->getValue('header_search_path') : '/browse';
$config->set('header_search_path', $path);
```

---

### SUGGESTIONS

**7. `#title` on `backend` fieldset is missing `$this->t()`**
```php
'#title' => 'Search Backend',  // not translated
```
Should be `$this->t('Search Backend')`. Not new to this PR but worth fixing since the surrounding code was touched.

---

### LOOKS GOOD

- Cache tag propagation (`getCacheTags()`) ensures the block renders correctly after settings changes without manual cache clears
- The two-instance responsive pattern (mobile/desktop) is well-documented in the SCSS file header, and `display: none` correctly removes the hidden instance from the accessibility tree
- `role="search"` as a landmark on the form element (not a wrapper div) is the correct ARIA pattern
- SVG icon: `aria-hidden="true"` and `focusable="false"` (IE11 fix) both present
- Update hook 40004 uses NULL-guards to preserve any existing site-specific overrides
- Config YAML defaults match the update hook defaults -- consistent for new installs vs upgrades
- `#states` wiring hides the path selector when the search toggle is off -- clean UX

---

## Summary of changes made

| File | Change |
|------|--------|
| SearchSettingsForm.php | Removed duplicate `$form['header_search']` block; reordered form fields correctly |
| _header-search.scss | `outline-offset: 2px` on input focus; `min-block-size/min-inline-size: 2.75rem` on button |
| block--mukurtu-v4-main-menu.html.twig | Added `aria-label="{{ 'Search'|t }}"` to both `<form>` elements |
| style.css | Recompiled from repo SCSS (fixes spurious `.node--status-banner` diff) |

michael-wynne-wsu:
All 8 issues addressed. Here's the complete tally:

| # | Issue | Status |
|---|-------|--------|
| 1 | Duplicate `$form['header_search']` definition | Fixed |
| 2 | Input focus ring `outline-offset: 0` fails 1.4.11 | Fixed |
| 3 | Button touch target ~34px, below 44px minimum | Fixed |
| 4 | Missing `aria-label` on `role="search"` landmark | Fixed |
| 5 | Spurious `.node--status-banner` in style.css | Fixed (recompiled) |
| 6 | `Cache` class import check | No action (already imported) |
| 7 | `header_search_path` no allowlist validation on save | Fixed |
| 8 | `'#title' => 'Search Backend'` missing `$this->t()` | Fixed |

joelle-birano-wsu:
The search bar staying exposed for smaller screens works nicely. This is nitpicky but I'm just wondering if it can be a bit higher and stay sandwiched between the logo and menu? It almost seems like it drops down to the block below

Screenshot 2026-06-26 at 11 27 51 AM

Also I noticed this padding at the top before but it always resolved with a refresh. It seems persistent on this branch.

Screenshot 2026-06-26 at 11 35 15 AM

michael-wynne-wsu:
I'll see about the first thing.

The second is just a tugboat cache issue, nothing that should really be a problem.

michael-wynne-wsu:
Okay it moves on smaller screens. If you're just dragging the window size you'll need to reload the page, but if it loads at any given size it's placed where we want it.

nick-deer (approved):

joelle-birano-wsu (approved):
Cool! Looks good!

PR #1728, opened 2026-06-13
Export collections and word lists with content

michael-wynne-wsu:
- In "add to export list" confirmation for collections and word lists there is a "include x child content" selection.
- Same selection on ad hoc export page.
- Includes options for recursive collection export.
- CR and MPI are handled in a different PR.
- Protocols are respected when content in collections/lists has differing protocols from parent.

To do:

- [x] Confirm that previous branches have no regressions:
   - [ ] https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1725
   - [ ] https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1720
   - [ ] https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1726
   - [ ] https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1727

Pre-merge checklist:

- [x] I have checked for and if required, created update hooks.
- [x] I have run an accessibility check, and resolved any issues.
- [x] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [x] I have updated from `main` and resolved any merge conflicts.
- [x] I have verified that all expected checks pass.
- [x] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1767, opened 2026-06-26
Feature/578 taxonomy term page improvements

### LINKED ISSUE #578: /taxonomy/term/____
For v4 launch, maintain same functionality as v3. May improve in a future release (see #356 )

Not themed yet.

- [ ] Add vocabulary name to page title. Eg: "Art" > "Keyword: Art"
- [ ] Remove "Referenced Content" label.
- [ ] Add informational text. "All available content using the [vocabulary name] term [term name]."
- [ ] Remove search and sort by.
- [ ] Default list view, I guess?
- [ ] Ensure that there is pagination.

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/578

- Add vocabulary-prefixed page title (e.g., "Keyword: Art") via title callback on the canonical route
- Add introductory text "All available content using the [vocab] term [name]."
- Display term description when present
- Remove "Referenced Content" heading
- Disable exposed search and sort controls on the referenced content view
- Render referenced content in a grid layout using existing .grid CSS rules
- Fix edge case: empty records array no longer renders the multi-tab UI

Pre-merge checklist:


michael-wynne-wsu:
Alright give it a go! @joelle-birano-wsu @christa-kavanagh-wsu 

Screenshot 2026-06-26 at 11 39 33 AM

christa-kavanagh-wsu:
This seems to work as expected! Are you going to style these in the same way that related content/referenced content cards are styled? 

image

joelle-birano-wsu (commented):
This is looking good to me. I think the term in the introductory sentence could use quotation marks or something to distinguish it. 

i.e. All available content using the keyword term "flowers."

michael-wynne-wsu:
>  Are you going to style these in the same way that related content/referenced content cards are styled?

Probably that yeah. I'm going to address that again later since we need to look at all the different view modes and cards. I should probably bump that up the list.

michael-wynne-wsu:
> This is looking good to me. I think the term in the introductory sentence could use quotation marks or something to distinguish it.

I make it italic (emphasis).
Also 4 column grid.

joelle-birano-wsu (commented):
4 column grid works for four items, but as soon as there's more it drops to 3 column.

Screenshot 2026-06-26 155617
Screenshot 2026-06-26 155534

michael-wynne-wsu:
oh, that's interesting...

michael-wynne-wsu:

Link:
- https://pr1767-xuoxpcdgvp45d8f1tjoslhqggrjpxscu.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a3dd36bb3c77dcef9ce8e10

michael-wynne-wsu:
Finally got it. 4 columns, scales down to 2 at a certain point.

Screenshot 2026-06-26 at 10 21 25 PM
Screenshot 2026-06-26 at 10 21 31 PM
Screenshot 2026-06-26 at 10 22 00 PM

michael-wynne-wsu:
## PR Review: Feature/578 taxonomy term page improvements (#1767)

**Blast Radius:** MEDIUM - affects all taxonomy term canonical pages sitewide (every vocabulary), plus the Place full-node template.
**Security:** No findings.
**Tests:** 0 new tests for 3 new controller methods.
**Breaking Changes:** None in config or schema.
**CI:** PHP tests fail on pre-existing `mukurtu_import` failures unrelated to this PR. Playwright and Tugboat pass.

---

### MUST FIX (Blocking)

**1. Facets source ID is stale after view change**

`getFacetSourceId()` still returns `search_api:views_block__mukurtu_taxonomy_references__content_block`. The referenced content now comes from the core `taxonomy_term` view via `taxonomy_index`, not Search API. Any facets configured for the old source ID won't filter the new results. If facets were already broken pre-PR (the comment confirms the UUID fulltext filter always returned empty), this is at minimum a continuation of a broken state that should either be fixed here or explicitly called out. If facets are expected to work, the source ID needs to be updated or facets need to be reconfigured for the new view.

---

### SHOULD FIX (Non-blocking)

**2. Accessibility: `.referenced-content.grid` has no accessible name**

The `<h2>Referenced Content</h2>` heading was removed. Screen reader users navigating by headings lose the landmark for this section. The fix is an `aria-label` on the container div:
```twig
<div class="referenced-content grid" aria-label="{{ 'Referenced content'|t }}">
```

**3. `getSingularVocabularyLabel()` fallback is not guaranteed singular**

For unmapped vocabularies the method falls back to `$vocabulary->label()`, which is whatever the admin typed - typically plural ("Keywords", "Tags"). The page title would then read "Keywords: Art" even though the map explicitly converts `keywords` to "Keyword". A site with custom vocabularies will hit this. At minimum, add a comment documenting the limitation and that any new Mukurtu vocabulary needs a map entry.

**4. View option mutations could leak across requests**

`$view->display_handler->overrideOption('header', [])` and `overrideOption('row', [...])` mutate the live view object. The core `taxonomy_term` view is used elsewhere (term listing pages). If the view object is cached in the Views static cache within a single request, these overrides could bleed into other places that call the same view. Standard Drupal practice is to use a fresh view or set display options at render time. Low risk in practice since term page is a full page render with no other taxonomy_term view on the same request, but worth noting.

---

### SUGGESTIONS

**5. `<em>` tag embedded in `t()` source string**

```twig
{{ 'All available content using the @vocab term <em>@term</em>.'|t({...}) }}
```

Translators will see the `<em>` tag in their translation interface. This is valid Drupal practice but somewhat confusing. An alternative is wrapping `term_name` in `<em>` in the template markup rather than the translation string. Minor.

**6. No tests for new public methods**

`title()` and `getTermDescription()` are public controller methods with logic worth unit-testing (vocabulary label mapping, check_markup integration, empty description handling).

---

### LOOKS GOOD

- The switch from Search API UUID fulltext to core `taxonomy_term` view with `setArguments([$id])` is the correct fix. The comment in the diff explaining why the old approach always returned empty is clear.
- `check_markup()` usage in `getTermDescription()` is correct - it returns a `Markup` object that Twig's auto-escape respects as safe, with no XSS risk.
- The `elseif records|length > 1` change is a subtle improvement over the old `else` - avoids rendering an empty tab pane when there are 0 community records.
- CSS Grid on `.views-element-container > div` cleanly sidesteps the `column-fill: balance` algorithm issue. Selector correctly targets Drupal 11's views HTML structure.
- `node--place--full.html.twig` `|without('flag_export_content')` fix is clean and correctly undoes the regression from PR #1728.
- No security issues: `@term`/`@vocab` in `t()` auto-escape user content, no raw SQL, no hardcoded secrets.

---

**Summary:** One functional blocker (facet source ID), one accessibility should-fix, and a documentation note on the vocabulary label fallback. The core work - the view fix, the grid layout, the page title callback - is solid.

michael-wynne-wsu:
All 6 PR review issues addressed:

1. Facets source ID stale -- Added @todo to getFacetSourceId() explaining facets are non-functional with the SQL-backed taxonomy_term view
2. Missing aria-label on referenced content grid -- Added aria-label="Referenced content" (WCAG 1.3.1)
3. Vocabulary label fallback not singular -- Extracted map to VOCABULARY_LABEL_MAP class constant with doc comment explaining the fallback behavior
4. View mutation leak (false positive) -- Added comment clarifying Views::getView() returns a fresh ViewExecutable each call
5. <em> in t() string -- Replaced <em>@term</em> with Drupal's %term placeholder (auto-wraps in <em>, no HTML in translatable strings)
6. No tests -- Added 8 unit tests for getSingularVocabularyLabel() and title(), all passing

Commits: c9bfdf5a, 3b994cc3

nick-deer (approved):

christa-kavanagh-wsu (commented):
This looks pretty good. Is there a way to further separate the "All available content using the Keyword term _______." from any other text? Like maybe using it as a header for the content section? It kind of runs together if you have a fuller description for a taxonomy term.

Image

joelle-birano-wsu:
4 column grid is working well. I agree with Christa about somehow differentiating the taxonomy description from "All available content..."

michael-wynne-wsu:
Alternatively, is that message even necessary?

christa-kavanagh-wsu:
I don't think it is necessary, but I do think there should be some sort of separation between the description and the content. Even a line or something to help define the different areas.

PR #1768, opened 2026-06-26
Fix kernel test failures caused by missing image config install

michael-wynne-wsu:
@nick-deer Hopefully this resolves that image kernel test failure...

Possibly unrelated, I see these kernel test are PHP 8.3, should we be using 8.4?

The image module was listed in \$modules but not in installConfig(), causing og's config dependency chain to fail when resolving the image field type. Affects all kernel tests extending ProtocolAwareEntityTestBase.

Pre-merge checklist:


michael-wynne-wsu:
There are other failing tests, I was planning to address those after.

nick-deer:
@michael-wynne-wsu This solved the kernel error! However I got one test failure:

1) Drupal\Tests\mukurtu_export\Kernel\ExportChildResolverTest::testRecursiveCycleDetectionPreventsInfiniteLoop
Failed asserting that an array does not have the key 1.

I'm also not sure where to set the PHP version for these tests... when I run them on my local, since I have my ddev PHP version up to 8.4, it runs the tests on 8.4.

michael-wynne-wsu:
@nick-deer I was going to address all the errors (this PR and then this one https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1769) and THEN look into the failures after.

Also I think I found the PHP setting in this doc https://github.com/MukurtuCMS/Mukurtu-CMS/blob/537dffb1c0fe6022e4d236f3d7f83e685a17f19c/.github/workflows/build-and-test.yml#L16

nick-deer (approved):
Alright, works for me!

PR #1779, opened 2026-07-01
Fix fatal error on visitor account registration

### LINKED ISSUE #1775: Error when visitor registers new account
If I try to request a new account as a visitor, I get an error. The account is still created - if I refresh itʻll say that usernameʻs already taken. At /admin/people, the new accounts are there. 

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1775

## Summary
- `user_role_names()` was removed in Drupal 11, causing `mukurtu_notifications_user_insert()` to fatal-error on every new user save and block visitor self-registration entirely.
- Replaced it with `Role::loadMultiple()` + `hasPermission()` to find roles with `administer users` permission.
- Added the now-required `accessCheck(FALSE)` to the follow-up entity query (a second latent Drupal 11 incompatibility that was unreachable until the first bug was fixed).

## Test plan
- [x] Reproduced the original fatal error locally via `drush php:eval` simulating a blocked user insert
- [x] Applied fix, cleared cache, re-ran the same simulation — no errors/warnings logged
- [x] Cleaned up test user
- [x] Confirm `/user/register` completes successfully in a live visitor registration flow


michael-wynne-wsu:
@joelle-birano-wsu this worked on my local, please confirm here.

joelle-birano-wsu (approved):
Works on my end!

michael-wynne-wsu:
@nick-deer just a small bug fix.

nick-deer (approved):

PR #1780, opened 2026-07-01
Require visitors to enter a password when no approval or email verification

### LINKED ISSUE #1776: Re-route users who create their own account
This is an odd result for these registration settings: 

Visitors, no approval required
Email verification: off
Visitor password entry: Visitors may enter a password on registration. 

If visitor doesnʻt enter a password, they donʻt receive an email (tested on ddev - it was provided in the form), they are logged in, and their password is auto-generated. If they log out, they wonʻt be able to log back in.

Image

They should probably either receive a welcome email with password reset link, OR if logged in theyʻre taken to their account page where they can add their password.

michael-wynne-wsu:
## Summary
- With "Visitors, no approval required" and email verification off, admins could previously set "Visitor password entry" to "visitors may enter a password". A visitor who left the password field blank was logged in immediately with an unknown system-generated password, and locked out once they logged out.
- Genpass mode "cannot enter a password" already has a working fix for this (redirects to set a password), so it's left alone.
- This PR disallows the broken combination directly: server-side validation now rejects it, the option help text is updated, and the client-side JS disables/auto-switches the "may enter" radio to "must enter" with an accessible live-region announcement when that combination is selected.

## Test plan
- [x] Verified in DDEV via Playwright: with register=visitors (no approval) + verify_mail off, the "may enter" radio is disabled and auto-switches to "must enter", with the live-region announcement firing on toggle.
- [x] Verified switching to "administrator approval required" re-enables the "may enter" option (no regression).
- [x] Verified server-side validation (`mukurtu_core_user_admin_settings_validate()`) rejects the combination directly.
- [x] Ran the WCAG accessibility skill against the changed elements — no issues found.
- [x] Reviewer: register a test visitor account under the new default (mode 0) to confirm the standard flow is unaffected.

Fixes #1776


michael-wynne-wsu:
@joelle-birano-wsu see how this does now. I removed the "may" create a password option for simplicity.

joelle-birano-wsu (approved):
Works great! I like that you disallowed "may."

michael-wynne-wsu:
@nick-deer another small change, not quite a bug fix, just a user registration setting improvement.

nick-deer (approved):

PR #1781, opened 2026-07-01
Fix bulk user actions on Site Users page skipping confirmation

michael-wynne-wsu:
Resolves bulk action user issues.

## Summary
- The three bulk actions on `/admin/people/list` (Block or delete, Unblock or approve, Set pending) skipped the confirmation step entirely and, after applying, showed a garbage status message like `(2)` instead of anything meaningful.
- Root cause: the actions' `confirm = TRUE` annotation key is not read by Views Bulk Operations at all (dead config), and `execute()` returned nothing, so VBO's batch-finished message became an empty string plus the entity count.
- "Unblock or approve" and "Set pending" now use VBO's built-in confirmation step (with action-specific help text) and return a real success message (e.g. "Unblocked 2 user(s).").
- "Block or delete" now redirects to a new confirmation form that reuses Drupal core's actual account cancellation logic (`user_cancel()` / `user_cancel_methods()`) — the same options offered at `/user/{uid}/cancel` (disable & keep content, disable & unpublish, delete & reassign, delete & delete) — instead of a bare "are you sure?" prompt.
- Added update hook `mukurtu_protocol_update_40039` to ship the view config change to existing sites.

## Test plan
- [x] Manually verified in DDEV via Playwright: Unblock or approve and Set pending each show correct confirmation text and produce a real success message.
- [x] Manually verified Block or delete shows the cancellation-method radios, Cancel returns to `/admin/people/list`, and Confirm with "Delete the account and its content" actually deletes the account (checked at the DB level).
- [x] Confirmed the update hook (`drush updb`) correctly re-imports the view's bulk-action preconfiguration on an existing site.
- [x] Accessibility review of the new confirmation form (pending)
- [x] Verify community-manager-scoped access still filters correctly on all three actions in a non-admin account


michael-wynne-wsu:
## Review: PR #1781 — "Fix bulk user actions on Site Users page skipping confirmation"

### Accessibility (WCAG 2.2 AA) — MukurtuUserCancelConfirmForm.php

Looks good: The radios group, conditional plain-text summary, and checkboxes all follow Drupal core's own accessible UserCancelForm pattern (fieldset/legend, aria-describedby linkage). No fieldset/legend defects, no missing labels, no keyboard/focus issues.

Findings:
1. Unlabeled account list (MukurtuUserCancelConfirmForm.php:158) - the item_list of affected accounts has no #title, so it renders as a bare <ul> with nothing announcing what it is. The very trait this form reuses (ViewsBulkOperationsFormTrait) always sets a title on its equivalent list - this is the one place it diverges. Needs a #title (e.g. "Users to be blocked or deleted") - flagging for approval per the user-facing-text rule before anyone writes copy.
2. "No accounts" fallback message (line 146) is plain #markup with no visual/semantic emphasis as an outcome message - minor, worth aligning with how other no-op admin messages are styled.
3. Low-confidence note: line 172 reads $form['user_cancel_method']['#default_value'] without an existence check - core always sets a default so this is very unlikely to fire, just flagged.

### PR review

Verified the three riskiest technical claims directly against the views_bulk_operations and Drupal core source rather than taking the PR description at face value:

- Confirmed: confirm_form_route_name really does make execute() unreachable. ViewsBulkOperationsBulkForm::viewsFormSubmit() redirects to the confirm route whenever it's non-empty and only falls through to direct batch execution otherwise (ViewsBulkOperationsBulkForm.php:946-950). The comment in MukurtuBlockUserAction.php is accurate.
- Confirmed: UID-1 / access checks are correct and not bypassable. The form filters accounts via $account->id() <= 1 plus $action->access($account, $current_user), and MukurtuBlockUserAction::access() independently blocks self-blocking, protects roles, and scopes community managers to their own communities - same per-account gate used elsewhere, so community-manager scoping still works through the new confirm form.
- Confirmed: Permission logic matches core exactly. select_cancel = hasPermission('administer users') || hasPermission('select account cancellation method') is byte-for-byte the same condition core's UserCancelForm uses.
- Confirmed: Update hook is safe. Full view-YAML overwrite via setData()->save(TRUE) is idempotent and matches the exact established pattern already used 6+ times elsewhere in this same .install file - no new risk.

Must fix (blocking) - this is a real bug, not a style nit:

- MukurtuSetPendingUserAction.php and MukurtuUnblockUserAction.php still return '' from their early-return branches (wrong entity type / access denied) - only the success path was given a real message. VBO's ViewsBulkOperationsActionCompletedTrait::finished() groups results by exact message string; anything that returns '' displays as a bare (N) - this reproduces the exact garbage-message bug the PR sets out to fix, but now only for partially-successful selections (e.g. bulk-unblocking 5 users where 2 are inaccessible to the current user shows "Unblocked 3 user(s)." and a separate "(2)" line). Fix: give the skip branches real messages too (e.g. "Skipped @count user(s) - access denied.").

Suggestions (non-blocking):
- Consider a short message for the no_accounts fallback in the confirm form consistent with the messenger pattern noted in the accessibility review.

Everything else (view config, routing, form structure, self-cancel/UID-1 protection) is solid and consistent with existing conventions.

michael-wynne-wsu:
All three files edited and lint-clean. Summary of changes on fix/bulk-user-action-confirmation:

- MukurtuSetPendingUserAction.php / MukurtuUnblockUserAction.php - skip branches (wrong entity type, access denied) now return "Skipped @count user(s) (access denied)." instead of '', preventing the bare (N) garbage-message regression in partially-successful bulk selections.
- MukurtuUserCancelConfirmForm.php - the "no accessible accounts" fallback is now wrapped in a container with Drupal's standard messages messages--warning classes and role="alert", giving it visual emphasis and screen-reader announcement (text unchanged, per your call).

nick-deer (approved):

PR #1782, opened 2026-07-01
Feature/1773 archive cancelled user content

### LINKED ISSUE #1773: Block and Unpublish at /admin/people not working
at /admin/people (only applicable to bulk actions):

When deleting a user,  choosing to block the account and unpublish their content, the account is blocked but the content stays published. 

Image

Image

Image

michael-wynne-wsu:
Fixes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1773

When blocking/deleting a user or users,

- [ ] Replaces "Block the user account(s) and unpublish their content." with "Block the user account(s) and archive their content." since we removed unpublish from our states. Should would with the basic and advanced editorial workflows.
- [ ] Adds the archived state to the basic workflow.

Pre-merge checklist:


michael-wynne-wsu:

Link:
- https://pr1782-89dhvg53xlrb82efcspkhshsoxkfxfvw.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a4587ccb79fb052e4ead566

joelle-birano-wsu (commented):
Block and archive works. As an admin, if I edit a blocked user's archived item I only see the draft state. This happens with both the basic and editorial workflows - just wanted to double check that that was expected behavior.

Image

 I also checked admin/people/list. Operations menu looks fine but bulk actions still have the old verbiage for all the options. It seems to work fine - user is blocked and content is archived, so I think it's just the text that needs updating.

Image

michael-wynne-wsu:
> if I edit a blocked user's archived item I only see the draft state. 

No, we'll want to be able to move it back to published. That might have been part of #1742 but I'll confirm or resolve.

> I also checked admin/people/list.

Good catch, on it.

michael-wynne-wsu:
I think I've got both of those resolved, test again please.

christa-kavanagh-wsu (approved):
I noticed that we have two Block options on here: "Block" and "Block and Delete". I don't remember if that was intentional? It is also that way on main. 

Otherwise everything looks good!

michael-wynne-wsu:
@christa-kavanagh-wsu We included the separate block for community managers, since we didn't want to over-expose tools. SO yes it is intentional, if a little odd for regular site admins.

Screenshot 2026-07-02 at 7 58 08 AM

michael-wynne-wsu:
## Accessibility check — wcag-accessibility-compliance

Focused on the only new markup in this diff, MukurtuUserCancelConfirmForm.php (the views.view.mukurtu_people.yml confirmation dialogs render through VBO/core's own templates — no issues found there):

1. modules/mukurtu_protocol/src/Form/MukurtuUserCancelConfirmForm.php:142 — $form['#title'] is dead code. The route has a static _title and no _title_callback, and TitleResolver never reads a _form route's #title. The dynamic confirm question never becomes the page heading (WCAG 2.4.6).
2. MukurtuUserCancelConfirmForm.php:111-117 — The "Cancellation method" radios render as a bare <div> (core's radios.html.twig), with no <fieldset>/<legend> tying the heading to the 4 radio options (1.3.1). Pre-existing Drupal core pattern, not a regression, but surfaced here.
3. MukurtuUserCancelConfirmForm.php:119-125 — The no-permission fallback (#type: item / #plain_text) showing the operative cancel method isn't programmatically associated with the checkboxes/submit button it governs (1.3.1).
4. MukurtuUserCancelConfirmForm.php:87-96 — role="alert" on the no-accounts message is part of the initial server-rendered page, not a dynamic update, so it won't reliably announce to screen readers. Combined with #1, a screen-reader user can land on this page without the warning being surfaced (4.1.3).

## PR review — pr-review-expert

MUST FIX

1. modules/mukurtu_workflows/mukurtu_workflows.module:590-604 — mukurtu_workflows_user_cancel() force-archives every moderated node owned by the cancelled user, with no check on the node's current state. A draft or needs_review node that was never public gets pushed straight to archived, even though the workflow's own transition graph only allows archive from published. This bypasses the FSM, mislabels never-published content as "archived," and writes a misleading revision log ("Moderation state changed to archived.") on content that was never live. Fix: only archive nodes whose current moderation_state is published (or $node->isPublished()); leave other states untouched since they're already non-public.

SHOULD FIX

2. modules/mukurtu_protocol/src/Plugin/Action/MukurtuBlockUserAction.php:31-34 — execute() is now dead code (verified: VBO redirects to the confirm-form route before ever calling execute()), kept only "for safety/BC." Comment is honest about it, but it's untested dead code in a security-relevant path (account cancellation) — worth either removing or adding a test that exercises it if it's meant as a real fallback.
3. modules/mukurtu_protocol/src/Form/MukurtuUserCancelConfirmForm.php:113 and modules/mukurtu_core/src/Hook/FormHooks.php — the new form reaches into FormHooks::relabelCancelMethods() (a mukurtu_core hook-handler class) as a static utility from mukurtu_protocol. It works, but couples an unrelated module's hook class to this form; consider extracting the shared cancel-method relabeling into a small service if it needs a third caller.

TESTS

No test files changed for this branch (git diff main...HEAD --name-only has no Test matches). The new mukurtu_workflows_user_cancel() hook and MukurtuUserCancelConfirmForm are both good candidates for a kernel/functional test, especially given finding #1.

LOOKS GOOD

- confirm_form_route_name / add_confirmation usage in the VBO action plugins and view config verified correct against the installed views_bulk_operations module (route params, tempstore flow, confirm_help_text rendering all line up).
- Bulk-form-key decoding ([$uid] = $item) matches VBO's actual key encoding for a user-base view.
- Access control (uid <= 1 guard, $action->access() check, route's _custom_access) is consistent with existing patterns (MukurtuUserController::access).
- New update hooks (mukurtu_workflows_update_40005/40006, mukurtu_protocol_update_40039) are correctly sequential, no collisions, and the config/install YAML for fresh installs matches what the update hooks backfill for existing sites.
- No secrets, SQL injection, eval, or other security-scan hits in the diff.

joelle-birano-wsu (approved):
Looks good!

michael-wynne-wsu:
All 7 findings are fixed, verified, and synced into DDEV. Nothing has been committed yet since you didn't ask me to.

PR review fixes:
1. Archive bug (MUST FIX) — mukurtu_workflows.module now only archives nodes that are currently published; drafts/needs-review content is left untouched. Covered by a new UserCancelArchiveTest.php (3 tests).
2. Dead code in MukurtuBlockUserAction — kept as-is (it's a required interface method and a reasonable defensive fallback), but it's no longer untested: new MukurtuBlockUserActionTest.php exercises both branches directly.
3. Cross-module coupling — extracted relabelCancelMethods() out of FormHooks into a new shared UserCancelMethods.php utility class; both FormHooks and MukurtuUserCancelConfirmForm now call the same neutral utility instead of one reaching into the other's hook-handler class.

Accessibility fixes (all in MukurtuUserCancelConfirmForm.php):
4. Added a real _title_callback on the route so the page's actual heading/browser title reflects the confirm question (verified live via drush php:script — it renders correctly).
5. Wrapped the cancellation-method radios in a <fieldset>/<legend> so the group has a programmatic label (verified in rendered HTML output).
6. Associated the read-only "when cancelling these accounts" text with the submit button via aria-describedby.
7. Left role="alert" on the no-accounts message as-is (harmless) — the real fix is #4, since a correct page title/heading is what actually gets announced to screen readers on load.

Verification: all 58 existing Kernel tests for mukurtu_protocol/mukurtu_workflows still pass, the 2 new test files pass, phpcs --standard=Drupal,DrupalPractice is clean on every file I touched (one pre-existing warning unrelated to my changes), and I confirmed the rendered form HTML directly. No update hooks are needed — nothing here touches active config or schema, only PHP logic and a route definition.

nick-deer (approved):

PR #1783, opened 2026-07-02
Make category caption boxes uniform height in Browse by Category grid

### LINKED ISSUE #661: Browse category thumbnail ratio
- [x] Should use a 3:2 ratio (I believe that's what @pjudge suggested)
- [x] Add auto scale and crop 
- [ ] Adjust all text boxes for a uniform grid look. I guess they need to extend up, not down on line breaks? Or a similar result. Image

Shown on front page and /categories

michael-wynne-wsu:
## Summary
- Fixes the remaining checklist item on #661: category caption boxes in the "Browse by Category" grid (homepage + `/categories`) now render at uniform height across a row instead of short-name boxes staying short while a wrapping long name grows its box taller.
- Caption box now fills the stretched grid-cell height and vertically centers the category name, reusing the existing fill-remaining-space pattern from `featured-items`.
- CSS-only change in `themes/mukurtu_v4/components/04-templates/categories/_categories.scss` (compiled `style.css`/`.map` included).

Closes #661

## Test plan
- [x] Verified on homepage "Browse by Category" section: short name ("Lands") and long wrapping name ("This is a category with a much longer name") render with equal-height caption boxes, text centered
- [x] Verified same on `/categories` page
- [x] Checked mobile (1-col) and desktop (2/4-col) breakpoints — no regressions
- [x] Confirmed 3:2 thumbnail image crop unaffected
- [x] Ran WCAG accessibility check — no concerns (no markup changes, focus/reading order unaffected)
- [x] Ran PR review — no blocking findings


michael-wynne-wsu:
## PR Review: Uniform-height category caption boxes (Issue #661)

**Blast Radius:** LOW — CSS-only change, scoped selectors, single theme component
**Security:** No findings (no injectable/executable code)
**Tests:** No test suite exists for SCSS in this project — N/A, not flagged
**Breaking Changes:** None

**Scope check:**
- `.category-teaser` is used in exactly one Twig template (`taxonomy-term--category.html.twig`), so no other markup is affected.
- `.category-grid`/`.category-card` only exist for the `mukurtu_categories` Views display (homepage block + `/categories` page), both intentional targets per the issue.
- The `.field--name-name` rule is nested under `.category-teaser`, so it can't leak into the unrelated `.communities__item .field--name-name` rule elsewhere in the compiled CSS — no cross-component collision.
- Committing the compiled `css/style.css`/`.map` alongside the `.scss` source matches this repo's established convention (confirmed via `git log` on prior category/theming commits, e.g. `574b68b9`, `55e85029`).

--- MUST FIX (Blocking) ---
None.

--- SHOULD FIX (Non-blocking) ---
None.

--- SUGGESTIONS ---
None — the change reuses the existing `flex: 1` / `block-size: 100%` fill-remaining-space pattern already established in `featured-items`, keeping the codebase consistent rather than inventing a new approach.

--- LOOKS GOOD ---
- Minimal, targeted diff (9 lines of SCSS) that fully addresses the remaining issue #661 checklist item.
- No regressions to the already-completed 3:2 ratio / auto-crop work in the same file.
- Verified visually on both the homepage and `/categories` at desktop and mobile breakpoints.

michael-wynne-wsu:
Before

image

After

Screenshot 2026-07-02 at 9 21 42 AM

christa-kavanagh-wsu (commented):
Category styling is looking good, something weird is going on with adding thumbnails. Alt text is not being generated, and I have to try twice to add any images as the category thumbnail. I also got an error with the first image I added. 

Image

michael-wynne-wsu:
is that new to this branch?

it worked completely as expected for me...

christa-kavanagh-wsu:
It's new to this branch; I didn't run into this earlier and it's working as expected on Main.

christa-kavanagh-wsu:
It times out here and doesn't let me add metadata: 

image

Then after I re-input the image it takes me to the metadata, but doesn't include alt text:

image

michael-wynne-wsu:
Check with another browser too, I can't replicate it.

christa-kavanagh-wsu:
Ok, it's just being weird in Firefox. Edge works fine.

christa-kavanagh-wsu (approved):
Works as expected in Edge

joelle-birano-wsu (approved):
Works as expected for me

michael-wynne-wsu:
Okay the claude inspection is actually interesting. TLDR it's a weird fringe case, NOT from this branch, and I'll spool it up on a separate branch/PR.

Both investigations are done, and neither error is caused by PR #1783 (which only touched SCSS/CSS).

Error 1 — Media library array_filter(null) crash:
This is a known Drupal core edge case in MediaLibrarySelectForm::viewsFormValidate() — it doesn't guard against a NULL value when the "widget_table" view returns zero rows. field_thumbnail_image on the Category vocabulary is restricted to image media only, and this Tugboat preview was built via a fresh drush site-install with zero media items of any kind. Any fresh Mukurtu site would hit this the first time an admin opens that media browser before any image has ever been uploaded — it's environment/data-state, not a regression. That explains "doesn't affect everyone": it only bites on a site with zero existing image media.

Error 2 — getEntityTypeId() on null on /admin/categories/manage:
This one's a real, pre-existing bug, unrelated to the editorial workflow merge or PR #1783. MukurtuCategoryManageController::content() (and its twin MukurtuManageTaxonomyController) call \Drupal::formBuilder()->getForm('Drupal\taxonomy\Form\OverviewTerms', $vocabulary) — passing the class name as a string plus the vocabulary as an extra argument. That pattern worked when OverviewTerms was a plain FormBase, but in current Drupal core it's an EntityForm, which requires the entity to be set via getFormObject()->setEntity() before getForm() is called (the same correct pattern already used a few lines below for the "Add category" form). This code dates to a 2022 commit and has been dormant until the profile's core version caught up to the refactored OverviewTerms. It'll fail identically on any environment running current core, not just Tugboat.

Since neither issue stems from PR #1783, they shouldn't block that PR. The second one is a legitimate standalone bug worth fixing.

christa-kavanagh-wsu:
Oh that's interesting

nick-deer (approved):

PR #1784, opened 2026-07-02
Fix taxonomy overview form fatal error and media library null-selection crash

michael-wynne-wsu:
## Summary
Two unrelated pre-existing bugs were surfaced while testing PR #1783 (CSS-only) on its Tugboat preview. Neither is caused by that PR; both are fixed here on a separate branch.

- **`MukurtuCategoryManageController`/`MukurtuManageTaxonomyController`**: fixed a fatal error (`Call to a member function getEntityTypeId() on null`) on `/admin/categories/manage` and `/admin/taxonomy/{vocab}/manage`. Both controllers invoked `Drupal\taxonomy\Form\OverviewTerms` using the old `FormBase` calling convention, but the class is now an `EntityForm` requiring `setEntity()` before `getForm()`, and also takes the vocabulary as an explicit `buildForm()` argument. Reproduces on any environment running current Drupal core, not just Tugboat.
- **`mukurtu_media.module`**: added a defensive `#validate` guard for a Drupal core defect in `MediaLibrarySelectForm::viewsFormValidate()`, which calls `array_filter()` on a form value that stays `NULL` (instead of `[]`) when the media library view returns zero rows — e.g. opening the widget for a field/bundle with no existing media yet. Patching core directly isn't viable across `composer update`, so the fix normalizes the value via `hook_form_alter()` before core's handler runs.

## Test plan
- [x] Added Kernel tests (`modules/mukurtu_taxonomy/tests/src/Kernel/TaxonomyManageControllersTest.php`) covering both controllers
- [x] Added Unit tests (`modules/mukurtu_media/tests/src/Unit/MediaLibrarySelectFormGuardTest.php`) covering the guard logic
- [x] Verified live in DDEV: `/admin/categories/manage` and `/admin/taxonomy/{vocab}/manage` render correctly (term list, drag-and-drop, add-term/category forms)
- [x] Verified live in DDEV with a genuine A/B test: disabling the media library guard reproduced the exact reported `TypeError` (confirmed via watchdog log), re-enabling it eliminated it on the same flow (opening media library for a zero-item bundle, clicking "Insert selected" with nothing selected)
- [x] Ran PR review — no blocking findings
- [x] No update hooks needed (no schema/config changes)


michael-wynne-wsu:
See if this resolves those category-flagged issues.

christa-kavanagh-wsu (commented):
I am seeing the same behavior I saw in the categories PR when trying to add media assets.

michael-wynne-wsu:
I can't replicate it in any of my browsers and there's nothing in site logs. Maybe send a video recording?

joelle-birano-wsu (approved):
This seems fine to me.

michael-wynne-wsu:
Given that this doesn't seem consistent or replicable, I don't know that I want to sink more time into it unless a clearer pattern emerges.

PR #1789, opened 2026-07-04
Add custom access denied (403) page

michael-wynne-wsu:
## Summary
- Replaces Drupal's generic 403 message with a themed, full-width "Incorrect Permissions" page at `/mukurtu/access-denied`, styled to match Basic Page layout.
- Adds a settings form (`/admin/config/mukurtu/access-denied`, linked from Dashboard > Site settings) so Mukurtu Managers can edit the page's title and message, gated by a new `administer mukurtu access denied page` permission.
- The page's "log in" link carries a `destination` back to the originally requested page, so users land back where they started after authenticating (leverages Drupal core's existing `RedirectDestination` service and this repo's existing login-redirect logic).
- Includes `mukurtu_core_update_40072` to roll the change out to existing sites (sets `page.403`, grants the new permission to Mukurtu Manager, installs default config).

Closes #241

## Test plan
- [x] Visit a protected page while logged out; confirm HTTP 403, "Incorrect Permissions" title, full-width styled layout, and the fixed "log in and try again" sentence.
- [x] Click the "log in" link, authenticate, confirm redirect back to the originally requested page (not the front page).
- [x] As a Mukurtu Manager, edit the title/message via the settings form and confirm the access denied page reflects the change while the fixed sentence stays constant.
- [x] Confirm a non-manager gets 403'd on the settings form route itself (no redirect loop).
- [x] Run `drush updb` on an existing site and confirm `mukurtu_core_update_40072` applies cleanly and is idempotent on a second run.
- [x] Compiled SCSS via `npx gulp buildSass`, verified visually in DDEV.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks.
- [x] I have run an accessibility check, and resolved any issues.
- [x] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [x] I have updated from `main` and resolved any merge conflicts.
- [x] I have verified that all expected checks pass.
- [x] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1791, opened 2026-07-06
Constrain and center content on the Page not found (404) page

michael-wynne-wsu:
## Summary
- The 404 page fell back to the generic `page.html.twig`, so its body text rendered full-width and left-aligned instead of matching the site's standard centered content column.
- Added `page--404.html.twig` (Drupal's built-in `page__404` theme suggestion) that adds a `layout-content--404` modifier class to the content wrapper.
- Styled that class in `_page.scss` using the existing `layout--full-width-grid` mixin (same pattern already used for `.page__account-menu` and `.region__highlighted`), constraining width to the standard ~1280px centered column and centering the text.
- Compiled SCSS via `npx gulp buildSass`.

Scope note: the 403 "Access denied" page has the same underlying issue but is already being addressed in #1789 (not yet merged), so it's intentionally left untouched here.

## Test plan
- [x] Verified locally on DDEV: visited a nonexistent URL, confirmed the new `layout-content--404` class renders and the "requested page could not be found" text is centered and width-constrained to match the "Page not found" heading above it
- [x] Confirmed compiled `style.css` includes the new rule
- [ ] Spot-check a couple of other pages (basic page, front page) in review to confirm no regression


michael-wynne-wsu:
Minor change, just updating 404 page layout.

nick-deer (approved):

PR #1794, opened 2026-07-07
Add theming, privacy badges, and pretty URLs for personal collections

### LINKED ISSUE #1523: Personal collection styling
Needs work

Image

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1523

## Updates

- [ ] User's personal collections page should have same styling as regular collections page.
- [ ] User's personal collections should have same styling as regular collections, with the addition of a public/private marker.
- [ ] Personal collection URLs are now prettified.

## Summary

Personal collections (`/user/personal-collections` overview, `/personal-collection/{id}` individual pages) had no working theme hook, no full/teaser templates, and no path aliases, so they looked inconsistent with the rest of the site.

- Wires up a proper view builder for the `personal_collection` entity (mirroring the existing Protocol/Community pattern), which was required before any custom templates could render at all.
- Adds full and teaser templates that reuse the existing `/collections` styling (`.collection`/`.collections` classes) so personal collections match the site-wide look.
- Adds a small Private/Public badge, shown on both the full page and listing rows.
- Shows a thumbnail image in the "My Personal Collections" listing (previously hidden).
- Registers `personal_collection` with Pathauto so it gets pretty URLs (`/personal-collection/my-collection-name`) like every other content type, and backfills aliases for existing personal collections via an update hook.
- Fixes an unrelated pre-existing Drupal core quirk (`enable_page_title_template`) that was causing the personal collection page title to render twice with raw field markup.

## Known issue

The accessibility check flagged that the "Private" badge variant (`--color-warning-dark` on `--color-warning-light`) measures 4.14:1 contrast, just under the 4.5:1 AA minimum for normal-size text. The "Public" variant passes at 6.24:1. Flagging for follow-up rather than blocking this PR — see `themes/mukurtu_v4/components/01-atoms/privacy-badge/_privacy-badge.scss`.

## Test plan

- [ ] Run `ddev drush updb` and confirm the three `mukurtu_collection` update hooks apply cleanly
- [ ] Visit `/user/personal-collections` and confirm the listing matches `/collections` styling, shows thumbnails, item counts, and Private/Public badges
- [ ] Visit an individual personal collection page and confirm the full-page template renders (image header, summary, description, items grid, privacy badge)
- [ ] Create a new personal collection and confirm it gets a pretty URL automatically
- [ ] Confirm an existing personal collection (created before this change) also resolves at its new alias after `updb`
- [ ] Check the Private badge contrast issue noted above


joelle-birano-wsu:
- [x] public/private marker.
- [x] Personal collection URLs are now prettified.
- [ ] User's personal collections page should have same styling as regular collections page.

Styling is quite different: 

My-collection-Mukurtu-CMS-07-10-2026_04_04_PM

michael-wynne-wsu:
Good catch. I had just been looking at `/user/personal-collections`

michael-wynne-wsu:
## Accessibility Review (WCAG)

Commit 99fd0f2a5 only touches config (view-mode switch), no Twig/CSS changes.
It routes personal-collection items through the same node--grid-browse.html.twig
+ vertical-card component already used by the regular Collection page, so no
new template/CSS defects are introduced.

Confirmed sound:
- Image-wrapper link uses aria-hidden="true" tabindex="-1", correctly removing
  it from the a11y tree/tab order; the <h2> title link is the single accessible
  entry point per card.
- No outline/focus suppression in _vertical-card.scss.
- Card title uses sitewide color tokens — no new contrast risk.

Pre-existing gaps newly exposed (not introduced by this commit):
1. Image alt text is now user-facing on personal collections (previously bare
   text links had no image). Personal collections are individual user uploads,
   less editorially reviewed than community-vetted content, so poor/placeholder
   alt text has more surface area now.
2. "Content in Collection" label renders as a plain <div>, not a heading, while
   each card has its own <h2> — screen-reader users browsing by heading get a
   flat list with no parent heading announcing the group. Affects the existing
   Collection page too; personal collections now inherit it.

## PR Review: Render personal collection items as grid cards instead of bare links (99fd0f2a5)

Blast Radius: LOW — single config value + one guarded update hook, scoped to
personal_collection.personal_collection.full display only.
Security: No findings.
Tests: No test files changed — consistent with this module's existing convention.
Breaking Changes: None — display-only change, no schema/API/field changes.

MUST FIX: none
SHOULD FIX: none

SUGGESTIONS:
- Update hook has no explicit rollback note, but matches the established
  pattern in mukurtu_collection.install (_40009/_40013) — not a defect.

LOOKS GOOD:
- isNew() guard before saving, matching convention.
- Update hook number (_40014) correctly follows the last one (_40013).
- Change is minimal and scoped.

## Follow-up found during manual testing

personal-collection--full.html.twig renders field_items_in_collection as a
sibling OUTSIDE .full-node__grid / .full-node__main / .full-node__content,
so it never gets the grid-column width rules the regular collection page's
version gets (there it's part of group_primary_fields, rendered INSIDE
.full-node__content). This caused the "content width is wrong" bug reported
during Tugboat testing — a pre-existing template issue this branch's
grid_browse change made visible. Fixed locally by moving the field print
inside .full-node__content; awaiting confirmation before committing.

michael-wynne-wsu:
Okay I've got it set so that it matches regular collections properly. But the content previews aren't really styled in EITHER kind of collection here - those are addressed here: https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1811

I think we can move this PR along and just confirm that personal collections get caught in the other updates after.

joelle-birano-wsu (approved):
I think I misunderstood the first time around. This looks good to me.

nick-deer:
On the full view, I'm seeing the personal collection image extending past the banner, and it has a title of "Personal Collection Image", anyone else?
image

michael-wynne-wsu:
Good catch Nick, I am removing that field__label now

michael-wynne-wsu:
@nick-deer that is resolved

nick-deer:
Ok I'm seeing the fix applied on my end 👍 one last thing I noticed was when I was creating a personal collection image from scratch in the form, the protocols sync option was auto-selected, even though personal collections are not protocol-controlled. I was able to select protocols once I toggled it off, but it was confusing.

michael-wynne-wsu:
@nick-deer oh that's a really good flag too. Not all authenticated users will have protocol creation permissions. Hmmm, do we disable images on personal collections, or allow uploaded images with no protocols... @christa-kavanagh-wsu @joelle-birano-wsu any opinions there?

nick-deer:
@michael-wynne-wsu hmm, both options have downsides... having public images seems reasonable, but the way we've set up the architecture for images, they're all protocol-controlled by default. I'm not sure what it would take to get around this. On the other hand though, removing the collection image might be bad UX, since it helps differentiate collections from one another at a glance...

joelle-birano-wsu:
> @nick-deer oh that's a really good flag too. Not all authenticated users will have protocol creation permissions. Hmmm, do we disable images on personal collections, or allow uploaded images with no protocols... @christa-kavanagh-wsu @joelle-birano-wsu any opinions there?

How do thumbnails on media assets work? Because we don't directly apply protocols to them, and they don't appear in the media library as standalone assets. Could we do something similar in this scenario?

michael-wynne-wsu:
Thumbnails are part of the media asset structure, not separate entities, so probably not there.

christa-kavanagh-wsu:
> > @nick-deer oh that's a really good flag too. Not all authenticated users will have protocol creation permissions. Hmmm, do we disable images on personal collections, or allow uploaded images with no protocols... @christa-kavanagh-wsu @joelle-birano-wsu any opinions there?
> 
I am in favor of having some kind of visual cue to differentiate the collections, like Nick said it makes them easier to find and just a little nicer visually. There are a couple of ways this could go:
1. Allow uploaded images with no protocols that are somehow tied to a user account? Not sure how that would work, though site-wide you don't have to apply protocols to things like the hero image, logo, or placeholder images (`/admin/config/mukurtu/media-settings`), so maybe that could be a path?
2. Automatically apply the image or thumbnail from whatever media asset is arranged first in the collection (kind of like when you create a playlist on Tidal or somewhere and don't provide a custom image it generates one from the first four song album covers, though maybe one image would be easier). I am not sure how complicated this would be to make happen in practice.

christa-kavanagh-wsu (approved):
The image question seems to need more discussion but everything else looks good.

nick-deer (approved):

michael-wynne-wsu:
> Automatically apply the image or thumbnail from whatever media asset is arranged first in the collection (kind of like when you create a playlist on Tidal or somewhere and don't provide a custom image it generates one from the first four song album covers, though maybe one image would be easier). I am not sure how complicated this would be to make happen in practice.

That is actually v3 collection functionality, we dropped it for time constraints I believe. 

I think for now, we merge this and address in a follow up issue. I really don't want to leave any open code at the moment.

PR #1797, opened 2026-07-08
Fix dictionary word comments not rendering

### LINKED ISSUE #1795: Enable comments on dictionary words

michael-wynne-wsu:
## Summary
- Fixes #1795: comments were already fully configured on Dictionary Words (field, form widget, protocol-aware access control) since 2022, but the full-page template referenced `content.group_primary_fields.comment`, a path that was never populated because `comment` was missing from `group_primary_fields`'s children list — so the comment thread/form silently never rendered.
- Adds `comment` to `group_primary_fields.children` in `core.entity_view_display.node.dictionary_word.full.yml`, matching the pattern already used by Collection and Digital Heritage.
- Adds a defensive, idempotent update hook (`mukurtu_dictionary_update_40032()`) that creates the comment field/display config on `dictionary_word` if it's ever missing, protecting any site that installed `mukurtu_dictionary` before the field was added to `config/install` in 2022.

## Pre-merge checklist:

- [ ] I have recompiled SCSS if required. (not required — no CSS/SCSS changes)
- [ ] I have updated or created CI/CD tests, if required. (not required — config + update hook only)

## Test plan
- [x] Ran `drush updb` in DDEV, confirmed `mukurtu_dictionary_update_40032` completes and correctly appends `comment` to the active `group_primary_fields.children` config.
- [x] Created a temporary test Dictionary Word node and confirmed the "Add new comment" form now renders on its full page (previously absent), then cleaned up the test content.
- [ ] Manual review of comment placement/accessibility on a real Dictionary Word page (checklist items above intentionally left unchecked for reviewer follow-up).


michael-wynne-wsu:
@joelle-birano-wsu dictionary words now have comments showing here.

joelle-birano-wsu (approved):
Looks good!

michael-wynne-wsu:
## Automated review: accessibility check + pr-review-expert

**Blast Radius:** LOW - single content-type config change, backfill hook, no shared library/API/schema touched.
**Security:** No findings.
**Breaking Changes:** None detected.
**CI:** All checks passing.

### Accessibility check (WCAG)

No findings against this diff. It's config-only (2 files, no Twig/CSS/JS), and it surfaces Drupal's already-vetted comment UI (semantic `<article aria-label>`, visible focus states, AA-passing contrast) on `dictionary_word` pages - the same rendering path already used on Collection/Person/Place/Word List/Digital Heritage.

Two pre-existing, sitewide characteristics worth noting as FYI (not regressions from this PR, not blocking):
- The "Comments" field label renders as a `<div class="field__label">` rather than a heading, so there's no `<h2>`/`<h3>` marking the comments section on any bundle that uses this formatter.
- Core's default "Reply"/"Edit"/"Delete" links have no per-comment distinguishing text, so a screen-reader user browsing by link list hears ambiguous repeated "Reply" links. This is a Drupal core limitation, not Mukurtu-introduced.

### PR review

**LOOKS GOOD**
- Root cause correctly identified: `comment` was missing from `group_primary_fields.children`, so the template path `content.group_primary_fields.comment` never populated.
- Verified every value the update hook (`mukurtu_dictionary_update_40032`) backfills (field config, form-display widget, view-display formatter, children append) matches the live `config/install` YAML exactly.
- `FieldConfig::create()` omits `field_type`, which is fine since Drupal sets it from the field storage definition (`field.storage.node.comment`, type `comment`) in `preSave()`.
- Follows the established idempotent-backfill pattern already used in `mukurtu_core.install` and `mukurtu_notifications.install` (`loadByName` guard + `create()`).
- Update hook number (`40032`) correctly increments from the last one in this file (`40031`), no collision risk.

**SUGGESTION (non-blocking)**
- `mukurtu_dictionary_update_40032()`: the `group_primary_fields` children append is gated by `if ($children && !in_array('comment', $children))`. If a site has no `group_primary_fields` config at all, `$children` resolves to an empty array, which is falsy, so the append is silently skipped. In that edge case the comment field would still render on the view display since that part isn't gated the same way, just outside the group. Worth a one-line comment or explicit check to make the intentional skip clear to future readers - not a functional bug.

---
Generated with the `wcag-accessibility-compliance` and `pr-review-expert` skills via Claude Code.

nick-deer (approved):

PR #1798, opened 2026-07-08
Hide export-list flag button on inconsistent content displays

michael-wynne-wsu:
## Summary
- The "Add to export list" flag button (`flag_export_content`) was shown, unhidden, on 16 view display configs across Collection, Dictionary Word, Word List, Digital Heritage, Person, Place, and Landing Page — while every other view mode for these content types already hides it. This was noticed while testing #1797 (a red, unstyled-looking button appeared in the sidebar of a Dictionary Word page).
- Confirmed this is a pre-existing, sitewide inconsistency, not something introduced by #1797 or scoped to a single content type.
- Moves `flag_export_content` into each affected display's `hidden` list (and removes it from `field_group` children where it was nested), matching the pattern already used everywhere else in the codebase.
- Adds defensive, idempotent update hooks in each of the 6 affected modules (`mukurtu_collection`, `mukurtu_dictionary`, `mukurtu_digital_heritage`, `mukurtu_landing_page`, `mukurtu_person`, `mukurtu_place`) so existing sites pick up the fix.
- Note: the legitimate "Export" / "Add to export list" admin local-task buttons (visible to admins in the page header, a separate feature from this flag) are untouched.
- Note: `mukurtu_dictionary_update_40033` is numbered assuming #1797's `mukurtu_dictionary_update_40032` merges first — may need renumbering depending on merge order.

## Pre-merge checklist:

- [ ] I have recompiled SCSS if required. (not required — no CSS/SCSS changes)
- [ ] I have updated or created CI/CD tests, if required. (not required — config + update hooks only)

## Test plan
- [x] Verified via `drush` that all 16 display configs (`content`/`hidden` and `field_group` children) show the expected before/after state.
- [x] Ran the update hooks in DDEV (`drush updb` for 4 modules; manually invoked the other 2 whose local schema version was ahead of what's in `main`, a pre-existing DDEV/DB sync quirk unrelated to this change) — all completed without error.
- [x] Created a temporary test Dictionary Word node and confirmed the red flag button (`flag flag-export-content` markup) is completely gone from the rendered page, while the legitimate admin "Export"/"Add to export list" local-task links remain. Cleaned up test content afterward.
- [ ] Manual accessibility/PR review (checklist items above intentionally left unchecked for reviewer follow-up).


michael-wynne-wsu:
Both checks are done. Here's the combined review of PR #1798 (https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1798).

## Accessibility check — clean, nothing actionable

- Setting flag_export_content to hidden removes it from the render array entirely (not CSS display:none), so there's no leftover focusable-but-invisible control and no orphaned aria-* reference — a real WCAG 2.4.3/2.4.7 concern the *other* way would be if it had been hidden via CSS instead.
- Removing it from the 5 affected field_group.<group>.children arrays only strips that one entry; each group's own label/region/parent structure and all sibling fields are untouched, so no group/fieldset accessibility structure is orphaned.
- The 6 update hooks only touch config objects — no template/CSS/JS — so they have no runtime accessibility effect.
- Net effect is a consistency improvement (an unstyled/inconsistent button no longer appears on some displays but not others), not a regression.

## PR review — one real blocking risk, rest is solid

MUST FIX (blocking) — update hook numbering race with PR #1797 (https://github.com/MukurtuCMS/Mukurtu-CMS/pull/1797):
Both PRs branch from the same point (last hook on main is mukurtu_dictionary_update_40031) and each independently appends the next one:
- #1797 (open, not yet merged) adds mukurtu_dictionary_update_40032
- #1798 adds mukurtu_dictionary_update_40033, explicitly assuming #1797 merges first

Drupal's update system only runs a hook if its number is greater than the module's last-recorded schema version — it does not backfill skipped numbers. If #1798 merges and is deployed (updb) to a site before #1797, that site's mukurtu_dictionary schema version jumps straight to 40033. When #1797 later merges and that site updates, 40032 will never run on it, silently leaving dictionary-word comments broken (the exact bug #1797 fixes). Function names don't collide (40032 vs 40033), so this won't cause a PHP fatal — it's a silent skipped-update risk, not a merge conflict.

Fix: merge #1797 before #1798 (and make sure updb runs on any environment between the two merges), or renumber before merging if order can't be guaranteed. This is the same risk the PR author already flagged in the description — confirmed real and currently live since both PRs are still open.

Correctness/idempotency of the 6 new update hooks: all good.
- Each mirrors its YAML diff exactly (same displays, same field_group names, NULL used correctly where no group exists).
- isNew() guard, isset() before unsetting, array_filter on children — all safe to re-run.
- Minor nitpick (non-blocking): $config->set('hidden', ...) and $config->save() run unconditionally even when nothing actually changed, causing a redundant config-save/cache-invalidation on repeat runs. Harmless, just slightly wasteful — not worth blocking on.

Consistency across the 16 YAML diffs: verified all 16 follow the identical pattern (remove content.flag_export_content, add hidden.flag_export_content: true, strip from field_group children only where it was actually nested). No stragglers or inconsistencies found.

Other: no security issues (pure config API, no raw queries/injection surface), no test-coverage gap beyond what the PR itself already calls out as not required.

Bottom line: accessibility sign-off is clean; the one thing worth resolving before merge is the #1797/#1798 update-hook merge-order dependency.

nick-deer (approved):

Issue #128, opened 2023-11-09
Define user pages


michael-wynne-wsu:
user pages (e.g., /user, /user/username

steve-taylor-wsu:
Redirect /user to /dashboard

michael-wynne-wsu:
/user/username

optional public profile
open personal collections

steve-taylor-wsu:
On /user/{uid} show public personal collections, public info if selected.

Review who has access to user accounts.
Review how we can limit access to only the fields we want.

michael-wynne-wsu:
I am actually not very inclined to have much if anything in the way of a profile page. It needs to be approached carefully.

michael-wynne-wsu:
All authenticated should have /dashboard - eg for my collections

PR #1802, opened 2026-07-09
Define user pages: redirects, profile access, personal collection sharing

michael-wynne-wsu:
Implements #128.

## Summary
- Anonymous visitors hitting `/user` are now redirected to the login form instead of the generic access-denied page; logged-in behavior is unchanged.
- Authenticated users can now view other users' profiles (previously only the owner and admins could).
- Profile URLs move from `/user/{uid}` to a display-name-based path via a new Pathauto pattern, with a uniqueness constraint on the display name field.
- The default profile display drops a dead "Email users" flag component (the underlying flag doesn't exist in Mukurtu) and adds a "Personal collections" section: a profile shows the owner's own public + private personal collections, and other logged-in users see only the public ones.
- Public, published personal collections are now viewable by anonymous visitors via direct link (granted the `anonymous` role the existing `view published personal collection entities` permission), matching the privacy field's own description ("Anyone with a link to a public personal collection can see it").
- Added update hooks for all permission/config changes so existing sites pick them up via `drush updb`.

## Test plan
- [ ] `drush updb` runs cleanly on an existing site and applies the new update hooks.
- [ ] Anonymous visiting `/user` redirects to `/user/login?destination=/user`.
- [ ] Logged-in user visiting `/user` still redirects to their own profile.
- [ ] A plain authenticated user can view another user's profile (200), but not edit it (403).
- [ ] Anonymous cannot view any user profile (403).
- [ ] A user's profile shows their own public + private personal collections; another logged-in user sees only the public ones; anonymous sees no collections section on the profile.
- [ ] A public, published personal collection is viewable by an anonymous visitor at its normal URL; a private one is not.
- [ ] Changing a display name updates the profile's URL alias; two users with the same display name don't collide (form validation error and/or `-1` suffixed alias).
- [ ] Automated tests pass: `UserProfilePageTest` (functional) and `PersonalCollectionEntityAccessTest` (kernel).


michael-wynne-wsu:
Needs more planning.

PR #1808, opened 2026-07-09
Add grid/list/map view switching to the collection page

### LINKED ISSUE #1315: Collections - add list and map view
From Matt 

>Not much code wise is transferable from Browse Content. That uses a view with a display for each of Grid | List | Map. Here we're dealing with a single entities entity reference field. Grid and List would be straightforward as those are just view modes. Map would need its own view to filter on the items in the collection. You'd need new Javascript to handle toggling between them like Browse Content as well. Also worth thinking about whether we remember the users selection on a per Collection basis, or globally, and would that be a shared preference with Browse Content, eg. if I select Map while at /browse, should that be remembered and used when looking at a Collection?

michael-wynne-wsu:
## Summary

- Adds a grid/list/map view switcher to `/collection/{collection-name}`, matching the existing pattern on `/browse` and `/digital-heritage` (#1315).
- Filters the switcher to the current collection's referenced items via a new custom Views argument plugin (`field_items_in_collection` is a base field with no auto-generated reverse Views relationship), and preserves the curator-defined item order.
- Reuses the existing `mukurtu_browse/mukurtu-browse-view-switch` JS and `.browse-links`/`.active-toggle` styling unmodified.
- Includes `hook_update_40010` to roll the new view/display config out to existing sites.

Screenshot 2026-07-09 at 10 48 05 AM

## Pre-merge checklist:

## Test plan
- [x] Verified in DDEV (as an authenticated user) that the grid/list/map switcher renders and correctly scopes to a test collection's items, in curated order
- [x] Verified the aggregate Leaflet map plots only geolocated items with correct coordinates, and popups link to the item
- [x] Verified the switcher JS/library loads and the `browse-content list/grid/map` CSS classes are present
- [ ] Manual click-through of the toggle buttons in a live browser (not done in this session — no browser tool available; recommend a quick manual pass before merge)


joelle-birano-wsu (approved):
Works as expected!

christa-kavanagh-wsu (approved):
This works as expected!

nick-deer (approved):

Issue #1624, opened 2026-05-13
Review all teasers and previews


michael-wynne-wsu:
## Locations

- [ ] `/browse`
   - [ ] List
   - [ ] Grid
   - [ ] Map
- [ ] `/digital-heritage`
   - [ ] List
   - [ ] Grid
   - [ ] Map
- [ ] `/dictionary`
   - [ ] Default
   - [ ] Map
- [ ] Content in collections
   - [ ] List
   - [ ] Grid
   - [ ] Map
- [ ] Words in word lists
   - [ ] Default
   - [ ] Map
- [ ] Taxonomy term pages
   - [ ] List
   - [ ] Grid

## Content types

- [ ] Digital heritage items
- [ ] Dictionary words
- [ ] Person records
- [ ] Place records
- [ ] Collections
- [ ] Word lists

## Media types

- [ ] Audio
- [ ] Document
- [ ] External embed
- [ ] Image
- [ ] Remote video
- [ ] Soundcloud
- [ ] Video

## Modes

- [ ] List
   - [ ] Media asset
   - [ ] Title/name
   - [ ] Summary
   - [ ] Community and protocol
   - [ ] content type (if relevant to the page)
   - [ ] category (only for DH items)
   - [ ] Collections or word lists that the content is in
- [ ] Grid (this might be too much)
   - [ ] Media asset
   - [ ] Title/name
   - [ ] Summary
   - [ ] Community and protocol
   - [ ] content type (if relevant to the page)
   - [ ] category (only for DH items)
   - [ ] Collections or word lists that the content is in
- [ ] Map (this is probably too much)
   - [ ] Media asset
   - [ ] Title/name
   - [ ] Summary
   - [ ] Community and protocol
   - [ ] content type (if relevant to the page)
   - [ ] category (only for DH items)
   - [ ] Collections or word lists that the content is in
- [ ] Dictionary default (this is already well-defined and probably does not need any modification)

michael-wynne-wsu:
@christa-kavanagh-wsu @joelle-birano-wsu am I missing any other places where there are previews, teasers, cards, etc?

PR #1809, opened 2026-07-09
Show video/remote_video/soundcloud media as their native player by default, with manual thumbnail override (#1624)

michael-wynne-wsu:
## Summary

Phase 1 of a multi-phase effort tracked in #1624 (unify content theming across content types, places, and modes). This PR targets media-type consistency in List/Grid/Map cards.

### This addresses ONLY:

- `/browse` and `/digital-heritage` pages.
- List, Grid, and Map views. 
- Tested with digital heritage items.
   - Other content types coming in phase 2.
- Only focused on getting media working.
   - The rest of the preview content and appearance is in phase 2.

## Details 

- `video`, `remote_video`, and `soundcloud` media now render their native player/embed by default in these card views, matching what editors expect to see. If an editor has manually uploaded an image via the existing `field_thumbnail` field on that media item, the card shows that thumbnail instead. A new conditional Twig template per bundle (`media--video--browse.html.twig`, `media--remote-video--browse.html.twig`, `media--soundcloud--browse.html.twig`) makes this choice based on `media.field_thumbnail.isEmpty()`.
  - An earlier version of this PR forced these three bundles to always show a thumbnail. That broke `soundcloud` entirely (no thumbnail is ever auto-generated for it, so cards showed nothing) and was the wrong default regardless — fixed based on testing feedback.
- Fixes `digital_heritage`'s Grid mode, which was using a different media view mode (`digital_heritage_sidebar`) than every other content type's Grid mode (`browse`) for its representative media — a real inconsistency bug found during investigation, now aligned with the rest. Confirmed matching between `/browse` and `/digital-heritage` Grid views.
- De-duplicates the repeated `aria-hidden`/`tabindex="-1"` thumbnail-link wrapper markup that was copy-pasted across 11 node card Twig templates into one shared include (`themes/mukurtu_v4/templates/content/_card-media.html.twig`). That include now only wraps bundles that are always static (`image`, `document`, `external_embed`) — `video`/`remote_video`/`soundcloud` are excluded because they can render a live interactive player, and wrapping interactive content in a full-card link would break its controls.
- Adds update hooks `mukurtu_core_update_40073`/`40074` to push the config changes to existing sites.

## Test plan

- [x] Visually confirmed video, remote_video, and soundcloud render correctly (native player by default) in List and Grid views on `/browse` and `/digital-heritage`.
- [x] Visually confirmed the same in Map view.
- [x] Confirmed `/digital-heritage` Grid mode now matches `/browse` Grid mode's representative-media rendering for the same content.
- [x] Confirmed accessibility/alt text look correct via manual testing.
- [x] Ran the WCAG accessibility skill against the final templates/include — no blocking issues. One non-blocking finding: when an editor manually sets `field_thumbnail` on a video/remote_video/soundcloud item, the resulting thumbnail renders like every other static-image card but isn't wrapped in the clickable overlay link (`_card-media.html.twig` decides the wrap based on bundle alone, one template above where the `isEmpty()` check happens, so it can't see that this particular instance is actually static). Narrow-impact — only affects manually-overridden thumbnails on these three bundles — tracked as a fast-follow rather than blocking this PR.
- [x] Ran a PR self-review pass on both the original and corrected commits — YAML configs and Twig templates checked for internal consistency; no blocking issues found.
- [ ] Spot-check alt text on the underlying thumbnail source images (pre-existing/out of scope for this diff).

## Pre-merge checklist:

- [x] I have run an accessibility check, and resolved any blocking issues (one non-blocking finding noted above, tracked as a fast-follow).

### DIFF
--- a/config/install/core.entity_view_display.media.remote_video.browse.yml
+++ b/config/install/core.entity_view_display.media.remote_video.browse.yml
@@ -8,6 +8,7 @@ dependencies:
   module:
     - image
     - layout_builder
+    - media
 third_party_settings:
   layout_builder:
     enabled: false
@@ -17,7 +18,18 @@ targetEntityType: media
 bundle: remote_video
 mode: browse
 content:
-  thumbnail:
+  field_media_oembed_video:
+    type: oembed
+    label: hidden
+    settings:
+      max_width: 800
+      max_height: 0
+      loading:
+        attribute: eager
+    third_party_settings: {  }
+    weight: 1
+    region: content
+  field_thumbnail:
     type: image
     label: hidden
     settings:
@@ -33,12 +45,11 @@ hidden:
   field_communities: true
   field_cultural_protocols: true
   field_identifier: true
-  field_media_oembed_video: true
   field_media_tags: true
   field_people: true
-  field_thumbnail: true
   flag_export_media: true
   langcode: true
   name: true
   search_api_excerpt: true
+  thumbnail: true
   uid: true
--- a/config/install/core.entity_view_display.media.soundcloud.browse.yml
+++ b/config/install/core.entity_view_display.media.soundcloud.browse.yml
@@ -4,7 +4,7 @@ dependencies:
   config:
     - core.entity_view_mode.media.browse
     - image.style.medium_480px_w_
-    - media.type.remote_video
+    - media.type.soundcloud
   module:
     - image
     - layout_builder
@@ -12,12 +12,12 @@ third_party_settings:
   layout_builder:
     enabled: false
     allow_custom: false
-id: media.remote_video.browse
+id: media.soundcloud.browse
 targetEntityType: media
-bundle: remote_video
+bundle: soundcloud
 mode: browse
 content:
-  thumbnail:
+  field_thumbnail:
     type: image
     label: hidden
     settings:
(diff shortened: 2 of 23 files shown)


michael-wynne-wsu:
## Review summary

**Accessibility check:**
- **Real, worth-knowing finding (non-blocking):** when an editor *does* set a manual `field_thumbnail` on a video/remote_video/soundcloud item, the resulting card shows a plain static image — visually identical to an image-bundle card — but `_card-media.html.twig` only wraps `image`/`document`/`external_embed` in the clickable overlay link. So that specific thumbnail looks clickable (matches every other card visually) but isn't. Root cause: the wrap decision is made one template level above where the `isEmpty()` check happens, so it can't see which branch fired.
- Confirmed no invalid markup, and the audio-style exclusion pattern is correctly applied.
- Flagged (out of scope for this PR, core Drupal behavior) that oEmbed iframes can ship without a `title` attribute if the provider doesn't supply one — worth a future ticket, not something to fix here.

**PR review:** the three YAML configs are internally consistent (formatter types match their declared module dependencies, hidden/content lists are complete with no orphaned fields), and all three new Twig templates mirror the existing audio pattern correctly. No blockers.

The one finding (inert-looking thumbnail for video/remote_video/soundcloud when manually overridden) is a narrow-impact polish item — it only shows up when an editor manually uploads a thumbnail for one of those three types. Want me to fix it now (thread a computed flag through so those thumbnails become clickable like every other card), or leave it as a fast-follow since it's non-blocking and you're otherwise happy with current behavior?

michael-wynne-wsu:
I've tested this pretty decently, but please confirm. Ask me if detail about the scope is needed.

joelle-birano-wsu (approved):
Looks good to me!

nick-deer (approved):

Issue #1807, opened 2026-07-09
Enable Layout Builder on Content types display


christa-kavanagh-wsu:
Enable Layout Builder on content types (DH items, Dictionary words, Person records, Place records, Collections, and Word lists) display.

michael-wynne-wsu:
Just at the template level (eg: customizing one layout for all DH items), or allowing per-node layouts as well?

christa-kavanagh-wsu:
I feel like at the template level seems most reasonable. Allowing per-node layouts seems a bit much for anything apart from _maybe_ collections/word lists (for something like a digital exhibit, though those could be reasonably created using basic pages).

PR #1814, opened 2026-07-10
Guard against Layout Builder breaking content type full views (#1807)

michael-wynne-wsu:
Layout Builder replaces field-keyed content with layout sections, which breaks the custom full-view templates for article, page, dictionary_word, word_list, collection, digital_heritage, person, and place (they hard-code direct field access). Disable the "Use Layout Builder" option on Manage Display for these content types until their templates are made compatible.

Also removes dead hero markup from the landing_page full template that referenced fields which don't exist, and converts the community/protocol sidebar list into a Block plugin as groundwork for a future Layout Builder rollout, without changing current rendering.

## Pre-merge checklist:

### DIFF
--- a/modules/mukurtu_core/src/Hook/FormHooks.php
+++ b/modules/mukurtu_core/src/Hook/FormHooks.php
@@ -1362,4 +1362,46 @@ class FormHooks
             }
         }
     }
+
+    /**
+     * Implements hook_form_FORM_ID_alter() for 'entity_view_display_edit_form'.
+     *
+     * Disables the "Use Layout Builder" option on Manage Display for content
+     * types whose custom full-view templates directly reference fields by
+     * name and are not yet compatible with Layout Builder's section-based
+     * rendering (see issue #1807).
+     */
+    #[Hook("form_entity_view_display_edit_form_alter")]
+    public function formEntityViewDisplayEditFormAlter(
+        array &$form,
+        FormStateInterface $form_state,
+    ): void {
+        if (!isset($form["layout"]["enabled"])) {
+            return;
+        }
+
+        $display = $form_state->getFormObject()->getEntity();
+        if ($display->getTargetEntityTypeId() !== "node") {
+            return;
+        }
+
+        $unsupportedBundles = [
+            "article",
+            "page",
+            "dictionary_word",
+            "word_list",
+            "collection",
+            "digital_heritage",
+            "person",
+            "place",
+        ];
+        if (!in_array($display->getTargetBundle(), $unsupportedBundles, TRUE)) {
+            return;
+        }
+
+        $form["layout"]["enabled"]["#disabled"] = TRUE;
+        $form["layout"]["enabled"]["#description"] = t(
+            "Layout Builder is not yet supported for this content type.",
+        );
+    }
 }
--- a/modules/mukurtu_core/tests/src/Kernel/FormHooksLayoutBuilderGuardTest.php
+++ b/modules/mukurtu_core/tests/src/Kernel/FormHooksLayoutBuilderGuardTest.php
@@ -1,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\mukurtu_core\Kernel;
+
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityFormInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\KernelTests\KernelTestBase;
+use Drupal\mukurtu_core\Hook\FormHooks;
+
+/**
+ * Tests the Layout Builder guard rail on Manage Display forms.
+ *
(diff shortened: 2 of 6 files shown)


michael-wynne-wsu:
@christa-kavanagh-wsu For now just putting in this guardrail, to avoid things breaking now. Future expansion later.

Screenshot 2026-07-10 at 3 45 01 PM

christa-kavanagh-wsu (approved):
This seems to work!

joelle-birano-wsu (approved):

michael-wynne-wsu:
## PR Review: Guard against Layout Builder breaking content type full views (#1807) (#1814)

**Blast Radius:** LOW-MEDIUM — new form alter is scoped to node Manage Display forms; new Block plugin is unused/unwired (disclosed as groundwork); template change only affects `landing_page` full view, confirmed no other consumers of the removed fields/CSS classes.
**Security:** No findings — no user input, no raw queries, access checks preserved identically to the code being duplicated.
**Tests:** Coverage delta: 0 — 143 new lines of logic (form alter + new Block), no new Kernel/unit tests, despite `mukurtu_protocol` having existing Kernel/Functional suites.
**Breaking Changes:** None detected — purely additive/restrictive UI guard, unused new plugin, dead template markup removed.
**CI:** PHP 8.4 build passing, Tugboat passing, Playwright skipping (unrelated), wait-for-tugboat pending.

---

**SHOULD FIX (non-blocking)**

1. **Duplicated rendering logic** — modules/mukurtu_protocol/src/Plugin/Block/CommunityProtocolListBlock.php:60-96
   The block's `build()` copies the entire community/protocol loop verbatim from modules/mukurtu_protocol/src/Hook/CommunityProtocolList.php:61-93. The PR description says it "reuses the same rendering logic," but the code doesn't call into it — it's two independent copies. A future bug fix (e.g. to the access-check fallback, or the cache metadata) applied to one will silently not apply to the other.
   Suggest: make `CommunityProtocolList::buildCommunityProtocolList()` public and inject `CommunityProtocolList` into the Block via DI, or extract the shared logic into a small service both classes call.

**SUGGESTIONS**

2. **No test coverage** for either the new form-alter guard rail or the new Block, though this module already has Kernel tests for similar access/rendering logic (`CulturalProtocolItemTest`, `AccessByProtocolTest`, etc.). A small Kernel test asserting the checkbox is disabled for the 8 listed bundles (and not for others) would guard against regressions here.

3. **Guard rail scope is broader than the stated problem** — modules/mukurtu_core/src/Hook/FormHooks.php:1378-1404 disables Layout Builder for *all* view mode displays of the 8 bundles (checked only by `getTargetBundle()`), but the PR description frames the issue as specifically about *full-view* templates hard-coding field access. If that's intentional blanket caution until other view modes are audited too, worth a one-line note in the PR body; if not, scope with `$display->getMode() === 'full'`.

**LOOKS GOOD**

- Form-alter pattern (`$form['layout']['enabled']['#disabled']`/`#description`) is copied from Drupal core's own approach to disabling this same checkbox (`LayoutBuilderEntityViewDisplayForm.php`), so it composes safely with core instead of fighting it.
- Landing page template cleanup is clean — verified no other `.twig`/`.scss`/`.php` file still references `field_hero_image`/`field_hero_title`/`field_hero_text`/`field_hero_full_width` or the `.landing-hero*` CSS classes.
- New Block plugin matches existing conventions in `mukurtu_protocol` (annotation-based `@Block`, `ContainerFactoryPluginInterface`, constructor promotion) — consistent with `BrowseByCommunityBlock`.
- Access-check-with-label-fallback pattern for protocols/communities is preserved unchanged from the original, so no regression in the sensitive-content visibility logic.

michael-wynne-wsu:
Both review issues are resolved and verified in DDEV (nothing committed — left for your review):

Issue 1 (duplicated logic): CommunityProtocolListBlock.php no longer copies the community/protocol loop — it now injects CommunityProtocolList (via class_resolver, since it's not a registered service) and delegates to its buildCommunityProtocolList() method, which I made public in CommunityProtocolList.php. One source of truth now; a future fix in one place can't drift from the other.

Issue 2 (test coverage): Added two new Kernel test files:
- FormHooksLayoutBuilderGuardTest.php — verifies the checkbox is disabled only for the 8 unsupported node bundles, left alone for other bundles/entity types, and that forms without a Layout Builder element are never touched.
- CommunityProtocolListBlockTest.php — verifies the block's output matches the shared builder exactly, and that access-restricted (strict, non-member) protocol/community pairs correctly fall back to plain labels while open ones render as links.

Issue 3 left as-is (guard rail scope confirmed correct).

Ran both suites plus the full mukurtu_protocol Kernel suite (51 tests) in DDEV — all passing, no regressions. Also ran phpcs (Drupal/DrupalPractice standards) against all four files — clean. Note: while writing the Block test I discovered the base test class's $this->currentUser is UID 1 (superuser, bypasses all access checks), so the test switches to a fresh non-privileged user before exercising access-restricted rendering.

nick-deer (approved):

PR #1816, opened 2026-07-11
Fix "Add to Collection" tab never appearing on content pages (#1011)

michael-wynne-wsu:
## Summary

- Re-implements the "add to collection" and "add to word list" buttons alongside edit/delete/etc.
- All content types get "add to collection". 
- Word lists also get "add to word list".
- Adds NEW modal dialog to select from existing word lists and collections instead of redirecting or using a bigger form. Can also create new ones with minimally required fields.

Screenshot 2026-07-12 at 9 52 26 PM

Screenshot 2026-07-12 at 10 06 34 PM

Screenshot 2026-07-12 at 10 05 24 PM

### Code

- `MukurtuAddItemToCollectionController::isValidCollectionItemBundle()` gated access on loading a `field_config` entity for `field_items_in_collection`, but that field is a base field defined in code (`Collection::bundleFieldDefinitions()`), not a configurable Field UI field. The lookup always returned `NULL`, so access was always denied and the "Add to Collection" tab could never appear on any node, for any user - this is the root cause of #1011.
- Fixed the lookup to use the entity field manager instead, which correctly resolves base/bundle fields.
- Moved the tab's weight so it appears next to "Add to Personal Collection" instead of trailing after Revisions/Devel.
- Added a kernel regression test (verified it fails against the pre-fix code and passes against the fix).
- The "Add to Word List" part of #1011 required no code change - it already works correctly (its access check doesn't have this bug); it's just not visible unless there's an eligible word list that doesn't already contain the word.

Fixes #1011

## Test plan
- [x] Verified live in DDEV: a `digital_heritage` node not yet in every editable collection now shows the "Add to Collection" tab; submitting the form adds it to the selected collection and shows a success message; the tab correctly disappears once the node is in every editable collection.
- [x] Added `CollectionEntityTest::testAddItemToCollectionAccess()`, confirmed it fails on the old code and passes on the fix.
- [x] Ran the full `mukurtu_collection` kernel test suite (18 tests) - all passing.
- [x] Re-checked `node/21` (dictionary word) still shows "Add to Word List" as before - no regression.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required - no config/install or schema changes; logic fix + tab-weight change only, per `docs/update-hooks.md`.)
- [x] I have run an accessibility check, and resolved any issues. (Not applicable - no templates/markup changed.)
- [x] I have recompiled SCSS if required. (Not applicable - no SCSS changed.)
[...]

Issue #620, opened 2024-11-06
Re-implement the add to collection/personal collection contextual icons


michael-wynne-wsu:
V3 has an easy add to collection/personal collection button. Might not be possible with the way collection management has changed though.

![Image](https://github.com/user-attachments/assets/eb13940b-cbc2-4951-bb25-c8e8e20e652a)
![Image](https://github.com/user-attachments/assets/2345f9ce-3fa1-4fb4-a395-d658df2122ae)

michael-wynne-wsu:
That's the star in above v3 screenshots.

michael-wynne-wsu:
We are not doing this for the time being. On full view we're sticking with text buttons, and on list teasers we are placing all actions under one ... menu. Can revisit in the future if there is demand.

PR #1818, opened 2026-07-13
Add quick-action icons to List-mode content cards (#620, #924)

### LINKED ISSUE #924: Theme "add to export list"
No theming done yet.

- [ ] Which pages get it?
   - [ ] Content
   - [ ] Groups
   - [ ] Taxonomies
   - [ ] Other
- [ ] Where is it displayed in all views? 
   - [ ] Maybe the local tasks for the full view?
   - [ ] Browse (grid, list, map)
   - [ ] Other teasers
- [ ] Would an icon be passable?

Image

michael-wynne-wsu:
## Summary

Adds quick-action icons to content cards on List-mode browse listings, addressing #620 ("Re-implement the add to collection/personal collection contextual icons") and #924 ("Theme 'add to export list'"). Both issues came down to the same underlying gap: no consistent quick-action affordance on cards.

- **Overflow menu (⋯ icon)** — Edit, Add to Collection, Add to Personal Collection, Add to Word List, Add to export list, Export. All the "Add to X" actions open as AJAX modals instead of navigating away from the listing, and stay on the listing afterward.
- No action currently renders as a standalone icon button — Add to Collection started that way but moved into the overflow menu alongside the rest, for a more consistent card layout. The underlying `hook_mukurtu_browse_quick_actions_alter()` API still supports an icon-button group for a future action that would warrant the extra prominence.

Scope: List mode only for v1 (the `.horizontal-card` component). Grid and Map are natural fast-follows once this pattern is proven.

**Architecture**: a pluggable `hook_mukurtu_browse_quick_actions_alter()` in `mukurtu_browse`, implemented by `mukurtu_collection` (Add to Collection / Add to Personal Collection), `mukurtu_export` (gives its existing "Add to export list" picker the same AJAX-modal treatment as the other Add to X actions, replacing its prior full-page-with-redirect flow), and `mukurtu_dictionary` (Add to Word List). Collection and word-list eligibility each get a per-user, cached access-check service in front of the existing controllers' per-node queries, so a 20+ row listing doesn't repeat an expensive query on every card. Shared Twig partial + SCSS + JS in the `mukurtu_v4` theme.

**Base branch**: targets `1011-fix-add-to-collection-tab` (#1816) rather than `main`, since this branch's collection-eligibility logic depends on that PR's `field_config` access fix. Since this branch's last sync, three more commits landed on #1816 (accessibility fixes + a dialog-picker regression fix); merging them in surfaced two latent bugs in that refactor (see below), both now fixed here.

**Bugs found and fixed along the way**:
- `mukurtu_collection.routing.yml`'s personal-collection route had `_admin_route: TRUE`, which would have pulled the Gin admin theme into a front-end modal.
- A `%placeholder` in a translated string was corrupting an aria-label attribute (Drupal wraps `%` placeholders in `<em>` markup — wrong for an attribute); switched to `@placeholder`.
- Icon color failed contrast (WCAG 1.4.11) against both shipped palettes; switched `--brand-secondary` to `--brand-primary`.
- A focus-return selector collision when a node has both the icon and overflow collection actions.
- The floating top-right icon position collided with core's contextual-links "Edit" gear icon (same corner, same positioned ancestor); offset ours to clear it.
- The overflow menu showed bullet points from a global `ul li { list-style-type: square; }` base rule that needed a more specific override.
- A cross-module procedural function call inside `hook_module_implements_alter()` (from #1816's refactor) crashed on a full `drush cr` — that hook can run during container compilation, before any module's services (or `\Drupal::$container` itself) are available. Inlined the reorder logic per-module instead of sharing it.
- A `#pre_render` callback (also from #1816's refactor) was passed as a plain function-name string, which Drupal's renderer rejects (`UntrustedCallbackException`) — it must be a closure or a `TrustedCallbackInterface` method. This broke every "Add to X" dialog's AJAX response: icons appeared but clicking one silently failed to open the dialog. Reverted to inline closures.

## Test plan

Verified end-to-end against a live DDEV instance via Playwright (not just unit/kernel level):
- [x] Icons render correctly for a user with eligible actions; absent for anonymous/ineligible users.
- [x] Add to Collection / Add to Word List open as AJAX modals, submit successfully, show a success message, close, and correctly return focus to the triggering icon.
- [x] Overflow menu opens/closes via click, Enter/Space, Escape (with focus return), and click-outside.
- [x] Cache invalidation is correct: an item's icon disappears once it's added to the only collection/word list it was eligible for, and reappears if removed - confirmed across reloads, not just within one page load.
- [x] No console errors introduced (verified against a clean baseline).
- [x] Ran the WCAG accessibility skill and a PR self-review pass; both findings from each were fixed before this PR.
- [x] SCSS recompiled via `npx gulp buildSass` (not `npx sass`).
- [ ] No new automated test coverage added (Kernel/PHPUnit) - verification here was live E2E (Playwright against DDEV), not an addition to the existing test suite. Flagging this gap rather than silently skipping it.

After merging in #1816's latest changes:
- [x] `drush cr` succeeds cleanly (previously fatal - see bugs above).
- [x] Add to Word List, Add to Collection, and Add to export list dialogs each verified via direct simulated AJAX requests (open → submit → response), confirming proper `closeDialog`/`message`/`invoke`(focus) commands and no new watchdog errors. Test data created during verification was cleaned up afterward.

## Pre-merge checklist

- [x] I have checked for and if required, created update hooks. *(None needed - no config/schema changes, only code.)*
- [ ] I have updated or created CI/CD tests, if required. *(See test plan note above.)*
- [x] I have verified that all expected checks pass. *(Pending CI run on this PR.)*


michael-wynne-wsu:
@joelle-birano-wsu @christa-kavanagh-wsu design input needed. Right now for this there is a "add to collection" folder icon, and then all other actions are under the kebab icon. I'm leaning towards scrapping the dedicated add to collection icon and putting everything under one. But I could also see doing more different ones. Any thoughts one way or another?

Screenshot 2026-07-13 at 3 25 16 PM

joelle-birano-wsu:
I'm also leaning towards keeping add to collection with everything else, but I'm curious about what other icons you might add there?

michael-wynne-wsu:
V3 has a star for add to personal collection (it goes from empty to filled when it's in a personal collection), and a phone for mobile sync. Everything else is under an general edit/more actions icons.

But there are 6? 7? possible actions, and I don't think that for the ones we have, an quick shortcut for some but not all helps, nor does 7 icons.

joelle-birano-wsu:
I see. I think it can stay in the menu for now, but if we do add back the star and the phone, I could see it making sense to have that folder as well. They're all about quickly adding items to groups for different purposes.

michael-wynne-wsu:
We'll want to add something when mobile exists again. Otherwise I could see an icon that covers collections/word lists/personal collections, and one that covers export/add to export list/(download, which is a future plan), and then mobile sync. So i think either a grouping like that, or just one menu.

joelle-birano-wsu:
This is working well. Just noticed a couple of things:

- When creating a collection on the fly, a success message says that the collection was created, but not that the item was added to it. It would be good to display that as well - it does display when adding an item to an existing collection.
Screenshot 2026-07-14 093910

- When a collection (personal or site) is selected, all this extra collection metadata is exposed.

Screenshot 2026-07-14 093638

joelle-birano-wsu:
Spotted one more thing: As a protocol member, when adding items to a personal collection, the modal doesn't seem to have a background.

Screenshot 2026-07-14 095151

michael-wynne-wsu:
Okay I believe I've addressed that, and cleaned up a few other things I noted...

joelle-birano-wsu:
The first two issues are resolved. 

Protocol members adding stuff to personal collections is better. Margins are non-existent (idk if styling is that important at this point though but it IS different from admins doing the same thing) and the success message for both adding to an existing personal collection and creating personal collections on the fly are odd.

personal collection form for protocol member:
Screenshot 2026-07-14 at 2 27 58 PM

admin form for reference:

Screenshot 2026-07-14 at 2 30 08 PM

success message for protocol member:
Screenshot 2026-07-14 at 2 24 00 PM

michael-wynne-wsu:
Got messages rendering properly for non admins.

Screenshot 2026-07-14 at 7 19 26 PM

michael-wynne-wsu:
AND got the the spacing, title, and background blur overlay working for non admins, phew.

Screenshot 2026-07-14 at 7 29 11 PM

christa-kavanagh-wsu (changes requested):
- I like the options being under the kebab menu; I think it looks a bit neater. 
- As a protocol steward, the dropdown menu in the modal for both collections and personal collections is transparent.

Image

- The styling of the modal doesn't match the rest of the updated site styling as a non-admin. 

As an admin:

Image

As a protocol steward:

Image

As a language steward:

Image

joelle-birano-wsu:
Overall looking much better!

Besdies what Christa noted, it looks like protocol stewards and curators don't have access to create collections. Same with language stewards and word lists.

Screenshot 2026-07-15 at 12 21 56 PM

michael-wynne-wsu:
FYI I'm not planning to chase down all the small admin/non-admin visual discrepancies, only if they present an actual issue (I know there are more). I don't mind addressed that in future work, but I don't know that it's a blocker for release.

michael-wynne-wsu:
All remaining issues seem resolved to me now.

Screenshot 2026-07-15 at 3 48 16 PM

Screenshot 2026-07-15 at 3 48 34 PM

Screenshot 2026-07-15 at 3 49 01 PM

joelle-birano-wsu (approved):
Looks good to me!

michael-wynne-wsu:
## Final QA pass: accessibility + PR review

Ran a final WCAG 2.2 AA accessibility check and a `/pr-review-expert` pass on this PR after the #1816 merge, per the pre-merge checklist. 2 issues found and fixed (pushed as e334d2ca3 and f2853cbd2); everything else checked out clean.

### Accessibility (WCAG 2.2 AA)
**Fixed:**
- `_quick-actions.scss` - overflow menu link text used `--brand-primary` (~4.0:1 contrast), which clears the 3:1 graphic minimum (1.4.11) but not the 4.5:1 text minimum (1.4.3). Switched to `--brand-primary-dark`.
- `message.theme.js` - AJAX-inserted status/warning messages (e.g. "Added to Collection") had no live-region role, so screen readers never announced them (4.1.3). Only errors got `role="alert"`; now status/warning get `role="status"`, matching core's own mapping.

**Notes (not fixed, out of scope for this PR):**
- `status-messages.html.twig` has the same role gap for server-rendered messages sitewide - pre-existing, worth its own ticket.
- The overflow menu's native `<details>/<summary>` relies on implicit ARIA semantics rather than explicit `aria-expanded`/`role=menu` - a defensible pattern (WAI-ARIA APG "Disclosure"), not a confirmed break.
- The unused `icon_actions` group's `aria-label` construction could reintroduce the already-fixed `%placeholder`-in-attribute bug if a future caller passes one - flagging for the planned Grid/Map fast-follow.

Everything else (icon `aria-hidden`, target sizes, focus-return selectors, dialog focus management) checked out clean.

### PR review
**Fixed (security, MUST FIX):**
- `mukurtu_export.add_node_to_export_list` route/`ExportListAddNodeController` only checked `_permission: 'access mukurtu export'`, with no check that the user could actually view the target node - directly reachable for an arbitrary node ID, including ones behind a protocol the user isn't a member of. Added a `_custom_access` callback requiring `$node->access('view', ...)`. **Requires `drush cr`** after this lands (routing.yml change).

**Notes (not fixed, pre-existing/out of scope):**
- The pre-existing `mukurtu_export.add_item_to_list` route (used by the actual browse-card quick action) and the collection/word-list "add" controllers have the same permission-only gap for arbitrary entity IDs. This PR is what newly surfaces the export one as a one-click link from every card, widening its exposure, but a fix needs broader scoping than this diff - recommend a follow-up ticket.
- The per-user cache tags on collection/word-list eligibility don't cover a brand-new list the user could add to until the cache's bounded lifetime expires - this is called out as an accepted tradeoff in the code's own comments, not an oversight.

**Verified correct:** the two bugs already fixed by this PR (`hook_module_implements_alter()` container-timing crash, `#pre_render` string-callback `UntrustedCallbackException`) are both complete with no remaining instances elsewhere in the codebase. Cache contexts/tags, CSRF posture, and focus-return selector logic all check out.

**Test coverage:** confirmed no new PHPUnit/Kernel tests were added for the new access-control code (`CollectionQuickActionAccessHelper`, `BrowseHooks::browseQuickActionsAlter`, `ExportListAddNodeController`) - verification was manual Playwright E2E only. Worth flagging given this review found a real bug on that exact surface.

christa-kavanagh-wsu:
I was still having the problem Joelle noted where language stewards can't add words to word lists and protocol stewards can't add content to collections. I definitely could yesterday.

Tried clearing the cache and it worked, so it may have something to do with that.

image

nick-deer (approved):

PR #1819, opened 2026-07-13
Fix fatal ZipArchive crash during export packaging

michael-wynne-wsu:
## Summary
- `BatchExportExecutable::package()` assumed the `private://exports/{uid}` destination directory already existed and passed `realpath()`'s result straight into `ZipArchive::open()` with no validation. When that directory didn't exist yet (e.g. a user's first export, or a fresh install), `realpath()` returned `FALSE`, and PHP threw an uncaught `ValueError`, crashing the whole batch with a 500 error.
- The directory is now created up front, and the resolved zip path plus the `ZipArchive::open()` call are validated. On failure, the batch now ends cleanly with a logged error instead of a fatal error page.
- Also stopped swallowing two related silent failures in the CSV exporter (`exportSetup()`'s `prepareDirectory()` and `getOutputFile()`'s `fopen()`) so a broken/unwritable private filesystem is caught early with a clear watchdog message instead of silently producing missing files or only surfacing as the fatal crash above.

## Test plan
- [x] Reproduced the crash on a fresh site where the `exports/{uid}` directory had never been created; confirmed the fix resolves it and produces a downloadable zip.
- [x] Confirmed `php -l` passes on both modified files.
- [x] Verified no config/schema changes, so no update hook is required.
- [x] Ran the pr-review-expert skill against the diff (blast radius, security, correctness); found and fixed one bug caught during review (a PHP "pass by reference" notice from an intermediate fix) before opening this PR.
- [ ] Reviewer: run a normal export end-to-end to confirm no regression in the happy path.

Pre-merge checklist:
- [x] I have checked for and if required, created update hooks. (Not required — no schema/config changes.)
- [ ] I have run an accessibility check, and resolved any issues. (Not applicable — no UI/template changes.)
- [ ] I have recompiled SCSS if required. (Not applicable — no SCSS changes.)

### DIFF
--- a/modules/mukurtu_export/src/BatchExportExecutable.php
+++ b/modules/mukurtu_export/src/BatchExportExecutable.php
@@ -5,6 +5,7 @@ namespace Drupal\mukurtu_export;
 use ZipArchive;
 use Exception;
 use Drupal\file\Entity\File;
+use Drupal\Core\File\FileSystemInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 
 class BatchExportExecutable implements MukurtuExportExecutableInterface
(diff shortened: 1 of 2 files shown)


christa-kavanagh-wsu (approved):
This works for me!

PR #1840, opened 2026-07-17
Make the UID 1 super admin warning dismissible for the current session

### LINKED ISSUE #917: Logged in as super admin warning discussion
Logging this warning as a discussion point: it may be irrelevant once the user roles are implemented.

Image

michael-wynne-wsu:
Updated the admin warning message and added a per-session dismiss "X'.

Screenshot 2026-07-17 at 3 07 17 PM

## Summary
- Fixes #917 — UID 1 previously saw the "you are logged in as the super admin" error on every single page load with no way to dismiss it.
- Adds a small `×` dismiss control to the message. Dismissing it sets a per-session flag (via the request session, not `PrivateTempStore`/State API), so it goes away for the rest of the current login session and reappears the next time UID 1 logs in.
- New CSRF-protected route + controller (`SuperAdminWarningController`) restricted to UID 1 handles the dismissal and redirects back to the page the user was on (via Drupal's built-in `destination` query param handling).

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — no config/schema changes, only code, routing, and CSS.)
- [x] I have run an accessibility check, and resolved any issues. (One moderate, non-blocking note: the dismiss link sits inside Drupal core's shared `role="alert"` message wrapper, which ARIA guidance says shouldn't hold interactive controls. Accepted as a known trade-off rather than overriding the shared `status-messages.html.twig` template used for every error site-wide — happy to revisit if reviewers feel differently.)
- [x] I have recompiled SCSS if required. (Not applicable — plain `.css` file added under `mukurtu_core/css/`, matching the module's existing non-SCSS stylesheets, not the theme's SCSS pipeline.)
- [ ] I have updated or created CI/CD tests, if required. (No automated test added — there's no existing test coverage for `mukurtu_core_page_attachments()` or the comparable `mukurtu_setup` dismiss-controller pattern to extend, and this change was manually verified end-to-end instead. Flagging for reviewer input on whether a kernel test is expected here.)
- [x] I have run the pr-expert-review skill and resolved all relevant issues. (Found and fixed one duplication issue: the session-storage key was hardcoded as a literal string in the `.module` file instead of reusing the controller's `SESSION_KEY` constant.)

## Test plan
Verified end-to-end in a fresh DDEV environment (site with content, logged in as UID 1):
- [x] Message renders with the `×` control, correct `aria-label` ("Dismiss for this session"), correct CSRF token + destination link
- [x] Clicking `×` redirects back to the same page and the message disappears
- [x] Stays dismissed across page loads/navigation in the same session
- [x] Reappears in a genuinely new login session (confirms per-session, not permanent)
- [x] A non-UID-1 user gets a 403 hitting the dismiss route directly
- [x] Existing suppression logic unchanged (empty site, ajax/modal/dialog requests)
- [x] Keyboard-only: tab to the `×` control shows a visible focus ring; no console errors
- [x] No new PHP warnings/errors in watchdog logs


joelle-birano-wsu (approved):
Works for me!

christa-kavanagh-wsu (approved):
Works for me!

nick-deer (approved):

PR #1845, opened 2026-07-17
Migrate phpunit.xml to the modern PHPUnit schema

michael-wynne-wsu:
One of a handful issues trying to clean up all the noise in our test suites.

## Summary
- \`phpunit.xml\` validated against a deprecated PHPUnit schema (1 of the 97 runner deprecations flagged in CI for #1839): \`<filter>/<whitelist>\`, \`printerClass\`, and \`<listeners>\` are all removed in the modern schema
- \`\Drupal\Tests\Listeners\DrupalListener\` and \`HtmlOutputPrinter\` no longer exist anywhere in this Drupal core version, so removing them isn't a functional regression — they were already dead references
- Ran \`vendor/bin/phpunit --migrate-configuration\` against a real DDEV site (not hand-edited blind) to get a verified-correct schema conversion (\`<filter>/<whitelist>\` → \`<source>/<include>\`)
- Restored the browser-output-on-failure capability those old classes used to provide, via the modern \`<extensions>\` mechanism (\`HtmlOutputLogger\`), matching Drupal core's own current \`core/phpunit.xml.dist\` exactly
- Added the \`xmlns:xsi\`/\`xsi:noNamespaceSchemaLocation\` schema attributes

## Test plan
- [x] Ran the \`mukurtu_collection\` kernel test group (18 tests, 929 assertions) against the migrated config in a real DDEV environment — no failures
- [ ] Confirm the "Mukurtu v4 - PHP 8.4" CI job passes with this config and the schema deprecation warning is gone


nick-deer (approved):
@michael-wynne-wsu I'm having trouble testing this on my end. On my local, I run PHPUnit tests through the `phpunit.ddev.xml` file instead of `phpunit.xml`. These two are identical, so I tried to copy the contents of `phpunit.xml` into my local `phpunit.ddev.xml` and run the tests as usual, but I ran into errors. I don't want to hold up this PR though, so I'm inclined to approve and just keep tinkering with my local independently.

PR #1847, opened 2026-07-18
Add Klaro for GDPR/cookie consent management

### LINKED ISSUE #1052: GDPR  and cookies compliance
Javier recommended using https://www.drupal.org/project/klaro

It will require some work with the optional consent pop up.

michael-wynne-wsu:
## Summary

Closes #1052. Adds the [Klaro](https://www.drupal.org/project/klaro) module (v3.1) for GDPR/cookie consent management, enabled by default alongside the existing `google_tag` and `consent_popup` modules.

- In theory this has included "sensible" defaults, but I have not vetted the settings yet. Any thoughts on what we should look at are appreciated.
- I think the only UI thing to confirm is that it doesn't collide with the consent popup (which it doesn't in my testing).

Screenshot 2026-07-17 at 6 58 50 PM

Screenshot 2026-07-17 at 7 01 34 PM

- Enables `drupal/klaro` (^3.1) and grants its `use klaro` permission to Anonymous/Authenticated — without this permission the module installs but silently shows no consent UI to any visitor, so it's required for the feature to actually do anything.
- Overrides `klaro.settings` to turn on Klaro's built-in "reopen consent manager" toggle button so visitors can revisit their choices later (no custom footer/JS code needed — Klaro ships this itself).
- Fixes a real z-index collision: `consent_popup`'s full-viewport protocol-consent backdrop and Klaro's banner/modal both shipped at `z-index: 1000`. Raises the protocol gate above Klaro's range so it always takes precedence when a site enables both, and hides Klaro's UI while the gate is open so it isn't left invisible-but-still-focusable behind the backdrop (caught by the accessibility check, see below).
- Adds `mukurtu_core_update_40076` to install the module and apply the same settings/permission changes on existing sites. Verified by rolling back a test site's schema version and re-running `drush updatedb`.
- Documents in the README how to activate Klaro's bundled Google Consent Mode v2 services (`gtm_consent_mode`, `ga_consent_mode`, `google_ads_consent_mode`) for sites that configure a real Google Tag Manager container — this is Klaro's own built-in integration, no custom wiring required.

## Test plan

- [x] Fresh `drush site-install` in a scratch DDEV project confirms Klaro installs cleanly and is enabled by default.
- [x] Confirmed the consent UI only attaches once the `use klaro` permission is granted and at least one Klaro service is enabled (by design) — verified via `drupalSettings.klaro` and the attached library appearing in page output.
- [x] Simulated an existing-site upgrade (rolled back `mukurtu_core`'s schema version, uninstalled klaro) and re-ran `drush updatedb`; confirmed the module, settings, and both role permissions all end up correctly applied.
- [x] Confirmed the compiled z-index/accessibility CSS rules are present and correctly ordered in the site's aggregated CSS output.

## Pre-merge checklist:


joelle-birano-wsu:
This seems to be working well. Does Klaro support Soundcloud and external embeds? It makes sense to include that if possible since we have Youtube and Vimeo there.

Also noticed that when Youtube and Vimeo are blocked, the warning messages look like this, which is maybe not a big deal once a video is allowed, but a little messy if not. I'm guessing we don't have much control over styling?

in browse:
Screenshot 2026-07-20 162718

in the full content item:
Screenshot 2026-07-20 163026

christa-kavanagh-wsu (commented):
I noticed the same styling things as Joelle.
I agree about the SoundCloud assets, but I would also add external embeds in there as needing the consent management.

michael-wynne-wsu:
Okay styling cleaned up quite a bit, and a generic "other services" added for our external embeds.

Screenshot 2026-07-23 at 10 01 40 AM

Screenshot 2026-07-23 at 10 01 46 AM

nick-deer (approved):

Issue #1852, opened 2026-07-19
Legacy TK Labels: decommissioning workflow for legacy projects post-migration


michael-wynne-wsu:
Follow-up to #206.

## Problem

Once a v3-migrated site has fully transitioned its content off the synthetic "legacy" Local Contexts projects (`default_tk`, `sitewide_tk`, `comm_{nid}_tk`) and onto real Local Contexts Hub projects, there's currently no defined process for removing the legacy projects/labels. #206 explicitly raised "how do we decommission legacy projects after update?" as an open question.

## Scope

- Depends on #1851 (batch reassignment tooling) — decommissioning only makes sense once there's a real path to get content off legacy projects.
- Define what "safe to decommission" means: presumably a legacy project with zero remaining content references.
- `LocalContextsProject::inUse()` and `LocalContextsSupportedProjectManager::removeProject()` (`modules/mukurtu_local_contexts/src/`) already implement the underlying "is this project referenced, and if not, delete it" mechanics used by the existing per-project admin delete action — decommissioning legacy projects specifically may just need surfacing/guiding that existing capability, rather than new deletion logic.
- Consider whether admins need a dedicated view to see, per legacy project, how many/which items still reference it, to know when it's actually safe to decommission (vs. only finding out via the generic "cannot be removed because it is in use" error today).
- Consider whether decommissioning should be a manual admin action only, or could optionally be an automated cleanup once reassignment tooling confirms zero remaining references.

## Related context

- Phase 1 of #206 (enforcing that legacy labels can't be newly selected, only kept on content that already has them) is being addressed separately on branch `206-legacy-tk-label-readonly`.
- #1851 is the prerequisite batch reassignment tooling issue.

michael-wynne-wsu:
Implemented in #1855, on branch `1852-legacy-project-decommission` (based on `1851-legacy-project-remap` / #1854, since this depends on the permission and query methods added there - will need a rebase once #1853/#1854 merge).

With this, all three parts of #206's original decommissioning question are addressed: #1853 (read-only enforcement), #1854 (batch reassignment), #1855 (decommissioning).

michael-wynne-wsu:
[...]

Issue #1851, opened 2026-07-19
Legacy TK Labels: batch reassignment tooling (legacy project → real Hub project)


michael-wynne-wsu:
Follow-up to #206.

## Problem

v3-migrated sites end up with synthetic "legacy" Local Contexts projects (`default_tk`, `sitewide_tk`, `comm_{nid}_tk`) referenced by existing content. The expected user path is to set up real Local Contexts Hub projects and move their content over to them, but today there's no tooling to do that at scale.

From the original discussion on #206, two possible workflows were identified, neither assessed for feasibility yet:

1. Add real Hub projects, then manually (in-site or via a batch UI) add the new project/label info to each item, remove the old legacy info, and eventually delete the legacy projects. Tedious but allows per-item review.
2. Build a system where an admin maps a legacy project to its replacement real Hub project (and legacy labels to their real-project equivalents), then automates reassigning all affected content's `field_local_contexts_projects` / `field_local_contexts_labels_and_notices` values from the legacy IDs to the real ones.

If site's real Hub project structure ends up very different from their v3 label setup, a simple 1:1 automated mapping may not be practical, and a review-oriented workflow may be more appropriate.

## Scope

- Design (and get sign-off on) which workflow(s) above to support before building anything.
- If a mapping/batch approach is chosen: an admin UI to define the legacy → real project/label mapping, plus a batch operation (leveraging Drupal's batch API, given this could touch a large number of nodes) to rewrite affected field values.
- Needs to work for both the sitewide (`default_tk`/`sitewide_tk`) and per-community (`comm_{nid}_tk`) legacy project scopes.

## Related context

- Existing admin UI (`ManageSupportedProjectsBase`/`ManageSupportedProjectsSite`/`ManageSupportedProjectsGroup` in `modules/mukurtu_local_contexts/src/Form/`) only supports "Add/Sync" and "Delete" of supported projects today — no merge/reassign capability exists.
- `LocalContextsProject::inUse()` (`modules/mukurtu_local_contexts/src/LocalContextsProject.php`) already knows how to detect whether a project is referenced by content, which will likely be useful here.
- Phase 1 of #206 (enforcing that legacy labels can't be newly selected, only kept on content that already has them) is being addressed separately on branch `206-legacy-tk-label-readonly`.

michael-wynne-wsu:
[...]

Issue #102, opened 2023-09-28
TK Label "Legacy" mode


steve-taylor-wsu:
Note for the team: This is old info, mostly book-keeping for me, this doesn't require discussion (unless you spot something abnormal).

With the new direction taken with the Local Contexts Hub integration, we will need to design a new technical approach to supporting "legacy" TK labels, as the previous implementation approach will no longer suffice because it existed prior to the new understanding of how the Hub works.

## Background
In v3, TK labels are customized within Mukurtu, at either the site or the community level. In v4, we want labels to come from the LC hub. We need a way to migrate existing v3 labels into v4. The functionality we decided on was to:
* Import existing labels as "legacy" labels.
* Allow these legacy labels to be used/referenced, giving users time to get going on the LC hub.
* Disallow users from creating "new" legacy labels or changing existing legacy labels, to incentivize using the Hub.

## Approach
I *think* we can add some sort of legacy flag to the existing DB cache tables for the fetched hub labels and use the existing `field_local_contexts_labels_and_notices` field to pick individual legacy labels for content. It might be worth evaluating between doing that and creating bespoke copies of those tables specifically for legacy labels. We'll also need to plan out the site vs community customization levels.

PR #1853, opened 2026-07-19
Enforce read-only legacy TK Labels (#206 phase 1)

### LINKED ISSUE #206: Legacy TK Labels migration
Is there a system for migrated (from v3) legacy labels to be converted into v4 Hub-compliant labels?

michael-wynne-wsu:
Part 1 of remaining LC Label work.

- No change to the migration process. Labels still show as before.
- However now the "legacy" project is only accessible to content that already references it.
- I've tested this with one real world migration that has site-wide labels. I do not have an accessible v3 site with per-community labels to test yet.

Screenshot 2026-07-19 at 12 31 11 PM

Screenshot 2026-07-19 at 12 36 22 PM

Screenshot 2026-07-19 at 12 30 36 PM

Follow-up work (deliberately out of scope here — tracked separately since they're larger and differently shaped): #1851 (batch reassignment tooling from legacy to real Hub projects) and #1852 (decommissioning legacy projects), both linked from #206.

## Summary

Closes the gap between #206 and the original design intent recorded in #102: v3-migrated "legacy" Local Contexts TK Labels projects (`default_tk`, `sitewide_tk`, `comm_{nid}_tk`) were fully selectable for new content assignments, when the intent was for them to remain usable on content that already references them, but not offered as a choice for new assignments (to incentivize moving to the real Local Contexts Hub).

- Adds `LocalContextsSupportedProjectManager::isLegacyProjectId()` as the single canonical way to classify a legacy project ID, replacing an existing incomplete `$exclude_legacy` check that only covered `default_tk`/`sitewide_tk` and missed the `comm_{nid}_tk` pattern.
- `LocalContextsProjectItem`/`LocalContextsLabelItem` now exclude legacy projects/labels from `getSettableValues()`/`getSettableOptions()` for *new* selections, while preserving any legacy value already stored on the entity being edited. This preservation is required, not just nice-to-have: Drupal's `AllowedValues` constraint validator calls these same methods against the entity being saved, so naively excluding legacy IDs would fail validation (and block saving) for any content that already has a legacy reference.
- Parity fix: the community/protocol Local Contexts project directory pages now exclude legacy projects like the sitewide directory page already did.
- Added missing `field.storage_settings.*` config schema entries for the two field types (pre-existing gap — nothing previously exercised these field types in a kernel test context, so it went unnoticed).
- New kernel test coverage (`modules/mukurtu_local_contexts/tests/`) — this module previously had zero automated tests.

## Test plan

- [x] 7 new kernel tests, including a regression test proving that resaving an entity with an existing legacy label/project reference (without touching that field) does not drop the reference, and produces no `AllowedValues` violations.
- [x] Full existing kernel test suite (`mukurtu_protocol`, `mukurtu_collection`, `mukurtu_import`, `mukurtu_export`, `mukurtu_drafts`, `original_date`) re-run: 142/142 passing, no regressions.
- [x] Verified against a real v3→v4 migration in a live environment: legacy labels/projects migrated as expected, an existing node with a legacy label survived a resave with zero violations as a real user, and a brand-new node correctly has legacy labels excluded from its selectable options.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — no schema/config-install changes, only `config/schema` which needs no import step.)
- [x] I have run an accessibility check, and resolved any issues. (No UI/markup touched — confirmed out of scope for this diff.)
- [ ] I have verified that all expected checks pass. (Pending CI run on this PR — verified locally via DDEV.)


michael-wynne-wsu:
No user review is needed, I think just a code review from @nick-deer is necessary.

michael-wynne-wsu:
Test coverage caveat: the real-world migration verification described above covered sitewide legacy labels (`sitewide_tk`/`default_tk`) only. I don't yet have access to a v3 site with per-community customized TK labels (`comm_{nid}_tk`), so that path is currently verified only by the synthetic kernel tests (`testGetSiteSupportedProjectsExcludeLegacyCoversAllPatterns`, `testIsLegacyProjectId`), not against a real migration. Worth testing against a real per-community dataset before/after merge if one becomes available.

nick-deer (approved):

PR #1854, opened 2026-07-20
Add legacy Local Contexts project remap tool (#1851)

### LINKED ISSUE #1851: Legacy TK Labels: batch reassignment tooling (legacy project → real Hub project)

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1851

- Again, only able to test with a set of site-wide labels at this time, but it seems good to me.
- admin and Mukurtu Manager have access to this.

Screenshot 2026-07-20 at 9 46 12 AM

Screenshot 2026-07-20 at 9 46 46 AM

Screenshot 2026-07-20 at 9 46 58 AM

## Summary

Implements #1851 (a follow-up to #206): admins currently have no way to move content off legacy (v3-migrated) Local Contexts projects (`default_tk`, `sitewide_tk`, `comm_{nid}_tk`) onto real Local Contexts Hub projects. This adds that tool.

**Note: base branch is `206-legacy-tk-label-readonly` (#1853), not `main`** - this work directly depends on `isLegacyProjectId()` and the test harness added there, which hasn't merged yet. Rebase onto `main` once #1853 lands.

- 3-step admin wizard (`RemapLegacyProjectBase`/`Site`/`Group`, following this module's existing site/community/protocol route pattern): select a legacy project and a target real Hub project (target can be from any scope, per discussion) → optionally map individual legacy labels/notices to their real equivalents → **mandatory preview** showing exactly which content will change (and a deduplicated node count, since a node can be referenced via both the whole-project field and a label at once) before anything runs.
- New `LegacyProjectRemapPreviewBuilder` computes that preview; new `LegacyProjectRemapBatch` (standard `BatchBuilder` pattern, modeled on `mukurtu_migrate`'s existing batch) does the actual rewrite, chunked 25 nodes/pass, with a new revision + log message per changed node, and a results report naming exactly which labels/notices are still unmapped for a follow-up pass. Mapping is intentionally partial/incremental-friendly - the tool can be run again as more labels get mapped over time.
- New dedicated permission `administer local contexts legacy projects` gates all three routes, rather than reusing the broad `administer site configuration` this module uses elsewhere, since this does bulk, hard-to-reverse rewrites of culturally sensitive TK Label references.
- **Bugfix bundled in**: `LocalContextsProject::inUse()` had `if (labels) {} else if (notices) {}`, so notices were never checked when a project had any labels (the common case) - this could make the existing delete-project safety check wrongly report a project as unused. Fixed by extracting `getReferencingNodeIds()`, which checks both independently, and having `inUse()` call it.

## Test plan

- [x] 16 new kernel tests: the `inUse()`/notices regression, straight/mapped/unmapped label+notice rewrites, mixed mapped+unmapped values on the same node, multi-pass batch chunking, incremental/partial remap across multiple runs, graceful handling of a missing node ID, and preview dedup.
- [x] All 7 kernel tests from #1853 and all 142 pre-existing kernel tests re-run: no regressions.
- [x] Full manual walkthrough against a real v3-migrated dataset in a live DDEV environment (no browser tool available, so driven via raw HTTP with a session cookie): seeded a real target project, drove all 3 wizard steps, confirmed the preview matched real per-label node counts (29/113/778 across the site's 3 real legacy labels), ran the actual batch (778 real nodes rewritten correctly in one request), verified the rewritten field value format and new revision/log message, then **restored a pre-batch DB snapshot** afterward so the dataset wasn't left permanently altered for further testing.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required - no `config/install` changes; routing/links/services/permissions YAML files aren't synced config.)
- [x] I have run an accessibility check, and resolved any issues. (One fix applied: heading level `h3` → `h2` on the wizard's step 2/3 headings to avoid skipping `h1`→`h2`→`h3` in the document outline. Everything else - radio fieldsets, per-row invisible `#title` on the label-mapping selects, non-color-only status text, full-page-reload steps instead of AJAX - already met WCAG 2.2 AA.)
- [x] I have updated from `main` and resolved any merge conflicts. (Base branch `206-legacy-tk-label-readonly` is itself up to date with `main`.)
- [ ] I have verified that all expected checks pass. (Pending CI run on this PR - verified locally via DDEV.)


nick-deer (approved):
Code and tests look good and the update hook ran successfully, but I did get an error complaining of the charts, charts_chartjs, and visitors modules not being installed. I'm assuming this will sort itself out on merge, but wanted to point it out just in case.

michael-wynne-wsu:
I think that should shake out, agreed. I'll keep an eye out.

PR #1855, opened 2026-07-20
Add legacy Local Contexts project decommissioning workflow (#1852)

### LINKED ISSUE #1852: Legacy TK Labels: decommissioning workflow for legacy projects post-migration

michael-wynne-wsu:
Adds a "decommission" action for legacy projects.
And adds the content count for all projects (not shown).

Screenshot 2026-07-20 at 12 09 52 PM

## Summary

Implements #1852 (a follow-up to #206, dependent on #1851): once a legacy project (`default_tk`, `sitewide_tk`, `comm_{nid}_tk`) has zero remaining content references, admins had no way to know that with confidence and no dedicated way to remove it.

**Note: base branch is `1851-legacy-project-remap` (#1854), not `main`** - this depends on `isLegacyProjectId()`, `getReferencingNodeIds()`, and the dedicated permission added there, none of which have merged yet. Rebase onto `main` once #1853/#1854 land.

Investigation found the underlying deletion mechanics already worked correctly for legacy projects (`ManageSupportedProjectsBase::submitDelete()` already checked `inUse()`, and `LocalContextsSupportedProjectManager::removeSiteProject()`/`removeGroupProject()`/`removeProject()` already handled full cleanup). The actual gap was pure visibility - the Manage Projects page showed only `Active`/`Not added`/`Not available`, with zero indication of in-use status until a generic Delete attempt failed with an error.

- **Manage Projects page**: legacy project rows (which fell into the "Not available"/deleted-from-hub bucket, misleadingly, since they were never really on the Hub) now show `Active — no remaining references` or `Active — referenced by N node(s)`.
- **New "Decommission" bulk action**, visible only to users with the dedicated permission from #1851 (not the broader `administer site configuration` that gates the rest of this page). Validates each selection is both a legacy project and has zero remaining references; rejects anything else with a clear per-item error.
- **New confirmation page** (`DecommissionLegacyProjectsConfirmForm`, a standard `ConfirmFormBase`): rather than deleting immediately, the validated selection is handed off via private tempstore to a confirm step that lists exactly what will be permanently removed (project title + label/notice counts), re-validates immediately before deleting (in case something started referencing the project in the window between selection and confirmation), and reuses the existing removal methods unchanged.
- **Manual only** - no automated/cron decommissioning, per discussion.

## Test plan

- [x] 7 new kernel tests: permission enforcement, valid-project acceptance, rejecting a non-legacy project, rejecting a still-in-use legacy project, the confirm form's `access()` logic (no permission / no pending request / valid), successful decommission (verifies all DB rows removed), and the re-validation-blocks-if-now-in-use safety net.
- [x] All 30 `mukurtu_local_contexts` kernel tests and all 142 pre-existing kernel tests re-run: no regressions.
- [x] Full manual walkthrough against real migrated data in a live DDEV environment (DB snapshot taken beforehand, restored afterward): used the #1851 remap tool to bring the real `sitewide_tk` project (920 real references across 3 labels) down to zero, confirmed the Status column showed the correct text for both a still-in-use project (`default_tk`) and the now-zero-reference one, submitted a **mixed** selection to Decommission and confirmed `default_tk` was correctly rejected while only `sitewide_tk` proceeded to the confirm page, confirmed the decommission, and verified via direct DB queries that `sitewide_tk`'s rows were fully removed while `default_tk` was untouched.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required - no `config/install` changes, only a new route and form logic.)
- [x] I have run an accessibility check, and resolved any issues. (One real issue found and fixed: three strings used a literal "(s)" pluralization that doesn't actually pluralize and reads awkwardly to screen readers - converted to `formatPlural()`, consistent with existing usage elsewhere in the same file.)
- [x] I have updated from `main` and resolved any merge conflicts. (Base branch is up to date with its own base.)
- [ ] I have verified that all expected checks pass. (Pending CI run on this PR - verified locally via DDEV.)
- [x] I have run the pr-expert-review skill and resolved all relevant issues. (Found and fixed one real edge case: if a community/protocol is deleted in the window between selecting its legacy project for decommission and confirming, the code now skips it safely instead of silently misapplying the removal to site scope and reporting a false success.)


nick-deer (approved):
Code reviewer checklist:
- [ ] Update hook(s) ran w/out errors (N/A)
- [ ] Update hook(s) changes verified in live site (N/A)
- [x] Clean rebuild
- [x] Manual confirmation of decommission workflow using a v3 -> v4 migrated site with legacy project

PR #1859, opened 2026-07-20
Add per-item review and removal tool for unmappable legacy labels/notices

### LINKED ISSUE #1856: Legacy TK Labels: per-item review workflow for labels with no real Hub equivalent
Follow-up to #206, #1851, #1852.

## Problem

The legacy project remap tool (#1851) lets an admin map each legacy label/notice to a real Hub equivalent, or leave it unmapped ("skip") for a later pass. But if a legacy label genuinely has no appropriate real-Hub equivalent, "skip" is permanent by default - the content keeps referencing the legacy label forever, which means `LocalContextsProject::inUse()` never returns false for that project, and it can never be decommissioned (#1852) as long as even one such reference remains.

There's currently no way to deliberately resolve this: no path to say "this legacy label has no equivalent, remove it from content" as an intentional action distinct from "I haven't gotten to mapping this yet."

## Why this needs its own design (not just a third wizard option)

A naive fix would be a third option in the remap wizard's mapping step ("Remove from content" alongside "map to X" and "skip"), applied as a blanket bulk operation across every node referencing that label - the same shape as the existing remap batch.

That's not appropriate here. TK Labels signal cultural protocol/access restrictions, and removing one from content is a more consequential, harder-to-justify-in-bulk action than remapping it to an equivalent (which preserves the intent, just pointed at new data). A blanket "strip this label from all N nodes that have it" risks silently dropping a meaningful cultural signal from content an admin never actually looked at.

Confirmed direction: this needs a **per-content-item review/selection step**, not a bulk action - closer to the manual "workflow (a)" described in the original #206 discussion (per-item review) than the bulk "workflow (b)" the remap tool already implements. Concretely, something like: for a given still-unmapped legacy label, show the admin the specific list of nodes that reference it, and let them choose (per item, or via an explicit multi-select) which ones should have the label removed - not an all-or-nothing bulk toggle.

## Scope (not yet designed in detail)

- Where this lives: likely a new step/tool reachable from the legacy project remap wizard or the Manage Projects page, once a label has been sitting unmapped for a project the admin is trying to decommission.
- What "remove" means precisely: strip that one label/notice value from the entity's `field_local_contexts_labels_and_notices` field (existing precedent: `LegacyProjectRemapBatch` in `modules/mukurtu_local_contexts/src/Batch/` already knows how to parse/rewrite this field - removal is simpler than remapping, just needs to drop the value instead of replacing it).
- Whether removal should require a per-item confirmation, a review queue, or something else - needs its own design pass rather than assuming the remap wizard's UI patterns transfer directly.
- Whether there's any value in flagging/logging which nodes had a label removed this way, for later audit.

## Related context

- #1851 (remap tool) and #1852 (decommission workflow) are both already implemented; this issue is specifically about the residual "unmappable label" gap discovered while building #1852.
- `LocalContextsProject::getReferencingNodeIds()` (`modules/mukurtu_local_contexts/src/LocalContextsProject.php`, added in #1851) already returns the exact per-label list of referencing node IDs needed to build a per-item review UI - no new query logic should be needed to enumerate candidates.

michael-wynne-wsu:
Closes #1856.

## Summary

This is the last piece of the legacy TK Labels stack (#206): the remap tool (#1851, PR #1854) lets an admin map each legacy label/notice to a real Local Contexts Hub equivalent, or "skip" it when there's no real match. A skipped label stays on content forever, and since a legacy project can't be decommissioned (#1852, PR #1855) while it still has any remaining references, one permanently-skipped label blocks decommissioning of the whole project indefinitely.

This PR adds a deliberate, admin-driven way to remove a specific unmapped legacy label or notice from content:

- **Breakdown page** (`/admin/local-contexts/legacy-projects/{project_id}/breakdown`) — for a legacy project, lists every label/notice that still has referencing content, with a reference count and a link into the review flow. Linked from the "Active — referenced by N nodes" status text on the Manage Projects page (which now links for legacy projects specifically).
- **Review page** — a per-content-item selection table (unchecked by default) of exactly the nodes referencing the chosen label/notice. This is deliberately **not** a bulk "remove everywhere" action — TK Labels signal cultural protocol restrictions, and silently stripping one from unreviewed content would be a real harm. An admin must explicitly opt each item in.
- **Confirm page** — shows the selected items, re-validates against the current state (in case something else resolved a reference in the interim), and dispatches removal via batch.
- **Batch removal** — removes only the one targeted compound value (`"{project}:{id}:{type}"`) from each selected node's `field_local_contexts_labels_and_notices`, leaving every other label, notice, and project reference on that node untouched. Creates a new revision with a log message.

Whole-project references are out of scope here — they're always resolved automatically whenever the project itself is remapped, since the remap wizard has no "skip" option for the whole-project field.

Two accessibility fixes surfaced during review are also included: a visually-hidden "Actions" header on the breakdown table, and a `#title` on the confirm form's item list (the latter also backported to the sibling `DecommissionLegacyProjectsConfirmForm` from PR #1855 for consistency).

## Test plan

- 3 new kernel test classes (`LegacyLabelRemovalBatchTest`, `LegacyLabelRemovalReviewFormTest`, `LegacyLabelRemovalConfirmFormTest`) covering access checks, the "only selected items change" guarantee, multi-pass batch chunking, re-validation against interim changes, and label vs. notice removal.
- Full existing `mukurtu_local_contexts` kernel suite re-run, no regressions.
- Manual walkthrough against real migrated data (DB snapshot before/after): breakdown → review (confirmed unchecked-by-default, selected a subset) → confirm → verified only the selected nodes had the label removed, all other labels/notices/projects on the same nodes untouched, new revisions created with the expected log message.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — no `config/install` changes, only routing and PHP.)
- [x] I have recompiled SCSS if required. (Not required — no SCSS/CSS changes.)
- [x] I have updated from `main` and resolved any merge conflicts. (This branch targets `1852-legacy-project-decommission`, the last unmerged PR in the #206 stack, which is itself up to date with `main`.)


nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

PR #1863, opened 2026-07-21
Fix protocol sync toggle defaults and placement (#1823)

### LINKED ISSUE #1823: Protocol Sync Updates
- Protocol sync: deselect if no protocols are selected and on protocol unaware content.
- Sync toggle placement 
   - In the modal: - leave as is
   - In the media asset library - move synced protocol toggle underneath sharing setting for single and bulk uploads
   - Bulk images - move synced protocol above identifier field

michael-wynne-wsu:
## Summary

Follow-up adjustments to the media-to-content protocol sync toggle (`mukurtu_protocol_sync`), addressing #1823:

- **Default state**: the sync checkbox now defaults to **unchecked** whenever no protocol data can be inferred, instead of always defaulting to checked. This covers both cases previously described in #1823:
  - Non-protocol-aware content types (landing page, basic page, article) — media created from these pages never has protocols to infer from.
  - Protocol-aware content types where no protocol has been selected yet before opening the media dialog.
- **Bulk upload placement**: investigation showed the bulk image upload form wasn't uniquely broken — no bulk upload form (image, audio, video, document, remote_video, soundcloud) was ever wired into the protocol-sync placement system in the first place. All of them are now wired in.
- **Toggle position**: for contexts where protocols are never pre-populated (standalone single media add/edit forms, and both bulk upload forms), the sync toggle now renders *after* protocol selection instead of before it, matching the natural workflow of picking protocols first and then deciding whether to keep them synced. The in-page media-library modal (opened from within a content page, where protocols are typically already inherited) is unchanged — toggle stays at the top.

## Changes

- `modules/mukurtu_protocol_sync/mukurtu_protocol_sync.module`
  - `_mukurtu_protocol_sync_default_sync_checkbox()` now defaults to unchecked (`0`) instead of checked (`1`).
  - Added `hook_form_mukurtu_media_bulk_upload_form_alter()` and `hook_form_mukurtu_media_bulk_url_create_form_alter()`, sharing a small helper that marks each uploaded item's sync checkbox for JS placement, mirroring the existing pattern used for the media-library modal.
  - Marked the standalone media form's sync checkbox with a "position after" attribute.
- `modules/mukurtu_protocol_sync/js/protocol-sync.js`
  - The existing DOM-relocation behavior now branches on a `data-mukurtu-protocol-sync-position="after"` attribute to append the toggle after the protocols fieldset's contents, instead of always inserting it at the top.

No field storage, config, or entity display changes are involved — no update hook required.

## Test plan

- [x] Verified PHP and JS syntax (`php -l`, `node --check`)
- [ ] Manual verification on Tugboat preview (author will test: non-aware content types default unchecked, protocol-aware types with no selection default unchecked, existing in-page modal behavior unchanged/regression-free, bulk image/video/document/etc. uploads show the toggle merged below protocols, standalone single media add form shows the toggle merged below protocols, multi-item bulk pages scope correctly per item)

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — no field storage/config changes.)
- [x] I have run an accessibility check, and resolved any issues. (No regressions found — DOM/tab order and existing aria-live toggle behavior unaffected by the reposition.)
- [x] I have recompiled SCSS if required. (Not applicable — no SCSS changes.)
- [ ] I have updated or created CI/CD tests, if required. (No existing test coverage for this module; verification is manual per the project's current testing approach for this feature.)
- [x] I have verified that all expected checks pass. (Pending CI run and author's Tugboat verification.)


michael-wynne-wsu:
Pushed a follow-up fix based on Tugboat testing feedback:

- **Root cause found for the "sync active by default on non-aware content" bug**: Basic Page's media field is machine-named `field_page_media_assets`, which wasn't in `mukurtu_protocol_sync_form_node_form_alter()`'s hardcoded field-name list. That meant its "Add media" button never stashed/cleared the per-user TempStore entry used to detect protocol selection, so a *previous* node's protocol selection could leak through as the sync default on an unrelated node. Replaced the hardcoded list with dynamic detection of any field with a media-library open button, so this can't silently recur for other content types with differently-named media fields.
- **Removed the guard that blocked opening "Add media" entirely** when a protocol-aware node had no protocols selected yet. Syncing is optional — the modal should always open regardless of protocol selection; the toggle already correctly defaults to unchecked when nothing can be inferred.

Also worth double-checking: your Tugboat screenshot showed a "Relaunch to update" button in the top right, which may mean the environment was serving a build from before the first fix landed — worth a manual relaunch before retesting either way.

christa-kavanagh-wsu (commented):
- The Add media views from the media library are working as expected.
- The Add media modal views from content still have the Sync Protocol option before any protocols are selected in either the content or the media asset, and the sync protocol option also appears in the modal on Basic Pages.

Image

michael-wynne-wsu:
> The Add media modal views from content still have the Sync Protocol option before any protocols are selected in either the content or the media asset, and the sync protocol option also appears in the modal on Basic Pages.

Yeah I want to keep it consistent across the modal, it's way too messy to move it around based on what is on the page before opening the modal. What should be happening is that if there is no way to select a protocol (eg: basic page), or a protocol is not selected (eg: just not choosing one on a DH item), is when the modal opens and media is added, the sync toggle is not turned on by default.

christa-kavanagh-wsu (approved):
Oh, great! That all appears to be working as expected.

joelle-birano-wsu (approved):

nick-deer (approved):

PR #1864, opened 2026-07-21
Fix marker clustering across Views grouping values

### LINKED ISSUE #1813: Map marker clustering not visibly grouping nearby points
## Summary

While reviewing map mode changes for #1624, noticed that nearby points on the Leaflet maps (`/browse`, `/digital-heritage`, `/dictionary`, homepage map block) don't appear to cluster together as expected — points near each other show as separate markers rather than a clustered count bubble.

Not clear yet whether this is a pre-existing regression or has always behaved this way; flagging for follow-up investigation since it wasn't caused by the #1624 work (see below).

## What's been ruled out

Verified during #1624 review that none of the following affect this:
- `leaflet_markercluster` config (`control: true`, spiderfy/coverage options) is present and unchanged in both the source config files and the live active config for all four map views (`mukurtu_browse`, `mukurtu_digital_heritage_browse`, `mukurtu_dictionary`, `mukurtu_browse_by_map`).
- No JS errors visible in the browser console when the map loads.
- The #1624 changes touched: popup content (added fields), popup link color (CSS), the "Reset View" control, and the row `view_modes` mapping (which view mode renders popup content) — none of which interact with the marker-cluster plugin's own clustering logic.

## Possible leads for investigation

- Each of the four map views has a pre-existing `grouping` feature configured (grouping markers into separate layers/overlays by `field_coverage`). If `leaflet_markercluster` only clusters *within* each group rather than across all groups, points with different `field_coverage` values would never visually cluster even if geographically close. This grouping config pre-dates #1624 and wasn't touched by it — but could explain the symptom if that's how the plugin interaction actually works.
- Worth checking whether this reproduces with points that share the same `field_coverage` grouping value, to isolate whether it's a grouping-vs-clustering interaction issue or something else (e.g. a `drupal/leaflet` version/library issue).

## Steps to reproduce

1. Visit `/digital-heritage` (or `/browse`, `/dictionary`, or the homepage map block).
2. Switch to Map view.
3. Look for multiple markers close together / at the same or nearby coordinates.
4. Expected: a clustered marker showing a count. Observed: individual markers, not clustered.

michael-wynne-wsu:
- Clustering points across multiple content items wasn't working.
- Test by creating several pieces of content with mapped locations and zoom out if needed to see clustering breakdown.

## Summary
- `leaflet_markercluster.drupal.js` gives every Views "grouping" value (e.g. `field_coverage` on the `/browse`, `/digital-heritage`, `/dictionary`, and homepage map views) its own `MarkerClusterGroup` instance. Leaflet only clusters markers within the same instance, so nearby markers in different groups never merged into one bubble.
- Adds a Mukurtu JS override (loaded immediately after `leaflet_markercluster.drupal.js` via `hook_library_info_alter()`) that re-implements `Drupal.Leaflet.prototype.add_features` so all groups share a single cluster group, while keeping the existing per-group layer-toggle control working by moving a group's markers into/out of the shared cluster on `overlayadd`/`overlayremove`.
- No changes to the four Views configs - grouping stays as-is, only the clustering behavior changes.

## Test plan
- [x] Verified in a live DDEV environment via a standalone harness loading the actual shipped JS files against synthetic grouped marker data: nearby markers across two different groups now merge into a single cluster bubble with the correct combined count.
- [x] Toggling a group's checkbox in the map's layer control correctly adds/removes its markers from the shared cluster (count updates accordingly), and re-toggling restores them.
- [x] No JS console errors during load or toggling.
- [ ] Manual re-check on real content on `/browse`, `/digital-heritage`, `/dictionary`, and the homepage map block recommended before merge (this branch was tested against synthetic data since a fresh install has no seeded content).

Fixes #1813

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required - JS-only change, no config/schema changes.)
- [x] I have run an accessibility check, and resolved any issues. (No new markup/ARIA surface; existing layer-toggle checkboxes unaffected.)
- [x] I have recompiled SCSS if required. (Not required - no SCSS changes.)


christa-kavanagh-wsu (approved):
This appears to be working now.

joelle-birano-wsu (approved):

nick-deer (approved):

PR #1866, opened 2026-07-21
Fix Layout Builder 'unsaved changes' warning leaking onto unrelated admin pages

### LINKED ISSUE #1822: Layout Builder "unsaved changes" warning leaks onto unrelated admin pages
## Summary

After interacting with Layout Builder (e.g. opening "Add block" and then canceling out without saving), the "You have unsaved changes." warning does not appear on the layout page itself, but instead shows up later on a completely unrelated admin page (e.g. Media admin content list).

## Steps to reproduce

1. As a non-administrator role (e.g. Mukurtu Manager), go to a basic page's Layout tab (`/node/{id}/layout`).
2. Click "Add block", pick any block type, then cancel/close the dialog without saving.
3. Navigate to a different admin page (e.g. `/admin/content/media`).
4. The "You have unsaved changes." warning appears there instead.

## Root cause

Traced to Drupal core's `PrepareLayout` event subscriber (`web/core/modules/layout_builder/src/EventSubscriber/PrepareLayout.php:72`), which calls `$this->messenger->addWarning('You have unsaved changes.')` whenever the Layout Builder tempstore already has pending changes for a section storage. This fires on every Layout Builder route render, including the off-canvas "choose/add block" AJAX requests.

This is intended (if easily misunderstood) Drupal core behavior — Layout Builder treats simply visiting the layout page as "editing has begun" and stages that in a session-backed tempstore until "Save layout" or "Discard changes" is used. Core even hides this same message while still on the Layout Builder page via CSS, expecting it to surface once you navigate away, as a legitimate reminder of pending edits.

The actual bug: this warning uses Drupal's session-queued messenger (`Drupal::messenger()`), but gin_lb's off-canvas dialogs render messages via a separate Toastify popup system instead of the normal messages block. The warning added during an off-canvas AJAX interaction is never displayed/consumed during that exchange, so it sits queued in the session and surfaces on whatever the next real full-page load happens to be — completely disconnected from the moment the user actually left Layout Builder.

## Discussion needed before fixing

Two different framings, worth deciding between before implementing:

1. **Display-timing fix**: ensure this message is flushed/shown right when the user actually navigates away from Layout Builder (or discarded if nothing was truly changed), rather than leaking onto an arbitrary later page.
2. **Tempstore/UX design question**: decide whether merely opening and canceling the "Add block" dialog should count as "unsaved changes" at all, versus only real edits (e.g. actually adding/configuring/saving a block).

Filed for follow-up rather than fixed as part of PR #1770, to keep that PR scoped to the Layout Builder basic-pages feature.

michael-wynne-wsu:
## Summary

- Fixes #1822: Layout Builder's "You have unsaved changes." warning used to leak onto unrelated admin pages after opening an off-canvas dialog (e.g. "Add block") and closing it without saving.
- Root cause (confirmed via direct testing against a live instance, not just static analysis): core embeds `['#type' => 'status_messages']` with no type filter on every AJAX canvas rebuild (`\Drupal\layout_builder\Element\LayoutBuilder::layout()`), which gin_lb toastifies since it's a genuine `layout_builder.*` route response. Separately, the Layout Builder edit page's own template has no "highlighted" region at all (only "content"), so *any* Messenger-queued message has nowhere to render normally on this page regardless of dialogs or AJAX. Trying to reliably display the warning meant fighting core's own embed mechanism head-on, which produced inconsistent timing, duplication, and inconsistent auto-hide behavior no matter which display approach was tried.
- Since no other content-editing page in Mukurtu shows an "unsaved changes" warning, this PR suppresses it entirely instead of trying to perfect its display: `LbSuppressUnsavedWarningSubscriber` hooks `LayoutBuilderEvents::PREPARE_LAYOUT` at a lower priority than core's own `PrepareLayout::onPrepareLayout()` (the sole source of this warning) and removes just that message immediately after it's queued, before core's rebuild self-embed or anything else can react to it. This keeps #1822 fixed -- there's simply nothing left to leak.
- Separately (unrelated to #1822, found during testing of this fix): "The layout override has been saved." also never appeared after clicking Save, for the same "no region to render into" reason, compounded by gin_lb's `save_behavior: stay` config redirecting back to the same edit page rather than the node's canonical view. `LbPendingMessagesController` + `lb-pending-messages.js` surface that confirmation via a direct Toastify call (persistent, manually dismissed, not gin_lb's auto-hiding native behavior) once per page load.
- No core or contrib (gin_lb) files touched, and no user-facing text added or changed.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required -- new PHP classes + service/route registration only, no config/schema changes.)
- [x] I have recompiled SCSS if required. (Not applicable -- no CSS changes.)
- [x] I have updated or created CI/CD tests, if required. (`LbSuppressUnsavedWarningSubscriberTest`, `LbPendingMessagesControllerTest`.)

## Test plan

- [ ] Open a page's `/node/{id}/layout`, make changes via any off-canvas dialog (add/configure/remove a block or section), then cancel or complete the action -- confirm "You have unsaved changes." never appears anywhere, and never leaks onto a later unrelated page (e.g. `/admin/content/media`).
- [ ] Click Save -- confirm "The layout override has been saved." appears as a dismissible, non-fading notification.
- [ ] Confirm other Messenger warnings/errors unrelated to Layout Builder (e.g. from other modules) are unaffected by the suppression subscriber.


michael-wynne-wsu:

Link:
- https://pr1866-pekxunxq5stolvqjjs3nmcqoaaobijnt.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a5f8bbca81146fbb7c04694

joelle-birano-wsu (commented):
Warning messages are staying on the basic page now, but the warning message appears immediately as soon as you add a block. I think that's a little too early? It would make more sense after the block is created.

Screenshot 2026-07-22 114901

Sometimes the "unsaved" message is layered on top of the "layout override has been saved" message. This happens if I save the page, and then make a change to the page. Both messages will display and they're contradictory, so it's confusing.

Screenshot 2026-07-22 115447

michael-wynne-wsu:
I made those initial changes but I'm finding other weirdness with this one. Surprisingly tricky.

michael-wynne-wsu:
@joelle-birano-wsu it was taking a lot of time and quickly turning into some really messy complex code. So I opted to suppress/remove the "unsaved changes" message. We don't do it anywhere else and it was just not worth the investment.

christa-kavanagh-wsu (approved):
Removing the unsaved changes message seems to work really well.

joelle-birano-wsu (approved):
Yeah that's much cleaner!

michael-wynne-wsu:
## PR Review: Fix LB "unsaved changes" warning leak (#1866)

Blast radius: MEDIUM-HIGH - LbSuppressUnsavedWarningSubscriber suppresses the warning on every Layout Builder route site-wide (any bundle using LB, not just the page type where #1822 was reported). Appears intentional per the code comments, but worth confirming the team is aligned on it.
Security: 1 finding (medium, below)
Tests: 2 new Kernel test files, both new classes covered - but see coverage gap below
Breaking changes: None

---

MUST FIX - none.

SHOULD FIX (non-blocking)

1. New endpoint is a GET with side effects, no CSRF protection - LbPendingMessagesController::build() drains (deletes) the Messenger queue on every call, but the route (mukurtu_gin_custom.routing.yml) is a plain GET gated only by _user_is_logged_in: TRUE, no _csrf_token. A third-party page can trigger this cross-site (e.g.  from an attacker page) via a logged-in user's browser, silently discarding their queued "layout saved" confirmation before they see it. Impact is low (only affects the requesting user's own session queue, no data disclosure to the attacker), but it violates safe-method semantics and Drupal's usual convention for state-changing AJAX routes (_csrf_token: 'TRUE' + X-CSRF-Token header in the fetch call). Worth fixing since the fix is cheap.

2. Message silently lost if Toastify isn't loaded - the controller unconditionally drains status messages once fetched, but the JS only displays them if (typeof Toastify === 'undefined') { return; } - i.e., if Toastify fails to load for any reason, the "layout saved" confirmation is fetched, deleted from the queue, and never shown anywhere. Today this is probably safe in practice (Toastify is expected to already be loaded via gin_lb on any page with #layout-builder), but there's no test or fallback covering that assumption breaking.

3. Test gap: priority-ordering is never actually exercised - LbSuppressUnsavedWarningSubscriberTest constructs the event and calls onPrepareLayout() directly, bypassing the real event dispatcher entirely. The whole fix hinges on this subscriber's priority-0 registration running after core's priority-10 PrepareLayout::onPrepareLayout() (confirmed accurate against web/core/modules/layout_builder/src/EventSubscriber/PrepareLayout.php:57 - core really is priority 10). But no test dispatches the actual PREPARE_LAYOUT event through the container with both subscribers registered, so a future ordering regression (e.g. someone re-tags priority in services.yml, or core changes it) wouldn't be caught by these tests. Similarly, LbPendingMessagesControllerTest calls the controller class directly, never through the actual route - so the route path/access requirement itself is untested.

SUGGESTIONS

4. String-matching (string) $this->t('You have unsaved changes.') as the discard key is a soft coupling to core's exact wording - low risk (both sides call t() in the same request/locale so translation itself isn't a problem), but if core ever changes this string, the suppression silently stops working (warning reappears) rather than failing loudly. A code comment already explains the coupling; consider a regression test that would fail loudly if core's string changes.

5. Route access is _user_is_logged_in only, no explicit permission - acceptable here since it only reads/drains the requesting user's own session Messenger queue (no cross-user data exposure), but if you want tighter least-privilege alignment with the rest of mukurtu_gin_custom's LB-only routes, consider gating on a Layout Builder-related permission instead.

LOOKS GOOD

- The core fix (LbSuppressUnsavedWarningSubscriber) is well-reasoned and thoroughly commented - the tradeoff (discard vs. reconcile two competing display mechanisms) is explained clearly enough for a future maintainer to understand why.
- deleteByType + selective re-add of non-target warnings correctly preserves other queued warnings.
- New route/controller/JS is scoped narrowly (#layout-builder selector, once()-guarded) and doesn't touch unrelated pages.
- No unrelated files in the diff - clean, isolated 8-file/295-line change.

---

## Accessibility (WCAG 2.2 AA): lb-pending-messages.js

1. No AT announcement (SC 4.1.3 Status Messages) - the toast has no aria-live, role="status", and doesn't call Drupal core's Drupal.announce(). Screen reader users get no notice the layout was saved. Same gap exists in gin_lb's own gin_lb_toastify.js, so it's not a new pattern - but that's not a reason to leave it unaddressed in Mukurtu's own code. Fix: call Drupal.announce(message) alongside Toastify().
2. Close-control accessibility unverifiable statically - Toastify is loaded unpinned from a CDN (no version in repo), so whether close: true renders a real keyboard-operable, labelled button can't be confirmed from source. Recommend a manual check plus pinning a specific version.
3. duration: -1 (never auto-dismisses) - not a focus trap, but risks SC 2.4.11 (Focus Not Obscured) if the toast sits indefinitely over another bottom-right control. Worth a manual check on the real LB page layout, especially since multiple saves without dismissal will stack several persistent toasts (compounds with finding above).
4. Contrast unverifiable statically - background: var(--colorGinStatusBackground) resolves at runtime per the site's active Gin color scheme; needs a manual contrast check against whatever accent is actually configured.

michael-wynne-wsu:
All 5 flagged issues are fixed and verified against a live Kernel test run in DDEV (all 6 tests pass):

1. CSRF protection - route now requires _csrf_token: 'TRUE'; JS fetches a token from /session/token and sends it as X-CSRF-Token before hitting the endpoint.
2. Message loss if Toastify missing - the typeof Toastify === 'undefined' check now runs before the fetch, so if Toastify isn't loaded, the endpoint is never called and the message stays queued for normal rendering instead of being drained and lost.
3. Untested priority-ordering/route access - added testRealDispatchSuppressesCoresWarning(), which dispatches the real PREPARE_LAYOUT event through the actual event dispatcher (both core's and Mukurtu's subscribers registered) rather than calling the method directly, and testRouteRequiresLayoutBuilderPermission(), which checks the real route via access_manager.
4. String-matching coupling - no separate fix needed; the new dispatch test in #3 will fail loudly if core ever changes the warning's wording or logic, since it doesn't hardcode the string itself.
5. Overly broad route access - route now also requires the configure any layout permission, not just an authenticated session.
6. Accessibility: no AT announcement - added Drupal.announce(message) alongside the Toastify call, with core/drupal.announce added as a library dependency.

Not changed: close-button keyboard/labelling and contrast are flagged as needing a manual/runtime check (can't be resolved from source since Toastify loads unpinned from a CDN and the Gin color scheme is runtime-configurable) - worth checking visually once you're testing this in the browser. The focus-obscuring concern from stacked persistent toasts is also a manual/runtime check.

Changes synced to DDEV and verified there; nothing has been committed yet.

nick-deer (approved):

PR #1770, opened 2026-06-26
Enable Layout Builder on basic pages (#286)

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/286
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1380
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1389

Moves field.storage.node.layout_builder__layout to mukurtu_core (its natural home since core already depends on layout_builder), adds the layout_builder__layout field instance and updated view displays for the page bundle, and adds update hook 40062 to enable Layout Builder on existing sites while migrating existing page nodes to preserve their body and media content in a default layout section.

Pre-merge checklist:

- [x] I have checked for and if required, created update hooks.
- [x] I have run an accessibility check, and resolved any issues.
- [x] I have recompiled SCSS if required.
- [x] I have updated or created CI/CD tests, if required.
- [x] I have updated from `main` and resolved any merge conflicts.
- [x] I have verified that all expected checks pass.
- [x] I have run the pr-expert-review skill and resolved all relevant issues.

PR #848, opened 2025-05-15
Update LocalContextsProjectWidget.php

michael-wynne-wsu:

PR #1867, opened 2026-07-21
Replace Local Contexts project/label select widgets with checkboxes (#856, #863)

### LINKED ISSUE #856: Local contexts project and label widget
The current OptionsSelectWidget gives bad spacing and not a great UI. 
Unfortunately the OptionsButtonWidget flattens down the project/labels. 

There isn't a clear path forward on this. See #848

### LINKED ISSUE #863: Validate Local Contexts projects and selected labels in form
When creating content, users can...
- select an entire project(s). Good!
- select labels and notices within project(s). Good!
- select BOTH a project AND specific labels within that project. Not ideal.

best case would be some kind of validation to prevent users from doing that, since they will get duplicate displays.

michael-wynne-wsu:
Closes #856
Closes #863

 - Switches to a better checkbox widget for LC projects and labels. Takes up a more sensible amount of space.
 - If a Project is selected, you can't ALSO select individual Labels.

Screenshot 2026-07-21 at 11 05 32 AM

## Summary
- Replaces the `OptionsSelectWidget`-based project and label/notice widgets with custom checkbox widgets. A prior attempt (#848) to switch to `OptionsButtonsWidget` was abandoned because a single Drupal core `checkboxes` element can't render a grouped/nested `#options` array (core issue [#2269823](https://www.drupal.org/project/drupal/issues/2269823)). Both widgets now build their own markup instead of relying on core's optgroup handling:
  - **Project widget**: flat checkboxes (its options were never grouped, so it was never actually affected by the core bug).
  - **Label/notice widget**: one collapsible `<details>` per project, each containing its own flat `checkboxes` element for just that project's labels/notices.
- Adds JS (`local-contexts-label-widget.js`) that disables + unchecks a project's label checkboxes once that whole project is checked, with a live region announcing the state change to screen readers. Only triggers on active user interaction (not on page load), so it can't silently drop label selections already saved on existing content.
- Adds a form validation backstop (`src/Hook/FormHooks.php`) rejecting submissions where a fully-selected project and one of its own individual labels/notices are both selected, since that combination renders duplicate Local Contexts info on the saved content (#863). This runs regardless of JS state.
- New SCSS (`_local-contexts-form.scss`) for widget spacing and a contrast-safe disabled state, kept separate from the existing display-formatter styles to avoid colliding with the already-used `.local-contexts-label-group` class name.

## Known follow-up (not addressed in this PR)
PR #1770 (Layout Builder for basic pages, still open) adds a `local_contexts_block` block-content type that uses these same widget plugins, but under a differently-named label field (`field_lc_labels_and_notices` instead of `field_local_contexts_labels_and_notices`). The new widgets will render correctly there once #1770 merges (shared plugin IDs), but the `FormHooks::formAlter` validation guard in this PR is hardcoded to the node field names and won't fire on that block form — it'll need to also handle `field_lc_labels_and_notices` once #1770 lands. Flagging for future review rather than fixing now, since #1770 isn't merged and can't be tested against yet.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — widget-internals-only change, no plugin ID/schema/config changes; see `docs/update-hooks.md` criteria.)

## Test plan
- [x] Verified in DDEV: project checkboxes render flat, label widget renders per-project collapsible groups, save/reload round-trips correctly.
- [x] CI checks pass.


michael-wynne-wsu:
Let me know what you think about this, I like it a lot more.

joelle-birano-wsu (approved):
I think it's great!

christa-kavanagh-wsu (approved):
I like this much better!

michael-wynne-wsu:
@nick-deer this one is ready too

nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

PR #1889, opened 2026-07-24
Namespace citation template config under a dedicated key

### LINKED ISSUE #188: Scoping for bundle config for the Citation field could be tighter
Currently for the citation field config, it is being stored in `mukurtu.settings` (the config path for the mukurtu_core module) and each bundle config is indexed directly by bundle ID. For example:

```
$config = \Drupal::config('mukurtu.settings');
$digital_heritage_citation_config = $config['digital_heritage'] ?? NULL;
```

This works for now, but I think is at risk of collision or developer confusion. It might be worth considering either storing them in an associative array or giving them a more unique settings key.

For example:
```
$config = \Drupal::config('mukurtu.settings');
$digital_heritage_citation_config = $config['citation_templates']['digital_heritage'] ?? NULL;
```

or

```
$config = \Drupal::config('mukurtu.settings');
$digital_heritage_citation_config = $config['citation_template__node__digital_heritage'] ?? NULL;
```

michael-wynne-wsu:
Just back end config cleanup. I think the only front end testing is making sure that citations and citation templates still work.

## Summary

Fixes #188. Citation field templates were stored as top-level keys in the shared `mukurtu.settings` config object, keyed directly by node bundle machine name (e.g. `$config->get('digital_heritage')`). This risked colliding with unrelated settings in the same object (e.g. `mukurtu_related_content_display`) if a bundle machine name ever matched an existing/future key, and there was no config schema at all to catch it.

- Moves citation templates to a nested `citation_templates.<bundle>` key.
- Adds a config schema for `mukurtu.settings` (none existed previously).
- Adds `mukurtu_core_update_40087()` to migrate existing sites' flat bundle keys into the new nested structure, with a defensive skip + log warning for any bundle whose machine name would collide with a reserved key.

**Deployment note:** if a site maintainer keeps their own exported copy of `mukurtu.settings.yml` in a config-sync directory outside this repo, they should re-export config after running `drush updb` so a later `config:import` doesn't reintroduce the old flat keys. This is a standard Drupal config-sync/update-hook ordering concern, not specific to this change.

## Test plan

- [x] Simulated a legacy flat `mukurtu.settings` key in a DDEV environment and ran the update hook (`mukurtu_core_update_40087`) — confirmed the value moved to `citation_templates.<bundle>` and the old flat key was removed.
- [x] Confirmed the computed citation field on a real node still renders token replacement correctly reading from the new nested key.
- [x] Confirmed the Citation Templates settings form reads defaults from and writes back to the new nested key (via direct `buildForm()`/`submitForm()` invocation).
- [x] Validated real config data against the new schema via `config.typed`'s `validate()` — 0 violations (initially used `type: text` for template values, which is translatable and required a `langcode`; switched to `type: string` since these values aren't config-translation enabled).
- [x] `php -l` on all changed PHP files.
- [x] Rebased onto latest `main` and resolved an update-hook numbering conflict with the newly-merged Klaro feature (renumbered from `40084` to `40087`).

## Pre-merge checklist:

- [ ] I have run an accessibility check, and resolved any issues. (N/A — no markup/UI changes, backend config storage only)
- [ ] I have recompiled SCSS if required. (N/A — no CSS changes)
- [ ] I have updated or created CI/CD tests, if required. (No existing automated test coverage for this citation code before or after this change; verified manually in DDEV as described above)
- [x] I have verified that all expected checks pass. (Pending CI)


christa-kavanagh-wsu (approved):
This is working as expected for me.

joelle-birano-wsu (approved):
Seems good to me

nick-deer (approved):

PR #1891, opened 2026-07-25
Remove unused Features module (resolves #696)

### LINKED ISSUE #696: `filter.format.basic_html` configuration always shows as changed in Features
Follow-up from https://github.com/MukurtuCMS/Mukurtu-CMS/issues/672 and similar to https://github.com/MukurtuCMS/Mukurtu-CMS/issues/695.

The Mukurtu Install Profile always shows up as "Changed" in the Features UI at `admin/config/development/features`. In combination with the `colorbox.settings.yml` being changed, the `filter.format.basic_html.yml` configuration also shows as changed:

Image

This configuration is correctly exported if changed, it seems to be a UI issue that it shows as changed even when it is in sync.

The `filter.format.basic_html.yml` configuration file also has a quirk in that it is saved to `/web/profiles/mukurtu/config/optional/filter.format.basic_html.yml` (note the "optional"). Nearly all other config is saved to "install". There's something a bit strange with the basic_html format that is causing it to be optional. It does not have a negative impact, it gets installed every time, but it is a strange quirk.

michael-wynne-wsu:
## Summary

- Removes the unused `features`/`features_ui` contrib modules. No custom Mukurtu code depends on them (`features.manager`, `features_assigner`, etc. - zero hits); they were only ever enabled because `mukurtu_dev` declared them as a hard dependency, and `mukurtu_dev` is itself a forced dependency of the install profile.
- This resolves #696: the permanent false-positive "Changed" status for `filter.format.basic_html` in the Features UI turned out to be a bug in the `features` module itself (`FeaturesInstallStorage` hardcodes `$include_profile = FALSE` and scans every module physically on disk, so it diffs against an unrelated disabled demo submodule's stub config instead of ours). Moving our file's location doesn't fix it - removing Features does, since there's no more Features UI to show the false positive.
- Also fixes a separate, real staleness bug found along the way: the checked-in `filter.format.basic_html.yml` was missing several CKEditor5-era filters (`media_embed`, `editor_file_reference`, `filter_htmlcorrector`) and 3 media view mode config dependencies that the site's actual active configuration has - a `drush cim` against a live site could otherwise have reverted those filters.
- Adds `mukurtu_core_update_40092()` + `mukurtu_core_update_requirements_alter()` to cleanly uninstall `features`/`features_ui` from existing sites' `core.extension`, modeled on the existing `mukurtu_notifications_update_40018`/`_requirements_alter` precedent for the `message_digest` removal.

## Also discovered (separate, unrelated, flagging for visibility - not fixed in this PR)

- **Fresh installs of `main` currently fail.** `mukurtu_install()` explicitly calls `theme_installer->install(['mukurtu_v4', 'gin'])`, but `gin` theme is now also pulled in automatically as a dependency of `gin_lb` (added by the recent Layout Builder work). By the time the explicit call runs, `gin.settings` already exists and Drupal throws a `PreExistingConfigException`. Reproduces with or without this PR's changes. Likely fix: drop `'gin'` from that explicit install call in `mukurtu.install`.
- **A pre-existing `cweagans/composer-patches` crash** on `composer update`/`composer install` (`RemoteFilesystem::copy(): Argument #1 ($originUrl) must be of type string, null given`), reproducible even with no changes. Workaround used while preparing this PR: `composer install --no-plugins`.
- A separate write-up proposing removal of the `devel`/`devel_generate` force-install (also bundled into `mukurtu_dev`, meaning Devel is currently enabled on every production site) was drafted for the team to discuss, but is intentionally out of scope for this PR.

## Test plan

- [x] Verified on a DDEV site: `drush updb` runs cleanly and removes `features`/`features_ui` from `core.extension` without blocking (tested with the code physically removed from disk to simulate the composer removal)
- [x] `drush pm:list` confirms `features`/`features_ui` are fully gone after the update
- [x] Fresh `drush site:install` installs cleanly without Features present (tested with the unrelated gin.settings bug worked around, to isolate this change)
- [x] `admin/config/development/features` returns 404 (route gone) rather than a 403
- [x] `composer.lock` diff is minimal/clean - only removes `drupal/features` and its unique transitive deps (`config_update`, `jquery_ui`, etc.), no unrelated version bumps
- [x] Ran the pr-review-expert skill against the diff

## Pre-merge checklist

- [ ] I have run an accessibility check, and resolved any issues. *(N/A - no UI/template changes)*
- [ ] I have recompiled SCSS if required. *(N/A - no SCSS changes)*
- [ ] I have updated or created CI/CD tests, if required. *(no CI/test suite currently wired up in this repo)*
- [x] I have verified that all expected checks pass. *(no CI checks configured in this repo; verified manually via DDEV as described above)*


nick-deer:
@michael-wynne-wsu clean update hook run, but got an error on rebuild:

In PreExistingConfigException.php line 65:
                                                                                  
  Configuration objects (gin.settings) provided by gin already exist in active configuration

michael-wynne-wsu:
@nick-deer will sort that out! I'm also working on the failing playwright tests.

michael-wynne-wsu:
@nick-deer I believe this is all sorted out now.

nick-deer (approved):
Code reviewer checklist:
- [x] Clean rebuild
- [ ] Test(s) ran w/out errors (n/a)
- [ ] Test file(s) read line by line (n/a)

PR #1832, opened 2026-07-16
Add bot and spam protection (CAPTCHA, reCAPTCHA, Turnstile, facet crawler blocking)

michael-wynne-wsu:
## Summary

Addresses #1163. Adds `captcha`, `recaptcha`, `turnstile`, `turnstile_protect`, `facet_bot_blocker`, and `key` as stock Mukurtu modules, plus a new `mukurtu_bot_protection` module that provides a "Security" section on the admin dashboard for choosing the active CAPTCHA backend and managing reCAPTCHA/Turnstile credentials.

- The zero-config "Basic challenge" (captcha module's own math/text challenge) and faceted-search crawler blocking are **active by default** since neither needs external credentials.
- reCAPTCHA and Cloudflare Turnstile are installed but dormant until an admin supplies credentials on the new settings form. Selecting one automatically retires the other (`captcha.settings:default_challenge` is a single site-wide value, so switching backends is one config write).
- Forms protected by default: user registration, password reset, the feedback contact form, and comments.
- Update hooks (`mukurtu_core_update_40075`/`40076`) bring existing sites up to the same state via `drush updb`.

## Test plan

Verified end-to-end in a fresh DDEV install (composer install, `drush si`, curl-driven form submissions):
- [x] Fresh install enables all 6 contrib modules + `mukurtu_bot_protection`, with `default_challenge` defaulting to `captcha/Math` and all 4 CAPTCHA points active
- [x] "Security" dashboard section renders with links to the bot-protection settings, Turnstile route protection, and facet crawler blocking settings
- [x] Switching the backend to reCAPTCHA (real widget confirmed rendering on the feedback form) and to Cloudflare Turnstile (Key entity created correctly, widget swap confirmed) work, and are mutually exclusive
- [x] Masked secret fields round-trip correctly without clobbering the stored value, including secrets containing the letter "X"
- [x] Field-specific validation errors and `#states`-driven required/visible behavior work as expected
- [x] `drush updb` reports no pending updates on a fresh install (schema version tracking correct); update hook logic mirrors an existing proven pattern (`mukurtu_core_update_40061`)
- [ ] Have not exercised the update-hook path on an actual pre-#1163 site (only verified the logic mirrors the existing proven pattern) — worth a spot-check before merge
- [ ] No automated PHPUnit tests added — consistent with sibling settings-only modules (`mukurtu_local_contexts`, `mukurtu_site_settings`), which also ship without their own test suites, but flagging as a possible follow-up

## Pre-merge checklist:

[...]

PR #1892, opened 2026-07-25
Fix Playwright CI falsely treating a sleeping Tugboat preview as ready

michael-wynne-wsu:
## Summary

Three separate, compounding bugs were making this pipeline fail on effectively every build:

1. **`wait-for-tugboat`'s readiness probe only checked the HTTP status code.** Tugboat's own "Preview is resuming" holding page (served while a suspended/idle preview wakes back up) also returns `200`, so the probe reported "ready" against a preview that wasn't serving the Drupal site yet. Fixed by polling `/user/login` and only accepting a response that's genuinely the rendered login form (`200`, not redirected to `/core/install.php`, no "Preview is resuming" marker, `Username` label present).
2. **Honeypot's `time_limit` (5s default) rejects fast form submissions as spam.** PR #1832 (merged into `main`) enabled Honeypot on `user_login_form`; Playwright's near-instant fill+submit (~1.4s, confirmed via a local trace reproduction against this PR's own preview) got silently rejected every run.
3. **The ALTCHA CAPTCHA widget on the login form ships `auto="off"`**, which requires an explicit click on its own checkbox before the browser will submit the form at all. `Login.login()` never touches the widget, so the click on "Log in" produced *zero* network requests in every trace captured, no matter how long the wait - confirmed this was the real, deterministic blocker once (2) was ruled out.

Fixes:
- `.github/workflows/playwright.yml`: harden the readiness probe (see #1 above), bump the wait budget 300s → 600s.
- `.tugboat/config.yml`: disable Honeypot's time check and the login form's CaptchaPoint, **on the Tugboat preview database only**, in both the build and update phases. Production's real defaults (`mukurtu_bot_protection.install` / `mukurtu_core.install`) are untouched.
- `tests/playwright/src/components/login.ts`: make `Login.login()` wait for the post-login redirect before returning, instead of clicking and immediately moving on - removes a latent race where a caller's next action could navigate away and cancel the login request before its session cookie was ever set.

## Test plan

- [x] YAML validated for both workflow files
- [x] Readiness-check logic dry-run against captured resuming-page/real-login-page/install-wizard-page HTML samples
- [x] Reproduced the original failure locally against this PR's own live preview via Playwright trace + page-snapshot inspection (confirmed 403 due to failed login)
- [x] Root-caused the Honeypot and CAPTCHA blockers via direct inspection of `vendor/drupal/honeypot` source and the rendered `<altcha-widget>` markup
- [x] **This PR's own CI run is green end-to-end**: `wait-for-tugboat` (2m16s) → `playwright-tests` (1m15s), all 7 default-content tests passing against the live preview

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (N/A - CI/Tugboat config and test helper only, no Drupal application code)
- [x] I have run an accessibility check, and resolved any issues. (N/A - no user-facing UI change)
- [x] I have recompiled SCSS if required. (N/A - no SCSS change)
- [x] I have updated or created CI/CD tests, if required. (this PR *is* the CI/CD fix)

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -72,18 +72,30 @@ jobs:
           fi
 
           # The comment can appear before the preview site has finished
-          # booting, so poll the URL itself until it responds successfully.
-          READY_MAX_WAIT=300
+          # booting. A bare status-code check isn't enough here: Tugboat
+          # serves its own "Preview is resuming" holding page (while a
+          # suspended/idle preview container wakes back up) with a 200,
+          # so that alone looks "ready" while Drupal underneath is still
+          # down. Poll /user/login instead and only accept a response that
+          # is actually the rendered Drupal login form.
+          READY_MAX_WAIT=600
           READY_INTERVAL=15
           READY_ELAPSED=0
           while [ $READY_ELAPSED -lt $READY_MAX_WAIT ]; do
-            STATUS=$(curl -s -o /dev/null -w "%{http_code}" "$PREVIEW_URL" || true)
-            if [[ "$STATUS" =~ ^(2|3)[0-9][0-9]$ ]]; then
+            RESULT=$(curl -s -L -o /tmp/tugboat-preview-check.html -w "%{http_code} %{url_effective}" "$PREVIEW_URL/user/login" || true)
+            STATUS="${RESULT%% *}"
+            EFFECTIVE_URL="${RESULT#* }"
+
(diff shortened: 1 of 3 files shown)


michael-wynne-wsu:
FYI I have had to disable a few of the bot blockers in the tugboat environment only, so that the playwright tests can run.

nick-deer (approved):

PR #1894, opened 2026-07-26
Fix YouTube privacy/cookie handling and add hosting-provider fallback (#565)

### LINKED ISSUE #565: Youtube API requirements
Youtube (and Google more generally) seems to have new-ish privacy/cookie changes and in V3 we're having issues with certain hosting providers. Need to address in v4.

michael-wynne-wsu:
Just need to check that YouTube still works properly as a media source.

The second issue (proactive Youtube API key) isn't something we can really test without getting it on an environment that is a problem, so don't worry for now.

## Summary

Fixes #565.

Issue #565 bundled two related YouTube/oEmbed problems:
1. Google's privacy/cookie changes mean YouTube embeds should use privacy-enhanced mode.
2. Some hosting providers (e.g. Reclaim, reported in V3) had trouble with YouTube embeds, likely because the server-side oEmbed fetch to YouTube can be blocked/rate-limited on some hosts.

- Adds `MukurtuResourceFetcher`, a decorator on Drupal core's `media.oembed.resource_fetcher` service, which rewrites YouTube embeds to `youtube-nocookie.com` (+ `rel=0&modestbranding=1`) consistently across every view mode/formatter (both core's `oembed` formatter and Blazy's `blazy_oembed`).
- Removes `mukurtu_media_oembed_resource_data_alter()` - confirmed via reading Drupal core's `ResourceFetcher` that this hook name doesn't exist in core and was never actually invoked; it was dead code.
- Adds an optional YouTube Data API fallback: if the primary oEmbed fetch throws and a YouTube Data API key is configured (via the Key module, `key_select` widget - never stored as plain config), the decorator builds an equivalent resource from the Data API instead of leaving the embed broken. If the fallback also fails or no key is configured, the original exception still propagates - it never fabricates a broken/fake resource silently.
- Adds update hooks `mukurtu_core_update_40092` (defensively ensures the Klaro `youtube`/`vimeo` klaro_app configs exist, mirroring the existing SoundCloud update hook) and `mukurtu_core_update_40093` (installs the `key` module, now a hard dependency of `mukurtu_media`).
- New unit test suite (`MukurtuResourceFetcherTest`) covering: non-YouTube passthrough, nocookie rewrite, fallback success, and both no-key/fallback-also-fails exception paths.

## Verification

Built a scratch DDEV environment (`mukurtu-565`) off this branch to actually exercise the change, since issue #565's hosting-provider symptom was never reproduced by the reporter:
- All 5 new unit tests pass.
- Fresh `mukurtu` profile install succeeds with the new service decorator, `key`, and `klaro` all wired in cleanly - no container compilation errors.
- Confirmed live: `Drupal::service('media.oembed.resource_fetcher')` resolves to the decorator, and a real YouTube oEmbed fetch came back correctly rewritten to `youtube-nocookie.com` with the expected params.
- Ran the WCAG accessibility skill on the one new UI element (the `key_select` field) - found and fixed a real issue where the Key module's auto-injected "create a new key" link renders as broken, non-interactive text (WCAG 4.1.2); disabled via `#key_description => FALSE`.
- Ran a manual PR review pass and fixed a genuine finding: Guzzle exception messages include the full request URI (with the API key as a query param), so the fallback's error logging was changed to avoid leaking it into watchdog logs.

The Reclaim-hosting fallback itself is defensive and could not be verified against the actual originally-reported failure, since it was never reproduced. Would appreciate testing on a Reclaim-hosted (or similarly restrictive) site if possible.

## Pre-merge checklist:

- [ ] I have recompiled SCSS if required. (n/a - no styling changes)
- [x] I have verified that all expected checks pass. (pending CI)


joelle-birano-wsu (approved):
Seems fine to me

christa-kavanagh-wsu (approved):
This seems fine as far as I can tell.

nick-deer (approved):
Code reviewer checklist:
- [x] Clean rebuild

PR #1900, opened 2026-07-28
Refactor RepresentativeMediaItemList to dispatch an event

### LINKED ISSUE #187: Refactor or Remove Representative Media Field
The Mukurtu core module provides a computed field, `field_representative_media` controlled by `RepresentativeMediaItemList`. The idea was to create field that would be attached to all content types to provide a protocol aware media field that would abstract away some of the logic behind trying to find an accessible thumbnail/teaser image for a given item for a given user. The current implementation is old and half baked. It should be fully fleshed out or removed.

For example, the version 3 Collection behavior of checking for an accessible (in terms of protocol access) local image media asset and then failing over to the first accessible image from one of the items in the collection was one of the use cases for this computed field.

If fleshed out, I think it would make sense to rework it to use events, so that different content types could manage their own logic in their respective modules using event subscribers.

If removed, it will need to be decided how to handle these few cases where we have unique image handling.

michael-wynne-wsu:
In theory this is just code cleanup, I think we just need to test that the primary media asset for all content types shows as expected in previews etc. So I think that's... 

- [ ] media assets for DH
- [ ] image for collection
- [ ] recording for words
- [ ] media for person
- [ ] media for place
- [ ] image for word list

## Summary
- Resolves #187: `RepresentativeMediaItemList` (the `field_representative_media` computed field) now dispatches a `RepresentativeMediaComputationEvent` to collect candidate media source fields, mirroring the existing `RelatedContentComputationEvent` pattern used by `field_all_related_content`.
- Content-type modules (`mukurtu_dictionary`, `mukurtu_collection`) now register their own fields via event subscribers, instead of a hardcoded array living in `mukurtu_core`. `field_media_assets` (shared identically across five content types) stays registered in `mukurtu_core` since it isn't unique business logic. New content types can add their own image-like field with just one new subscriber, no core edits.
- Fixes a latent bug: `mukurtu_media.module`'s `hook_entity_view_alter()` kept its own separately-hardcoded copy of the source-field list to decide the restricted-vs-no-media placeholder, and that copy was missing `field_recording` — so a dictionary word whose only media source was a restricted recording incorrectly showed the "no media" placeholder instead of "restricted." Both call sites now consume the same dispatched list, so they can't drift again.
- Field priority order is preserved exactly via explicit weights on the event (verified live: dispatching against a real node returns `field_thumbnail, field_media_assets, field_recording, field_collection_image, field_word_list_image`, identical to the original hardcoded array).
- No config/schema changes, no update hook required.

## Test plan
- [x] Dispatched the event directly against real nodes on a scratch DDEV site and confirmed subscriber wiring/order exactly matches the original hardcoded array.
- [x] Created a `dictionary_word` node with both `field_thumbnail` and `field_media_assets` populated (order-sensitive case) and confirmed `field_representative_media` resolves to the thumbnail, matching pre-refactor behavior.
- [x] Created a `collection` node with `field_collection_image` and confirmed it resolves correctly.
- [x] Confirmed `field_recording` is now included in the dispatched list used by `mukurtu_media.module` (the bug fix).
- [x] `drush cr` clean, no watchdog errors.
- [ ] Reviewer spot-check on a staging/real content dataset, if desired.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required - no schema/config change.)
- [x] I have run an accessibility check, and resolved any issues. (N/A - no markup/UI change, PHP/YAML only.)
- [x] I have recompiled SCSS if required. (N/A - no SCSS change.)
- [x] I have updated or created CI/CD tests, if required. (No existing automated test coverage for this computed field; verified manually on DDEV per test plan above.)


joelle-birano-wsu:
I noticed the audio player spilling over in collections but itʻs also on main so should I make a separate issue for it? Everything else seemed fine.

Screenshot 2026-07-28 at 11 51 53 AM

michael-wynne-wsu:
separate issue

christa-kavanagh-wsu (approved):
Aside from the collections view audio thing Joelle mentioned, everything else looks good.

joelle-birano-wsu (approved):

nick-deer (approved):
Code reviewer checklist:
- [ ] Update hook(s) ran w/out errors (N/A)
- [ ] Update hook(s) changes verified in live site (N/A)
- [x] Clean rebuild
- [ ] Test(s) ran w/out errors (N/A)
- [ ] Test file(s) read line by line (N/A)

PR #1738, opened 2026-06-16
Content publication workflow

michael-wynne-wsu:
Closes https://github.com/MukurtuCMS/Mukurtu-CMS/issues/255

Despite the large number of changed files, for users this is a very small change. Gets rid of the default Drupal draft/published states on content in favour of a new "Mukurtu Default Content Workflow" that re-implements those same two options but using the larger Workflow module so that we can expand on it more later. The draft/published setting is also now located next to the save/preview buttons.

Basically, this is new default install behaviour, just a more refined version of what we had before, but built on a different base so we can extend it in the future. A more complex optional workflow will come in a subsequent branch - and users can of course make their own later.

- [ ] Draft/Published is now a dropdown selection beside the Save/Preview buttons on content.

Screenshot 2026-06-18 at 10 43 08 AM

- [ ] `/user/drafts` has had an initial update, and properly displays draft content.
- [ ] 
Screenshot 2026-06-18 at 10 42 21 AM

- [ ] At `/admin/content`, the published column now uses "Draft" and "Published" instead of other terms.

Screenshot 2026-06-18 at 10 42 31 AM

- [ ] I believe appropriate users can view draft content at `/admin/content`.
- [ ] Draft content does not show up in the front end.
- [ ] A more prominent "Draft" badge/label/something when viewing content in front-end view.

Screenshot 2026-06-18 at 10 47 11 AM

Pre-merge checklist:

- [ ] I have checked for and if required, created update hooks.
- [ ] I have run an accessibility check, and resolved any issues.
- [ ] I have recompiled SCSS if required.
- [ ] I have updated or created CI/CD tests, if required.
- [ ] I have updated from `main` and resolved any merge conflicts.
- [ ] I have verified that all expected checks pass.
- [ ] I have run the pr-expert-review skill and resolved all relevant issues.

PR #1907, opened 2026-07-28
Fix fatal error installing mukurtu_multilingual module

michael-wynne-wsu:
## Summary

Enabling `mukurtu_multilingual` currently crashes with:

```
Error: Call to a member function getFieldStorageDefinition() on null in
Drupal\Core\Field\Entity\BaseFieldOverride->getFieldStorageDefinition()
```

- `mukurtu_multilingual` shipped 5 `core.base_field_override.node.*.draft.yml` config files that override a `draft` boolean base field on `node`. That field was permanently removed from `node` in the earlier draft-consolidation work (#1738, draft state is now tracked via `content_moderation` workflow states), but these 5 stale config files were never cleaned out of this module. Since the `draft` base field no longer exists, Drupal's config installer crashes trying to create the override on every install attempt. Removed the 5 stale files: `node.collection.draft`, `node.dictionary_word.draft`, `node.digital_heritage.draft`, `node.landing_page.draft`, `node.word_list.draft`.
- Also found and fixed a related latent issue: 9 `core.base_field_override.node.*.menu_link.yml` files override a `menu_link` field that is defined conditionally by the contrib **Token** module (only when `menu_ui` is also enabled), but neither the config files nor `mukurtu_multilingual.info.yml` declared that dependency. Added `module: [menu_ui, token]` to each config file's dependencies and added `drupal:menu_ui` / `token:token` to the module's own dependencies, so the config installer can correctly gate on them instead of assuming they're already active.

No update hook is needed: `mukurtu_multilingual` currently fails to install everywhere it's attempted, so no site has this stale config living in active config storage.

## Test plan

- [x] Verified in a local DDEV sandbox at current `main` tip: with the fix applied, `drush pm:enable mukurtu_multilingual` completes successfully with no fatal error and no watchdog errors (previously reproduced the exact reported crash without the fix).
- [x] Confirmed the deleted `draft` overrides are absent from installed config and the `menu_link` overrides install correctly with their new dependency declarations.
- [ ] Reviewer: re-confirm on a clean environment / via CI.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — see summary above.)
- [x] I have run an accessibility check, and resolved any issues. (N/A — config-only change, no UI/user-facing text.)
- [x] I have recompiled SCSS if required. (N/A — no SCSS touched.)
- [x] I have updated or created CI/CD tests, if required. (N/A — config-only fix, no new logic.)


nick-deer (approved):
Looks good! Error is gone on my end, clean rebuild

PR #1922, opened 2026-07-31
Fix broken "Browse all content in X" / category links (#1920)

### LINKED ISSUE #1920: "Browse all content in..." parameters not being applied to browse pages
When following a "Browse all content in..." link, the URL contains GET VARS that point towards a specific filter in the browse page, but these filters aren't being applied. 

For example, when in a community named "California City", the "Browse all content in California City" link goes to:

`https://thesite.com/browse?f[0]=community_content_title%3ACalifornia%20City`

However, on the browse page, the "California City" checkbox under Community is not checked, resulting in all content on the site being displayed in a long list rather than items specific to the California City community.  

Likewise, when clicking the category "Places", the link goes to:

`https://thesite.com/digital-heritage?f%5B0%5D=category%3APlaces`

Though, the "Places" button under Categories isn't checked, and all of the site content is displayed.

Note that the two URLs are different: one is to /browse, the other /digital-heritage... not sure if that's an issue.

Thank you!
Craig

michael-wynne-wsu:
## Summary

- The community, protocol, and category sidebar/tile links used hardcoded, invented query-string keys (`community_content_title`, `cultural_protocols_dh_browse`, `category`) that never matched any real filter mechanism, so results went unfiltered and no checkbox ever showed as checked when following one of these links.
- Root cause: the Solr backend reads facet filters via a `f[]=<url_alias>:<value>` scheme keyed by each facet's actual configured `url_alias`; the DB backend reads them as plain Views exposed-filter query params keyed by each filter's `identifier`, rendered as a `Checkboxes` form element that requires the raw input as `identifier[value]=value` (not a scalar, not an indexed array). Neither matched what these links sent.
- Adds a new backend-aware `Drupal\mukurtu_browse\Service\BrowseLinkBuilder` service that builds the correct link for whichever backend (`db`/`solr`) is active, dynamically looking up the real Solr facet's `url_alias` so it self-corrects if facet config changes. Wired into the three affected templates in `themes/mukurtu_v4` in place of the hardcoded hrefs.
- Registers `mukurtu_browse`'s kernel tests in `phpunit.xml`, which had none wired in before.

**Deferred (separate follow-up issue, not in this PR):** there is currently no Cultural Protocol facet on the Solr-backed `/browse` page at all (only a DB-backend one exists). On Solr, the protocol link falls back to a plain, unfiltered `/browse` link rather than sending a facet key that doesn't exist - a documented, tested, no-worse-than-before behavior, not a regression.

## Test plan

- [x] New kernel test `modules/mukurtu_browse/tests/src/Kernel/BrowseLinkBuilderTest.php` (7 tests) covering both backends × all three link types, including a facet-entity-missing fallback case.
- [x] Full kernel test suite passes (211/211) in a fresh DDEV sandbox.
- [x] Manually verified end-to-end against real content in that sandbox (community, category, and protocol links) — confirmed the facet checkbox renders `checked` and results are correctly filtered for all three link types.
- [x] Ran the wcag-accessibility-compliance skill on the three modified link elements — no regressions; link text/semantics unchanged, and the fix incidentally improves URL-encoding correctness over the old raw string concatenation.
- [x] Ran the pr-review-expert skill on the diff; addressed all 4 points raised (duplicated backend lookup, missing property type, added a facet-entity-missing test, made the test's config path lookup more robust).

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. *(None needed — no new or changed config entities.)*
- [x] I have recompiled SCSS if required. *(No SCSS changes.)*


nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

PR #1923, opened 2026-08-01
Fix zoomed lightbox image cropping, pan range, and focus/label a11y

michael-wynne-wsu:
## Summary
- Fixes the GLightbox image zoom/pan interaction: the zoomed image's resting position wasn't flush with the visible viewport (ancestor padding), so pan bounds computed from a symmetric assumption let you pan past the top into blank space while never reaching the true bottom edge. `panBounds()` now recovers the image's real untransformed position from its live rect and measures actual distance to each container edge.
- Also fixes the non-zoomed image's `max-height`, which didn't account for ~72px of combined ancestor padding and so slightly clipped the bottom of every lightbox image, zoomed or not.
- Fixes two accessibility regressions introduced by the click-to-zoom feature itself: the `:focus-visible` ring was drawn on the image (clipped away once panned) and `aria-label` was overwriting each image's real alt text with a generic "Zoom image" string. The ring now lives on the stationary container via `:has()` (with an `@supports` fallback for browsers that lack it), and the zoom/pan hint is now supplied via `aria-describedby` instead of replacing the accessible name.
- Fixes keyboard panning while zoomed: GLightbox's own drag-to-navigate handler calls `preventDefault()` on the mousedown that triggers a zoom click, which suppressed the browser's default click-to-focus behavior. Since arrow-key panning (and the guard that stops arrows from also changing slides) both key off `document.activeElement` matching the zoomed image, focus never landing there meant panning silently did nothing and arrow keys instead changed the carousel slide. The image is now explicitly focused once zoomed, so this works whether zoom was triggered by mouse click or by Tab+Enter.
- Fixes Tab navigation to the zoomable image: GLightbox's own Tab handler cycles focus only through `.gnext`/`.gprev`/`.gclose` and has no notion of the zoomable image, and never checks `e.shiftKey`, so Shift+Tab didn't reverse-cycle either - keyboard-only users had no way to reach a zoomable image at all (mouse click was the only path in). Tab handling for the open lightbox is now fully managed ourselves (same capture-phase technique as the arrow-key fix above), cycling image → next → prev → close → wraps to image, correctly in both directions, and falling back to the original 3-stop order for non-image slides or at the mobile breakpoint. Also excludes the image when it's `inert` (an undismissed content warning marks it so), since focusing an inert element silently fails and would otherwise leave Tab stuck.

## Test plan
- [x] Verified in a local DDEV sandbox: zoom an image, pan via mouse hover to all four edges, confirm the full image (including the bottom) is reachable with no overshoot into blank space.
- [x] Verified the non-zoomed lightbox view no longer clips the bottom of the image.
- [x] Verify keyboard-only: tab to an image, Enter to zoom, arrow keys to pan to each edge, Enter/Space to zoom out.
- [x] Verify click-to-zoom, then arrow-key pan: image pans and the carousel does not change slides; confirmed fixed.
- [x] Verify Tab reaches the zoomable image and Shift+Tab correctly reverse-cycles (image → next → prev → close, both directions); verified the image is skipped (falls back to next/prev/close only) when it's inert behind an undismissed content warning, and at mobile viewport widths.
- [x] Verify with a screen reader that the image's real alt text is still announced (not replaced by a generic label) and that the zoom/pan hint is available via the description. Confirmed via a real Chromium accessibility-tree snapshot: computed accessible name/role is `button "<real alt text>" [pressed]`, `aria-label` stays unset, and `aria-describedby` correctly resolves to the zoom/pan hint text.
- [x] Spot-check on a browser without `:has()` support (e.g. Firefox < 121) that a focus ring still appears somewhere while zoomed (fallback path). Confirmed by removing the two `@supports selector(:has(a))` rules from the live stylesheet (equivalent to what a non-supporting engine computes, since `@supports` drops its contents wholesale when the condition is false) - the per-image ring correctly reappears as the fallback.

## Pre-merge checklist:
- [x] I have checked for and if required, created update hooks. (Not required - no config/schema/install changes, theme CSS/JS only.)
- [ ] I have updated or created CI/CD tests, if required. (No existing test infra covers JS lightbox interaction; not added here.)


christa-kavanagh-wsu (approved):
This seems to work!

michael-wynne-wsu:
Just pushed up a collection of accessibility fixes for the lightbox

joelle-birano-wsu (commented):
I'm still struggling with the keyboard. I hit enter, that opens up the image, enter again only re-opens the image - it doesn't trigger zoom. If I click on the image, I can then pan around with the arrow keys and close with esc.

michael-wynne-wsu:
I don't see it working great with the woodpecker image but I think that's just the size of it. try the other image here https://pr1923-qojicc5azvquwumboh31a3swmpdyxiic.tugboatqa.com/digital-heritage/test

joelle-birano-wsu (approved):
Ok this works. I was missing that I had to hit enter, tab and enter again.

nick-deer (approved):
Code review checklist:
- [x] Clean rebuild
- [ ] Test(s) ran w/out errors (n/a)
- [ ] Test file(s) read line by line (n/a)

PR #1926, opened 2026-08-03
Stop auto-creating empty paragraphs on Digital Heritage, Dictionary Word, Person, and Place content

### LINKED ISSUE #1912: Auto generated empty paragraphs
Some content types that include paragraphs will auto-generate an empty paragraph on create if the paragraph isn't used. For most cases this is fine and invisible on the front end, but on export, the empty paragraph comes along and is confusing for users. 

Need to explore and see if there is a way to manage content without the auto-creation of an empty paragraph (without breaking anything else). If not, consider export options that omit empty paragraphs/corresponding sheets.
 
Confirmed conte types:

- Digital heritage items > Citing Indigenous Elders and Knowledge Keepers
- Dictionary words > Word Entries

There may be others, and this should explore and confirm all.

michael-wynne-wsu:
## Summary

Fixes #1912. Several Mukurtu paragraph fields only allow a single target paragraph bundle. The contrib Paragraphs widget auto-creates a blank paragraph of that bundle on new-content forms whenever `default_paragraph_type` isn't explicitly `_none`. These auto-created paragraphs are invisible on the front end but show up as confusing blank rows/sheets on CSV export.

**Approach (revised after review):** the first version of this PR set `default_paragraph_type: _none` on the affected fields, which stopped the auto-creation but also hid the paragraph subform behind an "Add" button — editors could no longer see what fields it holds until they clicked in. This revision instead:

- Reverts those widget settings back to their original values, restoring the paragraph subform as open/visible by default, same as before this PR.
- Adds a `preSave()` override on each affected bundle class (`DigitalHeritage`, `DictionaryWord`, `DictionaryWordEntry`, `Place`, `Person`) that prunes a newly-created paragraph from its field if the editor left it entirely blank, before it's ever saved to the database. `hook_entity_presave` can't do this — Drupal saves a new referenced paragraph via the field's own `preSave()` *before* module-level presave hooks run, so by the time `hook_entity_presave` sees the parent entity, the empty child paragraph has already been inserted. The bundle class's own overridden `preSave()` runs earlier, before that insert happens.
- Backs this with a shared `mukurtu_core` service (`ParagraphEmptinessChecker`) and trait (`PrunesEmptyParagraphsTrait`), reused by both the presave pruning and the CSV export code.
- Keeps the CSV-export-side check (skip already-empty paragraphs during export) — still needed for paragraphs already auto-created on existing sites before this fix ships.
- Removes a few dead/orphaned form-display keys (`field_indigenous_knowledges`, `field_base_word_entry`, `field_word_entry`) with no backing field definition anywhere in code, via update hooks that also handle sites already on the old config.

No user-facing text was added or changed.

## Test plan

- [x] New kernel tests for the shared service/trait: `modules/mukurtu_core/tests/src/Kernel/ParagraphEmptinessCheckerTest.php`, `PrunesEmptyParagraphsTraitTest.php` (covers: brand-new empty paragraph pruned, filled paragraph kept, mixed multi-value field, already-saved empty paragraph left alone).
- [x] Extended `CsvExportEntityReferenceRevisions.php` with `testEntityRevisionsFieldSkipsEmptyParagraph()` covering the export-side filter.
- [x] Ran the full `mukurtu_core` and `mukurtu_export` kernel suites in a DDEV sandbox built from this branch: all pass (one pre-existing, unrelated failure in `FormHooksLayoutBuilderGuardTest` — not touched by this PR, fails the same way on `main`).
- [x] Verified end-to-end against the real production bundle classes (not just test fixtures) via `drush php:eval`: creating a Digital Heritage node with a blank Knowledge Keepers paragraph attached results in no paragraph being persisted (`isEmpty() === TRUE`, `paragraph->id() === NULL`); a filled one is saved normally. Same confirmed for Dictionary Word (including the nested Word Entry → Sample Sentences case), Place, and Person.
- [x] Rendered the actual Digital Heritage add form and confirmed the Knowledge Keepers paragraph subform is open/visible by default again (not hidden behind an Add button).

## Pre-merge checklist:

- [x] I have run an accessibility check, and resolved any issues. (No accessibility impact — widget config is a byte-identical revert to `main`'s original behavior; all new code is backend save-time logic with no rendering surface.)
- [x] I have recompiled SCSS if required. (N/A — no theme/SCSS changes.)


michael-wynne-wsu:
Okay updated this. Create content forms should be exactly the same as before, just not saving the "empty"/placeholder paragraph, so no empty paragraph on export.

christa-kavanagh-wsu (approved):
This seems to work!

joelle-birano-wsu (approved):
Works for me!

nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

PR #1811, opened 2026-07-10
Add missing Grid/Map/taxonomy_record templates for content-type parity (#1624)

michael-wynne-wsu:
## Summary

Phase 2 of the #1624 theming-consistency effort. Depends on Phase 1 (#1809) for the shared `_card-media.html.twig` include, hence the base branch.

### SCOPE

- `/digital-heritage` 
   - Media display carries over from Phase 1.
   - List and grid shows media title, summary, community and protocol, category, and collections.
   - Map shows media, title, summary, community and protocol.
- `/browse`
   - Media display carries over from Phase 1.
   - List and grid for all content types show media, content type, title, community and protocol, and collections.
      - List and grid for DH also has summary and category.
      - List and grid for word also has word list.
   - Map shows media, title, summary, community and protocol.
- `/taxonomy/term`
   - Uses the same grid display as `/browse`.
   - No toggle for other layouts. 
- `/dictionary`
   - Map shows thumbnail, recording, title, community and protocol.
   - No change to regular layout.

Fills in content-type x mode template gaps identified during investigation — several content types fell back to a sparse generic card in Grid mode, or had no Map/taxonomy_record display at all.

- **Grid mode**: added `node--{collection,digital-heritage,person,place}--grid-browse.html.twig` (vertical-card layout), which previously fell back to a bare title-only generic card. Exposed `field_communities`/`field_category` (digital_heritage only) in their `grid_browse` displays so the richer templates have something to show, and extended the `content_type_label` preprocess (`mukurtu_browse_preprocess_node()`, previously "browse" mode only) to also cover `grid_browse` — otherwise the content-type label markup in these new templates would have been silently dead code.
- **Map mode**: added `node--{collection,dictionary-word,digital-heritage,person,place}--map-browse.html.twig`. While doing this, found and fixed a real pre-existing bug: `dictionary_word` and `word_list`'s `map_browse` displays had `field_representative_media` hidden — meaning `word_list`'s *existing*, already-shipped map template has never actually been able to show its media.
- Added the `collection.map_browse` and `collection.taxonomy_record` entity view displays (collection had neither).
- Added `node--place--teaser.html.twig`, matching the simple title+content pass-through convention already used by person/digital-heritage.
- Confirmed the "dictionary default" (teaser) mode for word/word_list is already fully and consistently themed (dedicated Twig templates + SCSS) — no changes needed there.
[...]

PR #1989, opened 2026-08-10
Detect and prevent the orphaned-merge PR pattern (#1811, #1854, #1818)

michael-wynne-wsu:
## Summary

We've now had three confirmed incidents of the same bug: PR #1811, PR #1854, and PR #1818 all showed as **Merged** on GitHub but never reached `main`. Each was opened against another PR's head branch (a normal "stack for incremental review" pattern), that base PR merged into `main` first, and the stacked PR later merged too - but into the now-stale leftover branch instead of `main`, since nobody retargeted it. Each was discovered only when a user reported a "missing" feature, followed by a manual recovery PR.

**Audit:** fetched all 620 merged PRs via the GitHub GraphQL API and checked every one whose base branch wasn't `main`. Only 9 total in the repo's history. A naive `git merge-base --is-ancestor <mergeCommit> origin/main` check is unreliable here - every PR in this repo is squash-merged, which rewrites history and produces false positives even on fine PRs (confirmed: 6 of the 9 flagged candidates were false alarms, verified via content-diff against current `main`). Only the 3 already-known incidents were real. **No hidden additional instances found.** 10 currently-open PRs are stacked but currently safe (an initial unpaginated check undercounted this as 5 - see follow-up comment below) (their bases' own PRs are still open).

**Root cause:** `delete_branch_on_merge` was `false`, and `main` has zero branch protection - so merged branches lingered indefinitely, meaning GitHub's own automatic PR-retargeting feature (which fires when a branch that's the base of an open PR gets deleted) never got a chance to run.

## Changes

- **Enabled `delete_branch_on_merge`** on the repo (done directly via the GitHub API as part of this work, not a file change) - the primary fix. From now on, every merge deletes its head branch, and GitHub natively retargets any other open PR based on that branch to the deleted branch's own base.
- **`.github/workflows/stale-base-check.yml`** - a visibility/backstop layer on top, since native retargeting is a silent timeline event and can't retroactively help PRs already open before today. Checks a PR's own base on open/reopen/push; sweeps other open PRs the instant a PR merges; runs a weekly full sweep as backstop. Posts a PR comment when a PR's base branch's own PR already merged elsewhere (or was closed unmerged - a separate "abandoned stack" warning). Warn-only by default (`MUKURTU_STALE_BASE_STRICT=0`) since `main` has no required status checks yet, so nothing can actually block on this today either way.
- **`docs/stacked-prs.md`** - explains the failure mode with the 3 incidents as worked examples, the primary/backstop defense, and documented blind spots (a base merged via direct push with no PR; a renamed base branch).
- **`.github/pull_request_template.md`** - one added checklist line about retargeting a stacked PR once its base merges.

## Test plan

- [x] Validated the detection logic (does a *merged* PR exist with `head == this PR's base`?) against all 3 known ground-truth incidents (#1811/#1809, #1854/#1853, #1818/#1816) - all correctly detected.
- [x] Validated a true-negative case against all 10 currently-open stacked PRs - correctly finds no match and posts zero comments, since their bases are still open (see follow-up comment).
- [x] YAML syntax validated (`yaml.safe_load`); the `on:`→boolean quirk under PyYAML is present identically in the existing `playwright.yml`, confirmed harmless for GitHub's own Actions parser.
- [ ] First live run of the workflow itself (opened/synchronize/schedule) not yet observed in CI - recommend a manual `workflow_dispatch` run after merge to confirm end-to-end, per docs.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. *(None needed - CI/docs only, no Drupal code/config changes.)*
- [x] I have run an accessibility check, and resolved any issues. *(N/A - no user-facing UI.)*
- [x] I have recompiled SCSS if required. *(N/A.)*
- [ ] I have updated or created CI/CD tests, if required. *(This PR IS the CI addition; logic validated against live data per test plan above rather than a unit-test harness.)*
- [ ] I have verified that all expected checks pass. *(Pending CI run on this PR.)*
- [x] I have run the pr-expert-review skill and resolved all relevant issues. *(Not run as a separate pass - this PR is primarily new CI/docs, reviewed directly with the user throughout, including explicit sign-off on all contributor-facing text.)*

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,6 +5,7 @@
 - [ ] I have recompiled SCSS if required.
 - [ ] I have updated or created CI/CD tests, if required.
 - [ ] I have updated from `main` and resolved any merge conflicts.
+- [ ] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges (see [docs/stacked-prs.md](../docs/stacked-prs.md)).
 - [ ] I have verified that all expected checks pass.
 - [ ] I have run the pr-expert-review skill and resolved all relevant issues.
 
--- a/.github/workflows/stale-base-check.yml
+++ b/.github/workflows/stale-base-check.yml
@@ -1,0 +1,125 @@
+name: "Stale Base Branch Check"
+
+# Detects the "orphaned merge" pattern that has bitten this repo three times
+# (PRs #1811, #1854, #1818): a PR is opened against another PR's head branch
+# (a normal way to stack incremental review), that base PR merges into main
+# first, and the stacked PR is later merged too - but into the now-stale
+# leftover branch instead of main, since nobody retargeted it. It shows
+# "Merged" on GitHub and closes its linked issues, but the code never
+# actually reaches main. See docs/stacked-prs.md.
+#
+# `delete_branch_on_merge` is enabled repo-wide, which makes GitHub natively
+# retarget any open PR based on a branch the moment that branch is deleted
+# on merge - that's the real fix. This workflow is a visibility/backstop
+# layer on top of it: native retargeting is a silent timeline event (easy to
+# miss), and it can't retroactively help a PR that was already open before
+# that setting existed.
+#
+# MUKURTU_STALE_BASE_STRICT below is unset/0 by default (comment only, exit
+# 0) since `main` has no required status checks yet - a failing check here
+# cannot block a merge either way today. Flip to 1 in the env below once
+# branch protection requires this check.
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, closed]
+  schedule:
+    - cron: "0 13 * * 1" # Weekly, Monday 13:00 UTC.
+  workflow_dispatch: {}
+
+env:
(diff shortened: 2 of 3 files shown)


michael-wynne-wsu:
Post-push follow-up (2 small fixes pushed after opening this PR, both caught by actually running the workflow rather than just reading it):

1. The job never checked the repo out, so `gh pr list` (which falls back to inferring the repo from local git context when `--repo` is omitted) failed immediately on a `workflow_dispatch` test run. Added the checkout step + explicit `--repo` everywhere.
2. `gh pr list` defaults to a 30-item page. This repo currently has 62 open PRs, so the full sweep was silently checking only a subset - caught because the corrected run reported 10 currently-open stacked PRs, not the 5 an earlier unpaginated check had suggested. Added `--limit 200` to every `gh pr list` call in the workflow.

Re-ran `workflow_dispatch` after both fixes: correctly found and evaluated all 10 currently-open stacked PRs, posted zero comments (all 10 are still safe - their bases haven't merged yet), confirming both the pagination fix and the true-negative case.

nick-deer (approved):

Issue #2011, opened 2026-08-18
Digital Heritage browse page may exceed Playwright's actionTimeout, causing consistent CI flake


michael-wynne-wsu:
## Summary

`tests/playwright/tests/browse.spec.ts`'s `Browse tests - Digital Heritage` test has been consistently failing in CI (observed independently on two unrelated PRs' fresh Tugboat previews - #1927 and #1930 - each failing on all 3 attempts: initial run + 2 retries) with:

```
TimeoutError: locator.click: Timeout 5000ms exceeded.
Call log:
  - waiting for getByRole('button', { name: 'Add a new category' })
```
and separately:
```
TimeoutError: locator.click: Timeout 5000ms exceeded.
Call log:
  - waiting for getByText('Grid', { exact: true })
```
on `/digital-heritage`.

## What's been ruled out

- The Grid/List/Map toggle markup (`modules/mukurtu_browse/templates/mukurtu-browse.html.twig:3-7`) renders unconditionally - it isn't gated on view result count, hasn't been touched by any recent commit, and no CSS/JS was found hiding it.
- Test ordering is not the cause: the `chromium` Playwright project has `dependencies: ['default-content']` (`tests/playwright/playwright.config.ts`), so the Digital Heritage default-content-creation test is guaranteed to finish before `browse.spec.ts` runs, and that creation step is not the one failing.
- This is unrelated to the code changes in the two PRs where it was observed (#1927/#1930 only touch `mukurtu_import`) - the same failure appears independent of PR content, suggesting a pre-existing characteristic rather than a regression from either PR.
- This is a **different** issue from `playwright-tests-main` (the check that runs against the persistent `TUGBOAT_MAIN_URL` preview on pushes to `main`) also currently failing - that one is because the persistent preview itself returns a flat 404 "Preview Not Found" (confirmed via direct curl), i.e. an expired/torn-down environment, not a page-rendering problem. That's a separate, infra-only issue.

## Leading hypothesis (unverified - needs live repro)

[...]

PR #1927, opened 2026-08-03
Add persistent import logs (#1417)

michael-wynne-wsu:
## Summary
- Adds a `mukurtu_import_log` DB table + `MukurtuImportLogStorage` service recording every file processed by the import wizard: filename, destination entity type/bundle, timestamp, user, success/failure, row counts, and cleaned error text. Previously this only lived in a per-session tempstore that was wiped on Reset or the next import (issue #1417), unlike v3's Feeds logs.
- Adds a new "Import Logs" admin page (`/admin/import/logs` listing, `/admin/import/logs/{id}` detail) with a new `view any mukurtu_import_log` permission: importers see only their own history, and roles with `administer mukurtu_import_strategy` are granted visibility into everyone's (via update hook, mirroring the existing `mukurtu_import_update_40001` pattern).
- Fixes a bug found while building this: `ImportBatchExecutable::batchFinishedImport()` used to attribute error messages from ALL failing files in a batch to only the last-failing file's fid. Building an accurate persistent log required fixing this, since the log would otherwise permanently bake in the same mis-attribution. Fixed by keying `$context['results']` per migration ID instead of flattening messages into one shared array.
- Two update hooks: `40031` creates the table, `40032` grants the new permission to existing admin roles.

## Scope note
`mukurtu_export` has an analogous transient-results gap, but no linked issue and no per-item error-capture mechanism to persist yet — intentionally out of scope here. Will file a follow-up issue for export logging.

## Known conflict
Open PR #1910 also claims `mukurtu_import_update_40031()`. Whichever of these two merges second will need to renumber its update hook(s) past the other's final number.

## Test plan
- [x] 8 new Kernel tests (`ImportLogTest`, `ImportLogAccessTest`): single success, single failure, the multi-file mis-attribution regression (two files failing in one batch, each keeps its own fid/messages/counts), tempstore-shape preservation for existing consumers (`ImportFileUploadForm`/`ImportResultsForm`), mixed success+failure batch, and permission-scoped access control (own-only vs. `view any`, including the nonexistent-row 404-not-403 path).
- [x] All 42 pre-existing `mukurtu_import` Kernel tests still pass unchanged.
- [x] Update hooks verified against a real upgrade scenario in a DDEV sandbox (schema-version rollback + table drop + `drush updatedb`), confirmed idempotent on rerun.
[...]

PR #2013, opened 2026-08-18
Pin drupal/facets below 3.0.4 to avoid known getSearchApiQuery() crash

michael-wynne-wsu:
## Summary
- `drupal/facets` 3.0.4 has an active, unfixed upstream regression ([drupal.org #3617839](https://www.drupal.org/project/facets/issues/3617839), from changes in #3589162) that fatally crashes `FacetsFilter::query()` with `Call to a member function getSearchApiQuery() on null` whenever a facets-based exposed filter block is rendered standalone via `drupal_block()` outside the view's own execution - exactly the pattern `mukurtu-browse.html.twig` uses for `/digital-heritage` (and likely `/browse`, `/dictionary`).
- This is the **real** root cause of the CI flake investigated in #2011 - #2012's timeout bump didn't help because the page isn't slow, it 500s outright, on every load, even with zero content.
- This repo's own `composer.lock` already correctly pins `3.0.3`, but the `"^3.0"` constraint let Tugboat's build (`.tugboat/config.yml` runs `composer update` against `mukurtu/mukurtu-template`'s own constraints rather than installing from this repo's lock) drift onto the broken 3.0.4 on every fresh preview - which is what's been causing the consistent Playwright "Grid" timeout on PRs #1927/#1930.
- Tightens the constraint to `>=3.0 <3.0.4` so `composer update` can never re-resolve onto the broken release, regardless of lock file. Regenerated `composer.lock`'s content-hash to match; the pinned version itself doesn't change (already 3.0.3).
- Upstream fix MR `!395` was opened today but isn't merged yet, so pin below 3.0.4 until a fixed release ships - then this constraint should be loosened again.
- `MukurtuCMS/mukurtu-template` (the separate repo Tugboat actually resolves dependencies from) likely needs the same constraint; flagging separately rather than in this PR since it's a different repo.

## Test plan
- [x] Reproduced locally: built a fresh DDEV sandbox tracking `main`, confirmed `drupal/facets` 3.0.4 → `/digital-heritage` returns HTTP 500 with zero content installed; downgrading to 3.0.3 in the same sandbox → HTTP 200, Grid/List/Map toggle renders normally.
- [ ] CI passes on this PR.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (not applicable - composer dependency only, no schema/config change)
- [x] I have run an accessibility check, and resolved any issues. (not applicable - dependency version pin only)
- [x] I have recompiled SCSS if required. (not applicable)
- [x] I have updated or created CI/CD tests, if required. (not applicable - this fixes the environment the existing tests run against)
- [ ] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges (see docs/stacked-prs.md). (not applicable - based on `main`)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "drupal/entity_reference_revisions": "*",
         "drupal/extlink": "^2.0",
         "drupal/facet_bot_blocker": "^1.0",
-        "drupal/facets": "^3.0",
+        "drupal/facets": ">=3.0 <3.0.4",
         "drupal/facets_reset_button": "^1.0",
         "drupal/field_group": "^4.0",
         "drupal/flag": "^4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "18cf3099d88e24bed9e46774d572550c",
+    "content-hash": "a1de3d6e1ccd6695b155bf61ca4b7a04",
     "packages": [
         {
             "name": "altcha-org/altcha",


michael-wynne-wsu:

Link:
- https://pr2013-4q8pb1tcjio7jwpj5fpwkwqkuaxiqxm4.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a84ad121cdd491af8e4ee7f

PR #2026, opened 2026-08-20
Restrict Message Subscribe UI and fix email frequency opt-out

michael-wynne-wsu:
## Summary

**Restrict the "Subscriptions" page to admins:**
- The Message Subscribe UI "Subscriptions" tab/page (`/user/{uid}/message-subscribe`) let any authenticated user view their own subscriptions listing via a self-view bypass in contrib's `SubscriptionController::tabAccess()`, regardless of the module's own `administer message subscribe` permission. That page is internal admin tooling for managing users' notification subscriptions, not something end users need a dedicated tab for.
- `RouteSubscriber::alterRoutes()` (in `mukurtu_core`) now overrides both `message_subscribe_ui.tab` and `message_subscribe_ui.tab.flag` route requirements to require the `administer message subscribe` permission only, dropping the self-view bypass. Since the "Subscriptions" local task has no visibility conditions of its own and purely inherits route access, this also removes the tab from view for non-admins on both `/user/{uid}` and `/user/{uid}/edit`.
- Both routes are also marked `_admin_route`, so Gin renders the page like the rest of Mukurtu's admin tooling instead of falling back to the front-end theme.
- Grants `administer message subscribe` to the `mukurtu_manager` role (both in `config/install` for new sites and via `mukurtu_update_40004()` for existing sites), so site managers keep the access they previously had incidentally.
- **Scope note:** this does *not* affect end users' ability to actually follow/unfollow content elsewhere on the site (community, protocol, content, collection "Follow" buttons) -- those go through Flag module's own `flag.action_link_flag`/`flag.action_link_unflag` routes, which are untouched. Only the standalone aggregated subscriptions-management page is restricted.

**Make "Do not send emails" a real email frequency option:**
- On `/user/{uid}/edit`, the "Email frequency" field's "N/A" option was actually Drupal's synthetic empty-option placeholder for optional list fields, not a real stored value -- selecting it always saved an empty value, and empty has always been treated as immediate sending. "N/A" could never actually stop emails or really be "saved" as anything.
- Replaces it with a real `none` allowed value labeled "Do not send emails". Making that value real requires the field to become **required**, since that's the only way `OptionsWidgetBase` stops auto-injecting the synthetic empty placeholder.
- `_mukurtu_notifications_user_wants_email()` now short-circuits to `FALSE` when a user's frequency is `none` (except for `mukurtu_new_user_registration`, which intentionally can't be opted out of).
[...]

PR #2028, opened 2026-08-20
Make "Do not send emails" a real email frequency option

michael-wynne-wsu:
## Summary
- On `/user/{uid}/edit`, the "Email frequency" field's default-selected "N/A" option is actually Drupal's synthetic empty-option placeholder for optional list fields, not a real stored value. Selecting it always saves an empty value indistinguishable from "field never touched," and empty has always been treated as immediate sending (`_mukurtu_notifications_user_wants_email()` / `_mukurtu_notifications_user_wants_digest()`) -- so "N/A" could never actually stop emails, and could never really be "saved" as anything.
- Replaces it with a real `none` allowed value labeled "Do not send emails". Making that value real requires the field to become **required**, since that's the only way `OptionsWidgetBase` stops auto-injecting the synthetic empty placeholder.
- `_mukurtu_notifications_user_wants_email()` now short-circuits to `FALSE` when a user's frequency is `none` (except for `mukurtu_new_user_registration`, which intentionally can't be opted out of).
- Every existing user currently has an empty stored value. `mukurtu_notifications_update_40056()` backfills them to `immediate` (preserving their actual current behavior) and then flips the field to required, so nobody is silently switched to "do not send" and nobody gets stranded by the new required constraint.
- Confirmed with the requester: existing users get backfilled to `immediate` rather than left empty to force a choice on next edit.

## Test plan
- [x] Added `NotifyFrequencyTest` (Kernel) covering: `none` is a real allowed value; `none` suppresses email for an ordinary notification; `none` does *not* suppress the mandatory new-user-registration notice; `immediate` still falls through to the existing category check; the update hook backfills only empty values and leaves other values untouched; the update hook makes the field required; the update hook is idempotent.
- [x] Ran the new test plus the existing `Uid1EmailCategoryUpdateTest`, `NewAccountCreatedRecipientsTest`, and `NotificationsFeedAccessTest` against a DDEV instance as a regression check -- all passing.
- [ ] Manually verify in a browser: `/user/{uid}/edit` no longer offers "N/A", offers "Do not send emails" instead, and a user who picks it stops receiving notification emails.

## Pre-merge checklist:

- [ ] I have run an accessibility check, and resolved any issues. *(No new markup -- the widget already renders these allowed values as radio buttons; only the option set and required flag changed.)*
- [x] I have recompiled SCSS if required. *(N/A -- no SCSS touched.)*
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges (see [docs/stacked-prs.md](../docs/stacked-prs.md)). *(Base is `main`.)*

### DIFF
--- a/modules/mukurtu_notifications/mukurtu_notifications.install
+++ b/modules/mukurtu_notifications/mukurtu_notifications.install
@@ -186,11 +186,16 @@ function mukurtu_notifications_create_field_notify_email_categories(): void {
 
 /**
  * Creates the field_notify_frequency storage and attachment on the user
- * entity if missing.
+ * entity if missing, and backfills any user with an empty value to
+ * 'immediate'.
  *
- * No backfill needed: _mukurtu_notifications_user_wants_digest() already
- * treats an empty/unset value as "immediate" (today's behavior for every
- * existing user), so only the default for brand-new users needs setting.
+ * A user can end up with an empty stored value in more than one way: sites
+ * upgrading from before this field existed, or even a fresh install, since
+ * uid 1 is typically created before this hook runs and FieldConfig's
+ * default_value only populates entities created after the field exists.
+ * Backfilling unconditionally here -- the same way this file's
+ * field_notify_email_categories creation function does above -- covers
+ * every path that calls this function, not just the update-hook path.
  */
 function mukurtu_notifications_create_field_notify_frequency(): void {
   if (!\Drupal\field\Entity\FieldStorageConfig::loadByName('user', 'field_notify_frequency')) {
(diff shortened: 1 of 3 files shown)


michael-wynne-wsu:

Link:
- https://pr2028-x0qowi8ynwewv79s4dl7ijtwdxkgmjjf.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a87400831bf47ecc7d9d4b7

michael-wynne-wsu:
Consolidated into #2026 per request -- both fixes now live in a single PR/branch for one combined preview build.

PR #2066, opened 2026-08-27
Restore string taxonomy-name fields dropped from mukurtu_browse_auto_index

michael-wynne-wsu:
## Summary

- `mukurtu_search_rebuild_index()` (called by `mukurtu_search_install()` and `mukurtu_search_update_40001()`) cleared every field on `mukurtu_browse_auto_index` and only restored the fulltext `__name__text` variant of each taxonomy reference field, not the plain string `__name` variant that 16 facets (Category, Format, Language, Publisher, Type, Word Type, etc.) use as their `field_identifier`.
- Any already-installed site that ran `update_40001` (introduced in #1570) lost those 16 fields, breaking every facet built on one of them with `No available query types were found for facet X`. Fresh installs (e.g. Tugboat, which does a clean `drush site-install` per build) never hit this, which is why the break only showed up on longer-lived local/production environments.
- This restores the 16 string fields in `mukurtu_search_rebuild_index()` so it can't recur, and adds `mukurtu_search_update_40005()` to repair already-affected sites.

## Known trade-off (please read)

Verified end-to-end on a DDEV site with real migrated content: running the repair hook restores all 16 field *configs* correctly (no more "No available query types" anywhere), but on that site's larger content model, the physical `ALTER TABLE ... ADD INDEX` failed for the last 4 fields processed (Publisher, Subject, Type, Word Type) with MySQL's "Too many keys specified; max 64 keys allowed." The columns still exist and those facets still work, they just lose the secondary DB index (slower filtering instead of a broken facet). This is the same key-budget ceiling the original `BaseFieldsSearchIndexSubscriber` comment already called out — restoring the intended full field set is what exposes it on sites whose total indexed-field count is already near the limit. Flagging for awareness/triage rather than blocking on it here.

## Test plan

- [x] Added `modules/mukurtu_search/tests/src/Kernel/StaticTaxonomyNameFieldRestoreTest.php` — covers both `mukurtu_search_rebuild_index()` preserving the fields and `mukurtu_search_update_40005()` restoring them, plus idempotency. Passing (3 tests, 328 assertions).
- [x] Reproduced the original error on a real DDEV site, ran `drush updb`, confirmed the "No available query types" message is gone from the live page.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (`mukurtu_search_update_40005`, confirmed no conflicting hook numbers)
- [ ] I have run an accessibility check, and resolved any issues. (N/A — backend-only change, no markup/UI)
- [ ] I have recompiled SCSS if required. (N/A — no CSS changes)
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges. (N/A — based on `main`)
- [ ] I have verified that all expected checks pass. (pending CI on this PR)
- [ ] If I added a content entity bundle... (N/A — no new bundle/view)

### DIFF
--- a/modules/mukurtu_search/mukurtu_search.install
+++ b/modules/mukurtu_search/mukurtu_search.install
@@ -59,3 +59,39 @@ function mukurtu_search_update_40004(): void {
   }
   $config->save();
 }
+
+/**
+ * Restore string (facet) taxonomy name fields dropped by update_40001().
+ *
+ * mukurtu_search_update_40001() called mukurtu_search_rebuild_index(),
+ * which wiped every field on mukurtu_browse_auto_index and only restored
+ * the fulltext "__name__text" variants for taxonomy reference fields, not
+ * the plain string "__name" variants that facets (e.g. Category, Format,
+ * Language) use as their field_identifier. This left those facets unable
+ * to resolve a query type. Re-add whichever of the fields are missing.
+ */
+function mukurtu_search_update_40005(): void {
+  $index = \Drupal\search_api\Entity\Index::load('mukurtu_browse_auto_index');
+  if (!$index) {
+    return;
+  }
+
+  $missing = FALSE;
+  foreach (mukurtu_search_static_taxonomy_name_fields() as $field_id => $definition) {
+    if ($index->getField($field_id)) {
+      continue;
+    }
+    $missing = TRUE;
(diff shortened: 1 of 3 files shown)


christa-kavanagh-wsu (approved):
That did it!

PR #2049, opened 2026-08-27
Wire the language_with_fallback Search API field (Phase 4 infra, #1260)

michael-wynne-wsu:
## Summary

First PR of Phase 4 (site-wide language fallback + a "not yet translated" indicator) of the #1260 multilingual roadmap. Scoped and reconciled against the full 40-view inventory from PR #2044 - see that PR's `EXEMPTIONS` list for the views still needing migration, which is separate, larger, follow-up work not attempted here.

- Adds `mukurtu_multilingual_update_40009()`: wires a `language_with_fallback` field, backed by the `LanguageWithFallback` Search API processor, onto all 6 profile-owned Search API indexes. That processor is already enabled (locked/hidden) on every index - this just gives it somewhere to write its value so views/facets can actually filter/facet on it later (`docs/content-language-policy.md`).
- Follows the established `Index::load()` → mutate → `save()` → `reindex()` idiom used repeatedly elsewhere in this codebase - closest precedent is `mukurtu_collection_update_40002()` (`modules/mukurtu_collection/mukurtu_collection.install:34-100`), minus the processor-creation step since `language_with_fallback` is already enabled everywhere (verified against each index's shipped config before writing this).
- Idempotent (guarded by a `getField()` check) and safely skips any of the 6 index IDs that doesn't exist on a given site - e.g. `mukurtu_solr`'s two indexes, since that module isn't installed by default.
- **Infra only - no visible behavior change.** No view's filters are touched here; that's the larger follow-up work.
- **Requires a full reindex on upgrade**, not just an instant config write - worth calling out in release notes, especially for a large Solr-backed install where reindexing takes real time.

## Test plan

- [x] Added `SearchApiLanguageFallbackFieldTest` (Kernel), modeled on the existing `BrowseCollapseIndexRestoreTest` pattern for Search API index update hooks: asserts the field is added with the correct type/property_path/no-datasource shape (verified against `vendor/drupal/search_api/src/Plugin/search_api/processor/LanguageWithFallback.php`'s own property definition before writing the assertions), asserts a missing index is skipped without error, asserts running the hook twice is a no-op.
- [x] `php -l` on both files - no syntax errors.
- [ ] Could not run the real Kernel test locally - no `web/core` scaffold in this checkout (same limitation as PRs #1950/#1970/#2044). CI is the real verification.

## Pre-merge checklist:

[...]

PR #2073, opened 2026-08-28
Ship the language_with_fallback field in config/install too (#1260)

michael-wynne-wsu:
## Summary

Found while scoping the next batch of Phase 4 view migrations - `mukurtu_multilingual_update_40009()` (#2049) only wires the `language_with_fallback` Search API field for sites **upgrading** from an older schema version. Update hooks never run on a fresh install, and no `hook_install()` in this profile called that logic either (checked `mukurtu_multilingual_install()` directly - it doesn't). A brand-new Mukurtu site therefore never got the field on any of the 6 indexes at all.

This is more than a coverage gap - it's a real functional issue: the 14 remaining Phase 4 views that will filter on this field would ship broken on any fresh install once added, since they'd reference a field that doesn't exist in the index.

**Fix**: ship the field directly in each owning module's `config/install/search_api.index.*.yml` (covers fresh installs, matching how other field/config additions in this codebase already handle both cases), leaving #2049's existing update hook in place unchanged for sites converging from an older version.

## Test plan

- [x] Added `SearchApiFallbackFieldShippedConfigTest` (Unit, no bootstrap): confirms all 6 shipped index configs carry the field with the exact shape the update hook also creates (`type: string`, correct `property_path`, no `datasource_id` since it's a processor-derived index-wide property), and that the `language_with_fallback` processor is actually enabled in each. This is a fresh-install-path check that didn't exist before - the only prior test (`SearchApiLanguageFallbackFieldTest` from #2049) only ever exercised the update-hook path.
- [x] Manually re-verified the same assertions the new test makes directly against the 6 edited files - all pass.
- [x] `php -l` - no syntax errors.
- [ ] Could not run the real Unit test locally - no `web/core` scaffold in this checkout, the same limitation hit throughout this work. CI is the real verification.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (N/A - this ships the field in `config/install` for fresh installs; #2049's existing update hook already handles upgrading sites, unchanged here.)
- [x] I have run an accessibility check, and resolved any issues. (N/A - no rendered markup/UI change.)
- [x] I have recompiled SCSS if required. (N/A.)
- [x] I have updated or created CI/CD tests, if required. (New Unit test added.)
- [x] I have updated from `main` and resolved any merge conflicts. (Branched directly off current `main`.)
- [x] If I added a content entity bundle... (N/A.)

### DIFF
--- a/modules/mukurtu_browse/config/install/search_api.index.mukurtu_default_content_index.yml
+++ b/modules/mukurtu_browse/config/install/search_api.index.mukurtu_default_content_index.yml
@@ -16,6 +16,10 @@ name: 'Mukurtu Default content index'
 description: 'Default Mukurtu content index'
 read_only: false
 field_settings:
+  language_with_fallback:
+    label: 'Language (with fallback)'
+    property_path: language_with_fallback
+    type: string
   changed:
     label: Changed
     datasource_id: 'entity:node'
--- a/modules/mukurtu_collection/config/install/search_api.index.mukurtu_collection_index.yml
+++ b/modules/mukurtu_collection/config/install/search_api.index.mukurtu_collection_index.yml
@@ -10,6 +10,10 @@ name: 'Mukurtu Collection Index'
 description: 'Index for Mukurtu Collections.'
 read_only: false
 field_settings:
+  language_with_fallback:
+    label: 'Language (with fallback)'
+    property_path: language_with_fallback
+    type: string
   changed:
     label: Changed
     datasource_id: 'entity:node'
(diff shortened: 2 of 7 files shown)


michael-wynne-wsu:

Link:
- https://pr2073-fjh0hymna7dhqtcij6tjjmkkp8cffo3l.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a91a5a1f8743f32ef2745ca

Issue #2078, opened 2026-08-28
Discovery spike: does legacy D7 (v3) migration data need translation support?


michael-wynne-wsu:
Phase 5 of the #1260 multilingual roadmap (export/import translation roundtrip)
has an open question blocking any migration-specific work: did real Mukurtu v3
(Drupal 7) sites store multilingual content via Drupal's entity-level translation
system, or via separate nodes per language (no entity translation relationship
between them)?

This matters because the two cases need entirely different migration handling:

- Entity-level translation on D7 -> migrate should preserve the translation
  relationship, needs source-side inspection of the D7 \`content_translation\`/
  \`entity_translation\` tables (whichever the source sites used).
- Node-per-language on D7 -> there's no relationship to preserve; each node
  migrates independently and none of this applies.

Scope of the spike: survey real/representative v3 site database exports (or
ask site maintainers directly) to determine which pattern was actually used
in practice, and whether both patterns coexist across different sites. Output
should be a short finding, not code -- it gates whether Phase 5's migration
track needs any translation-specific logic at all.

michael-wynne-wsu:
In reality, we are not expecting any Mukurtu 3 translations to migrate, since it was poorly supported.

PR #2080, opened 2026-08-28
Make CSV export translation-aware (#1260 Phase 5, export track)

michael-wynne-wsu:
## Summary
- Part of #1260 Phase 5 (export/import translation roundtrip) - the **export track only**. Import stays a separate follow-up (its riskiest sub-item, the legacy D7 migration discovery spike, is already tracked as #2078).
- Exporting content always emitted the default/original-language field values, regardless of which translation a user actually wanted. The langcode was already available in two places and silently dropped in both:
  - The manage-content view's VBO bulk-export action already carries each row's langcode.
  - Saved Export Lists had no language concept at all - items were always added as bare entity IDs.
- New `ExportItemIdentity` helper defines one item-key format (bare ID, or a composite `"$id:$langcode"`) used by every exporter source, so `CSV.php`'s batch logic only ever needs to understand one shape.
- `CSV::batchExport()` resolves a composite key to the requested translation, falling back to the entity's own language if that translation doesn't exist - never a hard failure.
- `AdHocExportStartController::startBulk()` now encodes the VBO row's langcode instead of discarding it.
- `ExportList` gains an additive `item_languages` base field (+ update hook) so saved lists can also carry a per-item language, with a language picker (shown only when an entity actually has more than one translation) in `ExportListAddItemForm`/`ExportListAddNodeForm`. `items`' own shape is deliberately untouched - `ExportListRemoveItemsForm`/`ExportListListBuilder` match/remove by bare entity ID, so encoding language into those keys would have broken removal. An item with no `item_languages` entry exports exactly as before this field existed.
- `CsvEntityFieldExportEventSubscriber` routes referenced-entity lookups (taxonomy term names, chained entity exports like shallow/full media or paragraph references) through `mukurtu_core`'s `EntityTranslationResolver`, using the row's own langcode - so a translated row's referenced content also renders in that row's language, not the referenced entity's own default language. Deliberately does *not* use the resolver's active-content-language fallback when the row itself is in its original language, to avoid a regression where referenced entities would silently pick up the current request's language instead of their own default.
- Scope boundary: one langcode per (entity_type, id) per saved list - re-adding the same item with a different language choice overwrites the previous one rather than creating two entries.
- See `docs/export-translation.md` for the full design writeup.

## Test plan
- [x] New Unit test (`ExportItemIdentityTest`) for the key encode/decode helper
- [x] New Kernel test (`ExportListItemLanguageTest`) covering the new field's set/get round-trip and `ExportListSource`'s key building, including that untouched items still export with a bare key (zero behavior change for existing lists)
- [x] New Kernel test suite (`CsvExportTranslationTest`) covering `EntityFieldExportEvent::getLangcode()`, a translated row's own fields, a referenced taxonomy term's name resolving to the row's language (and to the term's own default language for a non-translated row), a referenced entity being queued under a composite key, and an end-to-end `CSV::batchExport()` test proving a composite-key item exports the requested translation's field values
- [x] Ran the WCAG accessibility check skill on the new language-picker form element (standard Drupal Form API `select`, confirmed compliant, no custom markup)
- [x] `php -l` on every changed/new PHP file

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (`mukurtu_export_update_40026()` installs the new `item_languages` base field for existing sites; fresh installs get it automatically via `ExportList::baseFieldDefinitions()`.)
- [x] I have recompiled SCSS if required. (N/A - no SCSS changes.)
- [x] I have updated from `main` and resolved any merge conflicts. (Branched directly from `main`.)
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges. (N/A - based on `main`.)
- [x] If I added a content entity bundle, I added its `language.content_settings` and base field overrides to `mukurtu_multilingual`. If I added or changed a view, I applied the content language policy in that doc. (N/A - no new bundle/view; this is export-pipeline logic, documented separately in `docs/export-translation.md`.)

### DIFF
--- a/docs/export-translation.md
+++ b/docs/export-translation.md
@@ -1,0 +1,31 @@
+# Export Translation Support
+
+This documents how `mukurtu_export` handles translated content (#1260 Phase 5, export track). It exists because export previously always emitted an entity's default/original-language field values, regardless of which translation a user actually wanted.
+
+## Item identity
+
+Every exporter source (`AdHocExporterSource`, `ExportListSource`) hands `CSV.php` an `entity_type_id => [key => key]` map, same as before. A key is now either:
+
+- A bare entity ID — export that entity's original language. Unchanged from before this feature existed.
+- A composite `"$id:$langcode"` key, built via `ExportItemIdentity::encode()`/parsed via `::decode()` — export the specific translation. `CSV::batchExport()` swaps in `$entity->getTranslation($langcode)` when it exists, and falls back to the entity's own language otherwise (never a hard failure).
+
+Only one format is defined, in one place (`ExportItemIdentity`), so `CSV.php`'s batch logic never needs to know which source produced a given key.
+
+## Two ways a composite key gets produced
+
+- **VBO bulk export** (`AdHocExportStartController::startBulk()`): the manage-content view's selection already carries each row's langcode; it's encoded directly since this path is ephemeral (tempstore-backed, nothing persisted).
+- **Saved Export Lists**: `ExportList` gained an additive `item_languages` base field (`"$entity_type:$id" => $langcode`), separate from the existing `items` field. `items`' shape is intentionally untouched — `ExportListRemoveItemsForm`/`ExportListListBuilder` match/remove by bare entity ID, so encoding language into `items`' own keys would have broken removal. An item with no `item_languages` entry exports in its original language, exactly as before this field existed — no update hook was needed to migrate existing data, only the standard field-installation update hook for the schema itself.
+  - **Scope boundary**: one langcode per (entity_type, id) per list. Adding the same item again with a different language choice overwrites the previous one rather than creating a second entry.
+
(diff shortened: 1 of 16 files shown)


michael-wynne-wsu:

Link:
- https://pr2080-gaaajzxifb3rawzm29y79k2shfgidylx.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a91d0eff8743f32ef2bf655

michael-wynne-wsu:

Link:
- https://pr2080-gaaajzxifb3rawzm29y79k2shfgidylx.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a91d0eff8743f32ef2bf655

michael-wynne-wsu:

Link:
- https://pr2080-gaaajzxifb3rawzm29y79k2shfgidylx.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a91d0eff8743f32ef2bf655

PR #2081, opened 2026-08-28
Make CSV import translation-aware (#1260 Phase 5, import track)

michael-wynne-wsu:
## Summary
- The other half of #1260 Phase 5 (export/import translation roundtrip) - the export track shipped in #2080. Legacy D7 migration (`mukurtu_migrate`) needs no work here (#2078, closed - no D7 multilingual content is expected to be migrated).
- Import always created or overwrote the default-language entity, regardless of a mapped langcode column. Drupal core's own `entity:$entity_type_id` migrate destination already supports targeting a translation (`destination.translations`) - it was just never turned on here.
- `MukurtuImportStrategy::toDefinition()` now sets `destination.translations`, strictly opt-in: gated on both a column being mapped to the langcode field, and the target bundle actually having content translation enabled. Neither condition met → byte-identical to before this shipped.
- `getOverwriteProperties()` excludes non-translatable fields (`field_cultural_protocols` - sharing settings/protocol assignment, entity-wide not per-language - is the concrete example) from the writable-fields list whenever a definition is a translation import. Core's own `EntityContentBase::updateEntity()` writes every `overwrite_properties` field through whichever translation a row resolves to, with **no guard** against clobbering a non-translatable field's shared value across every other translation - this closes that gap.
  - Traced against core's `Entity::getEntity()` to confirm the precise scope: `overwrite_properties` only applies to the **update** path (an existing entity matched by ID/UUID/identifier column). Brand-new content creation goes through `storage->create()` with every mapped field, completely unaffected - so a translation-enabled strategy can still create new content with protocols set. The real, narrower limitation: it can no longer **update** a non-translatable field on already-existing content through that strategy. See `docs/import-translation.md` for the full writeup, including why a per-row custom destination plugin (correct but costlier, ~6 entity types' worth of plugin subclasses) was considered and declined in favor of this simpler static guard.
- `mukurtu_import.info.yml` gains an explicit `drupal:content_translation` dependency (the module used the service without declaring it before).
- `ImportFieldDescriptionListForm`'s langcode field help text (previously `taxonomy_term`-only, and already claimed langcode was "used for translation" before that was actually true) now applies to every entity type and explains the opt-in gating and the field-clobber limitation. Wording approved via AskUserQuestion before implementation.

## Test plan
- [x] New `ImportTranslationTest`: a translation import adds an 'es' title translation without touching the base (original-language) entity's title or its `field_cultural_protocols`
- [x] Two more tests proving the opt-in gate directly against `toDefinition()`'s output: no `translations` key when the bundle isn't translation-enabled, and no `translations` key when no langcode column is mapped
- [x] `php -l` on every changed/new PHP file
- [x] Ran `scripts/lint/multilingual-guardrails.sh` - clean
- [x] Confirmed no update hook needed - `toDefinition()`'s output isn't itself stored config, only `MukurtuImportStrategy`'s own fields are (id/uid/label/description/target_entity_type_id/target_bundle/mapping/configuration), none of which changed shape

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required — see above.)
- [x] I have run an accessibility check, and resolved any issues. (N/A — plain help text in an existing table, no new markup or interactive elements.)
- [x] I have recompiled SCSS if required. (N/A — no SCSS changes.)
- [x] I have updated from `main` and resolved any merge conflicts. (Branched directly from `main`, includes the merged export track.)
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges. (N/A — based on `main`.)
- [x] If I added a content entity bundle, I added its `language.content_settings` and base field overrides to `mukurtu_multilingual`. If I added or changed a view, I applied the content language policy in that doc. (N/A — no new bundle/view; this is import-pipeline logic, documented separately in `docs/import-translation.md`.)

### DIFF
--- a/docs/import-translation.md
+++ b/docs/import-translation.md
@@ -1,0 +1,35 @@
+# Import Translation Support
+
+This documents how `mukurtu_import` (CSV/spreadsheet import) handles targeting a *translation* of existing content, rather than always creating or overwriting the default-language entity (#1260 Phase 5, import track). It's the counterpart to `docs/export-translation.md`'s export track.
+
+Only the CSV import path is covered. `mukurtu_migrate` (legacy Drupal 7 migration) needs nothing here — the profile has no expectation of migrating D7 multilingual content, and will revisit only if that's actually requested (see #2078).
+
+## Turning it on: `destination.translations`
+
+Drupal core's own `entity:$entity_type_id` migrate destination already knows how to target a translation — set `translations: true` on the destination and, for any row whose mapped langcode differs from an existing entity's own language, it calls `addTranslation()`/`getTranslation()` before writing that row's fields, instead of overwriting the base entity. Mukurtu never turned this on; `MukurtuImportStrategy::toDefinition()` now does, **strictly opt-in**, gated on both:
+
+- The strategy's mapping includes a column mapped to the entity type's langcode field (already a normal, listed mapping target — `ImportFormTrait::buildTargetOptions()` labels it "... (langcode)" to disambiguate it from any other field that happens to also be called "Language").
+- The target bundle actually has content translation enabled (`content_translation.manager`'s `isEnabled($entity_type_id, $bundle)`).
+
+Neither condition met → `translations` is never set, and behavior is byte-identical to before this existed. See `MukurtuImportStrategy::isTranslationImport()`.
+
+## The non-translatable-field guard
+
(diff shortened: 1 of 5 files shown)


michael-wynne-wsu:

Link:
- https://pr2081-lyf5rhip2rdptcxsdhg4c7qruzkaa6f5.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6a91e8b4f8743f32ef2e8b90

Issue #2125, opened 2026-09-08
phpunit.xml testsuites exclude profile-root tests, so they never run in CI


michael-wynne-wsu:
Found while implementing Phase 1 of #2123.

[`phpunit.xml`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/phpunit.xml) declares its three testsuites like this:

```xml
<testsuites>
  <testsuite name="kernel">
    <directory>web/profiles/mukurtu/modules/**/tests/src/Kernel</directory>
  </testsuite>
  <testsuite name="unit">
    <directory>web/profiles/mukurtu/modules/**/tests/src/Unit</directory>
  </testsuite>
  <testsuite name="functional">
    <directory>web/profiles/mukurtu/modules/**/tests/src/Functional</directory>
  </testsuite>
</testsuites>
```

Every path is rooted at `modules/**`, so a test placed at the profile root (`web/profiles/mukurtu/tests/src/Kernel`) is never collected. CI runs `--testsuite kernel` and `--testsuite unit` ([build-and-test.yml](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/.github/workflows/build-and-test.yml)), so such a test would sit in the repo silently passing without ever executing.

The existing `tests/` directory holds only the Playwright suite, so nothing is currently affected. It becomes a trap as soon as anyone writes a PHPUnit test for profile-root code.

## Why it matters

There is real profile-root code to test. `mukurtu.install` is about 1,000 lines including `mukurtu_install()`, and `mukurtu.profile` now carries `mukurtu_form_install_configure_form_alter()` and `mukurtu_install_configure_form_submit()` as of #2123.

In that PR I wanted a kernel test for the new profile hook and had to put it in `modules/mukurtu_core/tests/src/Kernel/FeedbackFormRecipientInstallTest.php` so CI would run it, with a docblock explaining why it lives in the wrong place. That works, and it matches the `docs/update-hooks.md` convention of falling back to `mukurtu_core` when there is no clear module, but it is misleading: the test requires `mukurtu.profile` and exercises nothing in `mukurtu_core`.

## Suggested fix

Add a second `<directory>` to each testsuite:

```xml
<testsuite name="kernel">
  <directory>web/profiles/mukurtu/modules/**/tests/src/Kernel</directory>
  <directory>web/profiles/mukurtu/tests/src/Kernel</directory>
</testsuite>
```

[...]

Issue #2124, opened 2026-09-08
Dead administrator permission list in mukurtu_install() (is_admin roles discard grants)


michael-wynne-wsu:
Found while implementing Phase 1 of #2123.

`mukurtu_install()` builds a 186-line permission list for the `administrator` role at [mukurtu.install:559-744](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.install#L559-L744) and applies it through the loop at [mukurtu.install:753-759](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.install#L753-L759):

```php
foreach ($roles as $role_name => $permissions) {
  $role = Role::load($role_name);
  foreach ($permissions as $permission) {
    $role->grantPermission($permission);
  }
  $role->save();
}
```

None of those 186 permissions are ever stored, because [`config/install/user.role.administrator.yml`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/config/install/user.role.administrator.yml) sets `is_admin: true` and core's `Role::grantPermission()` returns early for admin roles:

```php
public function grantPermission($permission) {
  if ($this->isAdmin()) {
    return $this;
  }
  ...
}
```

`Role::getPermissions()` and `Role::hasPermission()` short-circuit the same way, so the role reports every permission as granted while storing none.

Confirmed on a fresh install of the profile:

```
administrator          exists: true  stored permission count: 0
mukurtu_manager        exists: true  stored permission count: 221
```

and in raw config: `user.role.administrator` has `is_admin: true` with `permissions: []`.

## Why it matters

Nothing is broken today. `is_admin: true` already grants everything, including permissions added by modules in future releases, which is standard Drupal behaviour. The problems are maintenance ones:

- The list reads as the authoritative definition of what an administrator can do, so it invites edits that have no effect. Phase 1 of #2123 removed `'access devel information'` from it, and that edit was a no-op.
- It is 186 lines of the profile's install path that can never be exercised, and it drifts further from reality with every release.

## Suggested fix

Delete the `$roles['administrator']` block and leave a short comment on the `administrator` entry in `config/install/user.role.administrator.yml` noting that `is_admin: true` is what grants the permissions, so no explicit list is needed.

[...]

PR #2126, opened 2026-09-08
4.0.0 release readiness

### LINKED ISSUE #2123: 4.0.0 public release readiness and 4.0.1 update-hook reset
`4.0.0-rc` is tagged. This issue covers the two releases that take us from a beta shared with known testers to a public product, and asks the team to review the approach before work starts.

**Phase 1 (4.0.0)** carries the full update-hook chain so existing testers can reach the release state. **Phase 2 (4.0.1)** removes every update hook and declares a clean baseline. New users install 4.0.1 directly; testers go 4.0.0 then 4.0.1 as a no-op. Everyone converges and we carry on fresh.

---

### The distribution still carries beta assumptions

The most serious one: **Devel and Devel Generate are enabled on every fresh Mukurtu site.** `devel:devel` and `mukurtu_dev:mukurtu_dev` are in the profile's `dependencies:` block ([mukurtu.info.yml#L73](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.info.yml#L73), [#L136](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.info.yml#L136)), which Drupal installs at install time. On top of that, `access devel information` is granted to `administrator` ([mukurtu.install#L562](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.install#L562)) and devel permissions to `mukurtu_manager` ([user.role.mukurtu_manager.yml#L31](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/config/install/user.role.mukurtu_manager.yml#L31)). Devel exposes container and state dumps and arbitrary PHP execution routes.

There is a related correctness bug hiding behind this. `drupal/devel` is only in `require-dev` ([composer.json#L92](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/composer.json#L92)), and a package's `require-dev` is never installed for consumers. So the profile's hard dependency on devel means a `composer install --no-dev` production deploy cannot install the profile at all. Removing devel from `dependencies:` fixes both problems at once.

Also outstanding: no LICENSE file anywhere, `version: 4.x-dev` in 27 `.info.yml` files, `"license": "ISC"` in the root `package.json` (conflicts with GPL), and a README that still says "Only use for testing and feedback purposes."

### 518 update hooks that almost nobody needs

37 beta tags have left 518 `hook_update_N()` implementations across 28 `.install` files, roughly 19,600 lines. Nearly all of it is catch-up work for sites that started on a beta. `docs/update-hooks.md` already commits to this cleanup; we are doing it one release earlier than that note anticipated.

---

## Verified mechanics

Checked directly against Drupal core 11.4.6, because the whole plan rests on these.

1. **Fresh installs never run update hooks.** `ModuleInstaller::install()` (`core/lib/Drupal/Core/Extension/ModuleInstaller.php:407-425`) seeds the schema version from `max(available updates)` and never invokes them. **Stripping hooks cannot break a fresh install.** Any `config/install` drift is a pre-existing 4.0.0 problem, not something the strip introduces.
2. **`hook_update_last_removed()` seeds fresh installs correctly.** Same function, line 422: a module with zero hooks but a `_update_last_removed()` is installed *at that version*, so a fresh 4.0.1 site is correctly marked as already past everything 4.0.0 shipped.
3. **`hook_update_last_removed()` does NOT block an out-of-date site.** This is the one that changes the plan. `update_get_update_list()` (`core/includes/update.inc:308`) only lists updates greater than the installed version. Core's only other use of `_update_last_removed` is the missing-schema repair at line 109. A site on beta30 pointed straight at 4.0.1 would silently report "no pending updates" and then run with stale config. **We have to supply our own guard.**
4. **`hook_update_requirements()` is that guard.** `update_check_requirements()` (`core/includes/update.inc:49-64`) invokes it, and Drush calls that from `UpdateDBCommands.php:552`. A `REQUIREMENT_ERROR` blocks both `update.php` and `drush updb`.
5. **Nothing but tests calls an update hook.** No `hook_install()` and no update hook invokes another update hook. The only callers outside their own definitions are 65 kernel test files.
6. **No module is installed solely by an update hook.** Everything an update hook installs is already in `mukurtu.info.yml`'s `install:` or `dependencies:`.
7. **Base fields added by update hooks are all declared in code** (`ExportList.php:88`, `Protocol.php:618`, `mukurtu_notifications_entity_base_field_info()`), so fresh installs get them from `baseFieldDefinitions()`.

---

## Phase 1: 4.0.0

Branch off `main`. Deliberately small: **only changes that need an update hook go here.** Anything with no config or data impact (docs, LICENSE, composer constraints, version strings) has no reason to be in 4.0.0 and rides in 4.0.1 instead. That keeps the tag testers must pass through as low-risk as possible.

Two new hooks in `mukurtu_core.install`, numbered from the current max of 40122.

### Devel, fresh installs

- [`mukurtu.info.yml`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.info.yml): delete `- 'devel:devel'` and `- 'mukurtu_dev:mukurtu_dev'` from `dependencies:`.
- [`modules/mukurtu_dev/mukurtu_dev.info.yml`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/modules/mukurtu_dev/mukurtu_dev.info.yml): add `hidden: true`. The module stays in the repo; developers enable it themselves.
- [`mukurtu.install#L562`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/mukurtu.install#L562): drop `'access devel information'` from the `administrator` grant.
- [`config/install/user.role.mukurtu_manager.yml`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/config/install/user.role.mukurtu_manager.yml): remove the `devel` dependency entry and the devel permissions.

### Devel, existing sites

`mukurtu_core_update_40123()` revokes the devel permissions from `administrator` and `mukurtu_manager`, and logs a message pointing at `/admin/modules/uninstall`.

Proposed: leave the modules themselves installed. Testers may be actively using them, and uninstalling a module out from under a live site is more disruptive than revoking a permission they can re-grant. **Open question for reviewers:** do we want to force-uninstall `devel`/`devel_generate`/`mukurtu_dev` on tester sites instead?

### Contact form recipient

[`config/install/contact.form.feedback.yml#L7`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/config/install/contact.form.feedback.yml#L7) ships `recipients: [admin@example.com]`. Core does not substitute this at install time (the contact module has no `hook_form_install_configure_form_alter()`), so every Mukurtu site's website-feedback form mails a dead address.

- Fixup in `mukurtu_install()` setting recipients to `\Drupal::config('system.site')->get('mail')`.
- `mukurtu_core_update_40124()` does the same for existing sites, guarded to act only when recipients is still exactly `['admin@example.com']` so it cannot clobber a site that configured its own.

### Cut

Bump `VERSION.md` to `4.0.0`, tag, publish release notes. Create a `4.0.0` branch pointer at the tag so an emergency patch has somewhere to live.

---

## Phase 2: 4.0.1

Branch off the `4.0.0` tag.

### 1. Strip the hooks

Delete from all 28 `.install` files:

- every `<module>_update_N()` function;
- the 5 `hook_update_dependencies()` implementations, which reference hooks that are going away (`mukurtu_core.install:824`, `mukurtu_import.install:556`, `mukurtu_export.install:22`, `mukurtu_protocol.install:215`, `mukurtu_taxonomy.install:17`);
- the 3 `hook_update_requirements_alter()` implementations (`mukurtu_core.install:3565`, `mukurtu_design.install:214`, `mukurtu_notifications.install:720`), which exist only to let the now-deleted `features` / `config_pages` / `color_field` / `message_digest` cleanup hooks run;
- any helper left with no caller.

**Keep** every `hook_install()`, `hook_uninstall()`, `hook_schema()`, and any helper still called from the install path. Note that `mukurtu_submissions_review_permissions()` and `mukurtu_search_rebuild_index()` are deliberately shared between install and update paths, so they stay.

### 2. Declare the baseline

Add `<module>_update_last_removed(): int` to each `.install` file that had hooks, returning that module's highest N **as tagged in 4.0.0**. Values must be computed from the 4.0.0 tag, not from today's `main`, since `mukurtu_core` moves to 40124 once the Phase 1 hooks land.

| Module | last_removed | Module | last_removed |
| --- | --- | --- | --- |
| `mukurtu_core` | 40122 * | `mukurtu_landing_page` | 40006 |
| `mukurtu_notifications` | 40057 | `mukurtu_solr` | 40005 |
| `mukurtu_import` | 40046 | `mukurtu_design` | 40004 |
| `mukurtu_protocol` | 40045 | `mukurtu_footer` | 40002 |
| `mukurtu_dictionary` | 40044 | `mukurtu_migrate` | 40002 |
| `mukurtu` (profile) | 40029 | `mukurtu_bot_protection` | 40001 |
| `mukurtu_export` | 40027 | `mukurtu_protocol_sync` | 40001 |
| `mukurtu_collection` | 40027 | `mukurtu_gin_custom` | 10000 |
| `mukurtu_media` | 40023 | `mukurtu_digital_heritage` | 40009 |
| `mukurtu_browse` | 40021 | `mukurtu_multilingual` | 40009 |
| `mukurtu_taxonomy` | 40021 | `mukurtu_person` | 40009 |
| `mukurtu_submissions` | 40014 | `mukurtu_place` | 40009 |
| `mukurtu_multipage_items` | 40011 | `mukurtu_workflows` | 40008 |
| `mukurtu_search` | 40007 | `mukurtu_local_contexts` | 40007 |

\* becomes 40124 after the Phase 1 hooks.

Modules that never had hooks (`mukurtu_community_records`, `mukurtu_content_warnings`, `mukurtu_drafts`, `mukurtu_setup`, `mukurtu_site_settings`, `mukurtu_dev`, `original_date`) need nothing.

### 3. Add the too-old guard

Because core does not enforce `last_removed` (mechanic 3 above), add `mukurtu_core_update_requirements()` to `mukurtu_core.install`. It loops over installed `mukurtu` / `mukurtu_*` modules, calls `loadInclude($module, 'install')`, and where `<module>_update_last_removed()` exists compares it against `\Drupal::service('update.update_hook_registry')->getInstalledVersion($module)`, skipping `SCHEMA_UNINSTALLED`. If any module is behind, it returns a `REQUIREMENT_ERROR` naming them and directing the operator to update to 4.0.0 first. Written generically over `_update_last_removed()` so it needs no maintenance.

Also add a matching runtime `hook_requirements()` so a site that skipped `updb` entirely still surfaces the problem on the status report. The runtime one can be an OOP `#[Hook]` class under `modules/mukurtu_core/src/Hook/`; the update-phase one must stay procedural in the `.install` file.

Exact wording of both messages to be reviewed before it lands.

### 4. Rework the tests

65 kernel test files (about 7,900 lines) invoke update hooks directly and will fail once the hooks are gone. Find them with:

```
grep -rlE '^\s+[a-z_]+_update_[0-9]{4,5}\(' modules/*/tests
```

Delete those, then add a focused set of fresh-install invariant tests that assert the same end states straight from `config/install`, with no hook invocation:

- views language fallback across browse / collection / dictionary / taxonomy / solr, currently spread over about 8 `*ViewsLanguageFallbackTest.php` files and collapsible into one data-provider test;
- the permission sets asserted by the `*PermissionUpdateTest.php` files;
- map edit-form defaults, currently four near-identical `*MapEditFormDefaultsUpdateTest.php` files;
- the `mukurtu_browse_auto_index` 64-key budget from `BrowseAutoIndexKeyBudgetTest.php`, which guards a real MySQL limit.

This is a net win: the assertions move from "the hook produced X" to "a fresh install produces X", which is what we actually care about from here on.

### 5. Add a numbering lint

The documented scheme (3 version digits plus 2 sequential) **has already overflowed**: `mukurtu_core` ran 40100 through 40122, consuming the entire 4.0.1 band. Post-strip hook numbers must exceed each module's `_update_last_removed()`, so the next series starts at **40200** (4.0.2).

To stop this recurring, add `scripts/lint/update-hook-numbering.sh` next to the existing [`multilingual-guardrails.sh`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/scripts/lint/multilingual-guardrails.sh), asserting that every `_update_N` in a file exceeds that module's `_update_last_removed()`. Wire it into [`build-and-test.yml`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/.github/workflows/build-and-test.yml) beside the multilingual gate.

### 6. Release-readiness work

None of this needs an update hook, so it rides in 4.0.1.

- **LICENSE** at repo root (GPL-2.0-or-later), a `"license"` key in `composer.json`, and fix `package.json:11` which declares `ISC`. Add Tagify's MIT license text under `libraries/tagify/`.
- **Version strings**: `VERSION.md`, the 27 `.info.yml` files off `4.x-dev`, and add a `version:` key to `themes/mukurtu_v4/mukurtu_v4.info.yml`, which has none. Worth a small `scripts/set-version.sh` given this is now a 28-file edit every release.
- **README**: drop the "testing and feedback purposes only" banner (line 8) and the beta-hopping upgrade instructions (lines 106-130). Replace with a stable-release upgrade section stating plainly that **any site below 4.0.0 must update to 4.0.0 before 4.0.1**. Change `composer create-project mukurtu/mukurtu-template:dev-main` (line 43) to a tagged constraint.
- **CHANGELOG.md**, **SECURITY.md**, **CONTRIBUTING.md**: none exist.
- **`.gitattributes`** with `export-ignore` for `tests/`, `modules/*/tests/`, `.github/`, `.tugboat/`, `.vscode/`, `phpunit*.xml`, `scripts/`, `docs/`. **Do not exclude `patches/`**: composer.json references those files by path and the README documents the failure mode when they go missing.
- **`.gitignore`**: add `.claude/`, currently only excluded via a machine-local `.git/info/exclude`.
- **composer.json**: `minimum-stability` to `stable`; pin `drupal/og` off `^1.x-dev` (it is the backbone of protocol access and currently tracks a moving branch); add a `php` constraint; replace the 15 wildcard `*` constraints with `^N` ranges; drop the four now-unnecessary `@beta` / `@RC` flags (`flag`, `message_ui`, `term_merge`, `phpgpx` all resolve stable already); delete the dead `drupal/views_combine` patch entry and the orphaned `patches/pathauto-source-module-missing-php8-attribute.patch`.
- **`fclose()` fatal**: [`mukurtu_core.module#L1184`](https://github.com/MukurtuCMS/Mukurtu-CMS/blob/main/modules/mukurtu_core/mukurtu_core.module#L1184) calls `fclose($version_file)` outside both guards, so a missing or unreadable `VERSION.md` throws a `TypeError` and fatals the dashboard menu build. Move it inside the inner `if`. Especially relevant given the `.gitattributes` work could exclude `VERSION.md` by accident.
- **`hook_requirements()`** for install and runtime: check `file_private_path` is set (access control depends on it) and warn on missing `pdftotext` / `ffmpeg`. There is currently no `hook_requirements()` anywhere in the codebase.
- **`docs/update-hooks.md`**: rewrite the to-do to describe what actually happened, and document the baseline plus the 40200 rule.
- **`.github/dependabot.yml`**: none exists, against 183 locked runtime packages with no automated advisory scanning.

### 7. Cut

Bump `VERSION.md`, tag `4.0.1`. Release notes lead with the must-pass-through-4.0.0 requirement.

---

## Prerequisite outside this repo

`mukurtu/mukurtu-template` is a **separate repository** and is what `composer create-project` actually installs. Filed as **MukurtuCMS/mukurtu-template#4**, which covers tagging it, constraining `mukurtu/mukurtu` to `^4.0`, and setting its own `minimum-stability`.

Three things found while investigating it:

- It is **completely untagged** and ships **no `composer.lock`**, so every new site resolves dependencies fresh against `minimum-stability: dev` and `"mukurtu/mukurtu": "*"`. That is why both READMEs and this repo's CI workflow install `dev-main`.
- `minimum-stability` and constraint stability flags like `@beta` are read **only** by composer's `RootPackageLoader` (`extractStabilityFlags()` is defined nowhere else in composer). They apply to the root package alone, so the profile's settings do **not** carry over to installed sites. Tightening has to happen in the template.
- `drupal/term_reference_change` has never had a stable release and is unnamed in any composer.json, arriving transitively via `drupal/term_merge`'s `*` constraint. It alone makes `minimum-stability: stable` fail resolution until it is declared with a flag.

## Deferred to 4.0.2

Kept out so 4.0.1 stays a clean no-op for testers:

- `rest` and `restui` enabled by default with no shipped endpoints
- the `drupal/facets >=3.0 <3.0.4` and `drupal/pathauto ~1.14.0` upper-bound pins, which block future security releases on those projects
- the `~11.4.0` core pin versus the profile's `core_version_requirement: '^10 || ^11'` claim
- the placeholder root `package.json` (`"name": "gulp"`, description is a pasted CI badge)

---

## Verification

CI runs only the `kernel` and `unit` suites. There is no functional or install test, so none of the release paths are covered automatically. All four scenarios need a real DDEV site.

1. **Fresh 4.0.1 install.** `drush si mukurtu` on the 4.0.1 branch. Assert the site installs, `devel` / `devel_generate` / `mukurtu_dev` are not enabled, `drush updb --dry-run` reports nothing pending, and each module's schema version equals its `_update_last_removed()`.
2. **The supported path.** Install an old beta (say `4.0.0-beta30`), check out 4.0.0, `drush updb`, then check out 4.0.1 and `drush updb` again. The second run must be a clean no-op. Exercise browse, dictionary, protocols and submissions afterward.
3. **The blocked path.** From the same beta site, jump straight to 4.0.1. `drush updb` must fail with the requirements error and the status report must show it. This is the check that proves the guard works, since core would otherwise silently report "up to date".
4. **Config parity.** `drush config:export` from the scenario 1 and scenario 2 sites, then diff the Mukurtu-owned config. Should be empty. This is the real test of whether `config/install` stayed in sync with the update hooks over 37 betas, and it is worth running **before we tag 4.0.0**: any drift it finds is an existing 4.0.0 bug that the strip would make permanent.
5. **CI** green after the test rework, with both lint scripts passing.

Standard pre-PR workflow applies to each PR: compile SCSS unminified, update from the base branch, run the accessibility check on modified elements, run the PR review, use the PR template.

michael-wynne-wsu:
Phase 1 of #2123. Closes nothing on its own; #2123 stays open to track Phase 2 (the 4.0.1 update-hook reset).

The organising rule for #2123 is that **only changes needing an update hook belong in 4.0.0**, because 4.0.0 is the tag every existing tester has to pass through. Everything with no config or data impact (LICENSE, version strings, README, composer constraints, `.gitattributes`) rides in 4.0.1. That makes this a deliberately small diff: two update hooks and the fresh-install equivalents.

## Devel and Devel Generate were enabled on every Mukurtu site

`devel:devel` and `mukurtu_dev:mukurtu_dev` were in the profile's `dependencies:` block, which Drupal installs at install time. So every site, including production, shipped with Devel and Devel Generate enabled. Devel exposes container and state dumps, user switching, and arbitrary content generation.

This also fixes a genuine correctness bug. `drupal/devel` is only in `require-dev`, and a package's `require-dev` is never installed for consumers, so the profile's hard dependency on devel meant a `composer install --no-dev` production deploy could not install the profile at all. (Confirmed from the other side too: `mukurtu/mukurtu-template` carries `drupal/devel` in *its* `require-dev`, which is why installs have worked so far and why CI has devel available.)

- `mukurtu_dev` is kept in the codebase and marked `hidden: true`, so it stays available to anyone working on Mukurtu itself. It contains no code at all, just an `.info.yml` declaring the two devel dependencies.
**Existing sites are left alone.** An earlier revision of this branch uninstalled the three modules on update. That was deliberately dropped: taking a tool away from a tester mid-upgrade is the site owner's decision, and no removal release is fixed yet. Instead they get a warning on the status report while any of the modules is still installed.

### Why the softer alternatives were rejected

These all look plausible and are dead ends, so recording the findings:

- **Revoking permissions is inert.** `Role::revokePermission()` returns early on `is_admin` roles exactly as `grantPermission()` does, so nothing can be revoked from `administrator`. uid 1 bypasses permission checks entirely. And the only revocable grant is the `administer site configuration` that Mukurtu Manager legitimately needs.
- **Hiding is cosmetic.** All 29 devel routes are reachable by URL regardless of whether menu links or entity tabs are visible.
- **Drupal 8 removed module disabling.** `ModuleInstallerInterface` has only `install()` and `uninstall()`; there is no state in between.

### What the warning says, and why it is specific

9 of devel's 29 routes are gated on core's `administer site configuration` rather than on a devel permission, and Mukurtu Manager holds that permission for unrelated reasons. So Managers can reach `/devel/config/edit/{name}`, `/devel/state/edit/{name}`, `/devel/reinstall` and `/devel/run-cron`. The warning names that exposure rather than being vague.

Implemented as `hook_runtime_requirements()` rather than `hook_requirements()`, which lets it be an OOP `#[Hook]` class per the project convention: `requirements` is on core's static deny list and would throw a `LogicException`, but `runtime_requirements` is not, and `HookCollectorPass` registers such classes as autowired services.

### The stale role dependency had to go in the same commit

`config/install/user.role.mukurtu_manager.yml` listed `devel` under its module dependencies while granting no devel permission at all. That entry could not be left behind: `ConfigInstaller` **silently drops** config whose dependencies are unmet and cascades to anything depending on it (`ConfigInstaller.php:281`). Removing devel from the profile while leaving that entry would have brought every fresh install up with no Mukurtu Manager role and no error message.

## The Website feedback form mailed a dead address

`config/install/contact.form.feedback.yml` shipped Drupal's `admin@example.com` placeholder, and nothing in core substitutes it during installation (the contact module has no `hook_form_install_configure_form_alter()`). Every Mukurtu site's feedback form mailed an address nobody reads.

- `mukurtu_core_update_40123()` replaces it on existing sites, guarded so it only touches the untouched placeholder and cannot clobber a site that configured its own recipients.
- For fresh installs the work hangs off `install_configure_form`'s submit handler in `mukurtu.profile`. It **cannot** live in `mukurtu_install()`: `install_install_profile` runs *before* `install_configure_form`, so `system.site:mail` is still empty at that point. My first attempt did put it there and silently no-opped; only a real `drush site:install` caught it. This follows the pattern core's own `demo_umami` profile uses.

## The Website feedback form was also unreachable

Fixing the recipient made the form work; it was still unlinked. That took two separate switches being off, which is why it was easy to miss:

- Core ships a footer menu link for `contact.site_page` with `enabled: 0`.
- `block.block.mukurtu_v4_footer` -- the block that renders the core footer menu -- ships with `status: false` **and** in the `sidebar_first` region.

So enabling the menu link alone would still have rendered nothing.

Mukurtu's real footer is its own `mukurtu_footer` block, whose intended contact affordance is `field_footer_contact_email`. But `mukurtu_footer_install()` only ever set the *label* for that field, never an address, and the template renders the mailto only `{% if contact_email_address %}`. So a fresh site had **no discoverable way to contact the site owner at all** - no `/contact` link and no footer email. There is no sensible default for an email address, so the link is the fix.

The link goes in `field_footer_other_links`, which already exists for arbitrary footer links, so it is editable and removable like any other and needs no new field or template work. `mukurtu_footer_update_40003()` adds it to existing sites, but only where Other links is still empty, so a curated footer is never reordered or added to.

Verified end to end: on a fresh install the anonymous homepage renders `<a href="/contact">Contact</a>` in the footer, and `GET /contact` returns 200 with the Website feedback form and its CAPTCHA. On the update path the hook runs, the link resolves to `/contact`, and `mukurtu_footer` lands on schema 40003.

## Verification

Kernel tests cover what they can, but the interesting failures here only appear in a real install, so both paths were run against a live site.

**Real update run** (site at schema 40108 with all three modules installed, restored from a pre-update snapshot):

```
schema: 40123
devel            still installed (correct)
devel_generate   still installed (correct)
mukurtu_dev      still installed (correct)

status report warning:
  title:    Development modules
  value:    Devel, Devel Generate enabled
  severity: Warning
```

**Fresh install** (`drush site:install mukurtu`):

```
total modules installed: 155
devel installed: false | mukurtu_dev installed: false
mukurtu_manager exists: true (perms: 221)     <- the ConfigInstaller cascade did not fire
recipients: ["support@mukurtu.example"]
captcha point for feedback: true
```

**Tests:** 170 `mukurtu_core` kernel tests pass. The single deprecation is pre-existing core noise (`mb_strtolower(): Passing null` from `Config/Entity/Query/Condition.php:39`), present on `main`.

**Lint:** `multilingual-guardrails.sh` passes in strict mode. No SCSS, CSS, JS or Twig is touched, so nothing to recompile.

## Two things reviewers should look at

**1. The second commit is a fix from reviewing the first.** `ModuleInstaller::uninstall()` signals failure by returning FALSE rather than throwing, and in that case uninstalls *nothing*. The first commit ignored that return value. The root cause was a source mismatch: the hook tested presence with `getAllAvailableInfo()` while `uninstall()` gates on `getList()`, and those are different sets. `getList()` only scans the *active profile's* modules directory, so `mukurtu_dev` is absent from it whenever no install profile is active. The check now uses `ExtensionList::exists()`, which is `getList()`-backed, and the return value is checked.

**2. One test leg is not kernel-testable, deliberately.** `KernelTestBase` runs with no active install profile (it has no `install_profile` handling at all), so profile-nested modules are invisible to the extension list and `uninstall()` bails on any list containing one. `DevelModuleRemovalUpdateTest` therefore covers `devel` and `devel_generate`, which live under `web/modules/contrib`, and the `mukurtu_dev` leg is covered by the real update run above. The class docblock says so.

Also worth knowing: `FeedbackFormRecipientInstallTest` tests a function in `mukurtu.profile` but lives in `mukurtu_core`, because `phpunit.xml`'s testsuites only scan `modules/**/tests/src` and a profile-root test would never run in CI. Filed as #2125.

## For reviewers: the removal hook, when you schedule one

No removal release is fixed. When one is chosen it needs its own update hook, and that hook must call `\Drupal::service('extension.list.module')->reset()` before uninstalling. Without it the uninstall aborts the whole update run:

```
The 'Mukurtu CMS' install profile requires 'Mukurtu Dev';
... requires 'Devel Generate'; ... requires 'Devel'
```

Nothing invalidates the cached extension list when an `.info.yml` changes on disk, and the upgrade procedure in `README.md` runs `drush cr` *before* `composer update`, so the cache is rebuilt from the pre-update `.info.yml` and is stale by the time updates run. Found and verified on this branch while the uninstall was still in place, reproduced from a pre-update snapshot with no intervening `drush cr`. Worth not rediscovering.

## Follow-ups filed, not fixed here

- #2124 - the 186-line `$roles['administrator']` permission list in `mukurtu_install()` is dead code, because `Role::grantPermission()` no-ops on `is_admin` roles. Removing `'access devel information'` from it in this PR was itself a no-op. That issue also notes an unguarded `Role::load()` in the same loop.
- #2125 - `phpunit.xml` testsuites exclude profile-root tests.

## Pre-merge checklist:

Notes on the checklist: no SCSS is touched so there was nothing to recompile; the accessibility check found no WCAG 2.1/2.2 AA or ATAG 2.0 surface at all, since no Twig, CSS, JS, render array or user-facing string changes (removing Devel is a net positive, as its Kint output and injected entity tabs are among the least accessible admin surfaces in the distribution); no content entity bundle or view was added or changed. All checks have now passed: Multilingual guardrails, check, playwright-tests, wait-for-tugboat, Tugboat, and Mukurtu v4 - PHP 8.4 (kernel + unit). Tugboat is the notable one here, since it runs `drush site-install mukurtu` followed by `drush updb -y`, independently exercising both the fresh-install and update paths in a clean environment.


michael-wynne-wsu:

Link:
- https://pr2126-7u5zh7vco84nny2rbburc0qjj13hfml1.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa041afd221bf0c533e7173

michael-wynne-wsu:

Link:
- https://pr2126-7u5zh7vco84nny2rbburc0qjj13hfml1.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa041afd221bf0c533e7173

michael-wynne-wsu:

Link:
- https://pr2126-7u5zh7vco84nny2rbburc0qjj13hfml1.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa041afd221bf0c533e7173

michael-wynne-wsu:

Link:
- https://pr2126-7u5zh7vco84nny2rbburc0qjj13hfml1.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa041afd221bf0c533e7173

nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

michael-wynne-wsu:

Link:
- https://pr2126-7u5zh7vco84nny2rbburc0qjj13hfml1.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa041afd221bf0c533e7173

PR #2128, opened 2026-09-08
Remove the dead administrator permission list from mukurtu_install() (#2124)

michael-wynne-wsu:
Closes #2124. Stacked on #2127, which is itself stacked on #2126.

The 186-line `$roles['administrator']` list in `mukurtu_install()` was never applied. The role ships with `is_admin: true`, and core's `Role::grantPermission()` returns early for admin roles:

```php
public function grantPermission($permission) {
  if ($this->isAdmin()) {
    return $this;
  }
  ...
}
```

`getPermissions()` and `hasPermission()` short-circuit the same way, so the role reports every permission as granted while storing none.

Nothing is broken by this today. `is_admin: true` already grants everything, including permissions added by modules in later releases, which is standard Drupal. The problem is that the list reads as the authoritative definition of what an administrator can do, so it invites edits that do nothing. #2126 removed `'access devel information'` from it, and that edit was a no-op.

## What replaces it

A comment recording why the role is absent from `$roles`, and pointing at git history for anyone who later wants to switch it to `is_admin: false` and grant permissions explicitly. The issue raised that deleting the list loses the only record of intended administrator permissions; this is the trade-off I took, and it is easy to reverse if you would rather the list live somewhere readable.

## The unguarded Role::load()

The loop called `$role->grantPermission()` on the result of `Role::load()` with no null check, which would fatal the install. That is reachable rather than theoretical: #2126 established that `ConfigInstaller` silently drops config whose dependencies are unmet and cascades to dependents, which is exactly how one of these roles could go missing with no error raised. Now guarded with a `continue` and a comment saying why.

## Tests

Three tests in `tests/src/Kernel/AdministratorRoleIsAdminTest.php` guard the assumptions the removal rests on, rather than testing the deletion itself:

1. The shipped `user.role.administrator` really does have `is_admin: true`. **This is the important one.** If anyone flips it to FALSE, administrators would silently end up with no permissions at all on a fresh install, because nothing grants them any now.
2. The shipped role carries no explicit permissions, so a non-empty list cannot creep back in.
3. Core still discards permissions granted to an admin role, asserted directly rather than trusted to hold.

These live at the profile root, which is why this PR is stacked on #2127: that PR is what makes profile-root tests run in CI. Putting them in `mukurtu_core` would have re-created the exact misplacement #2127 fixes.

## Verification

Deleting 186 lines from the install path has to be behaviour-neutral, so I ran a fresh install and compared against the baseline captured on #2126:

```
[...]

PR #2127, opened 2026-09-08
Collect profile-root tests in the PHPUnit testsuites (#2125)

michael-wynne-wsu:
Closes #2125. Stacked on #2126.

Every testsuite `<directory>` in `phpunit.xml` was rooted at `modules/**`, so a PHPUnit test placed at the profile root was never collected. CI runs `--testsuite kernel` and `--testsuite unit`, so such a test would sit in the repo silently passing without ever executing.

`phpunit.ddev.xml` had the same gap. The issue only mentioned `phpunit.xml`; both are fixed here.

## The open question in the issue is answered

The issue flagged that profile PSR-4 test namespace registration "is not necessarily set up the way a module's is" and was the part most likely to need work. It needs none. Core's `core/tests/bootstrap.php` scans `$root . '/profiles'` alongside `$root . '/modules'` (line 60) and registers `Drupal\Tests\<extension>\` for any extension that has a `tests/src` directory (line 95). The profile simply had no `tests/src`, so nothing was registered. Creating it is sufficient.

Verified rather than assumed:

```
$ phpunit --testsuite kernel --list-tests | grep FeedbackFormRecipientInstall
 - Drupal\Tests\mukurtu\Kernel\FeedbackFormRecipientInstallTest::testSubmitUsesSubmittedSiteMail
 - Drupal\Tests\mukurtu\Kernel\FeedbackFormRecipientInstallTest::testSubmitFallsBackToSiteConfig
 - Drupal\Tests\mukurtu\Kernel\FeedbackFormRecipientInstallTest::testSubmitIsNoOpWithoutAnySiteMail
 - Drupal\Tests\mukurtu\Kernel\FeedbackFormRecipientInstallTest::testSubmitIsNoOpWithoutFeedbackForm
 - Drupal\Tests\mukurtu\Kernel\FeedbackFormRecipientInstallTest::testFormAlterAppendsSubmitHandler
```

Those are reachable only through the new directory entry, since every pre-existing entry is `modules/**`.

## The test move

`FeedbackFormRecipientInstallTest` moves from `mukurtu_core` to the profile root, which is where it belonged: it `require_once`s `mukurtu.profile` and exercises nothing in `mukurtu_core`. It was only in `mukurtu_core` so that CI would run it, with a docblock apologising for the location. That note is now gone, the namespace is `Drupal\Tests\mukurtu\Kernel`, the group is `mukurtu`, and the profile root is resolved from `__DIR__` instead of walking up from a module path.

[...]

PR #2130, opened 2026-09-08
Add licensing, set the release version, and trim the packaged archive (#2123)

michael-wynne-wsu:
Release-readiness work for the first public tag. Stacked on #2128.

No update hooks and no config or data changes. All metadata, packaging, and one bug fix. Per the decision on #2123 this now targets **4.0.0** rather than 4.0.1: a public 4.0.0 with no LICENSE and every `.info.yml` reading `4.x-dev` is a poor first impression, and the original "hook-requiring changes only" rule had already relaxed in #2126.

## Licensing

The repo shipped **no license at all**, while `package.json` declared `"ISC"` — both wrong and in conflict with the GPL code around it.

- `LICENSE.txt` — the official **GNU GPL v3** text from gnu.org.
- `"license": "GPL-3.0-or-later"` in `composer.json`, `package.json` and the theme's `package.json` (the last had none at all). `composer validate` had been warning about its absence; it now reports `./composer.json is valid` with no warnings.
- `package.json` corrected.
- `libraries/tagify/LICENSE` for the vendored Tagify copy. Its MIT notice is already inline in `dist/tagify.js` and the polyfills, but the minified CSS carries none. The text is reproduced **verbatim from upstream's own headers** rather than reconstructed, so no copyright year is invented.

## License note

Originally committed as GPL-2.0-or-later (copying Drupal core's text); corrected to **GPL-3.0-or-later** on review. This is legally available because Drupal core is GPL-2.0-**or-later**, and that "or later" clause is what permits redistributing the combined work under v3.

Two things worth recording: it diverges from the Drupal ecosystem norm, and drupal.org requires GPL-2.0-or-later for projects it hosts, so this would foreclose ever moving Mukurtu there. No per-file GPL headers exist anywhere in the codebase, so nothing else declares a version that would now disagree. `libraries/tagify/LICENSE` is unaffected — Tagify is third-party MIT code.

`mukurtu/mukurtu-template` still declares GPL-2.0-or-later and needs matching; noted on MukurtuCMS/mukurtu-template#4.

## Version strings

`VERSION.md`, the profile, all 26 custom modules and the theme now read `4.0.0`. Every one of them said `4.x-dev` at `4.0.0-rc`, which is what a site owner saw at `/admin/modules`. The theme had no `version:` key at all.

That is 30 files to keep in step, so `scripts/set-version.sh` now does it:

```
$ scripts/set-version.sh 4.0.1
```

**`VERSION.md` is kept, not replaced.** It stays exactly where it is and the Mukurtu Version element on the Dashboard still reads it (`mukurtu_core.module:1182`), it is still tracked, and it is **not** excluded by `.gitattributes` — verified present in the dist archive. The only thing that changes is *who writes it*: the script rather than a hand edit. Test fixture modules under `tests/` are deliberately skipped — they are not part of the distribution and carry unrelated versions.

The dashboard element is in fact more robust after this PR: the `fclose()` fix below means a missing or unreadable `VERSION.md` degrades to an empty version string instead of fataling the whole dashboard menu build, which is what it did before.

**This does change the release cut**, though: bumping the version should now be `scripts/set-version.sh <version>` rather than editing `VERSION.md` alone, or the 29 other files drift out of step again.

`package.json` also gains `"private": true`. It is build tooling, never published, and its name was still the placeholder `"gulp"` with a pasted CI badge for a description — the first thing GitHub's sidebar reads.

## Packaging

`.gitattributes` marks development-only paths `export-ignore` so they stay out of the dist archive composer downloads from Packagist. **Verified with a real `git archive`** rather than assumed:

```
archive without .gitattributes:  3422 entries
archive with .gitattributes:     2983 entries

excluded:  378 module test files, 33 tests/, 12 .github/, 7 docs/,
           3 scripts/, 3 .tugboat/, 2 .vscode/, phpunit*.xml
retained:  patches/ (16 entries), composer.json, mukurtu.info.yml, VERSION.md
```

`patches/` **must** stay: composer references those files by path while installing dependencies, and a missing one breaks the update outright (see the "patch file could not be downloaded" section of the README). Path and source installs are unaffected, so CI — which adds the profile as a path repository — still sees everything.

The strongest reason here is not size but `.tugboat/settings.tugboat.php`, which carries preview database credentials and `skip_permissions_hardening = TRUE`, and has no business sitting inside a production site's profile directory.

`.gitignore` gains `.claude/`, which until now was only excluded by the machine-local `.git/info/exclude`, so nothing stopped another contributor committing it.

## Dependabot

Composer is configured with `open-pull-requests-limit: 0`, which disables version-update PRs while leaving **security updates enabled**. That is deliberate: `drupal/facets` is held below 3.0.4 and `drupal/pathauto` at `~1.14.0` against known-broken releases, and routine version PRs would fight those pins. npm (three directories) and github-actions get ordinary version updates.

## Bug fix

`mukurtu_core.module` called `fclose($version_file)` outside both guards, so a missing or unreadable `VERSION.md` left the variable undefined and `fclose(NULL)` raised a `TypeError` on PHP 8, fataling the dashboard menu build. Directly relevant here, since `.gitattributes` work is exactly how `VERSION.md` could have gone missing from the archive.

## Verification

- 183 kernel tests pass across `mukurtu_core`, `mukurtu_footer` and the profile root. The single deprecation is pre-existing core noise (`mb_strtolower(): Passing null`).
- No test asserted `4.x-dev`, `VERSION.md` or the version dashboard link, so the version bump breaks nothing.
- `composer validate` clean.
- `multilingual-guardrails.sh` passes in strict mode. No new user-facing strings.
- No SCSS, CSS, JS or Twig touched.

## Deliberately not in this PR

- **README rewrite.** It still carries the "testing and feedback purposes only" banner and beta-hopping upgrade instructions. It is substantial user-facing prose and wants reviewing on its own.
- **CHANGELOG.md / SECURITY.md / CONTRIBUTING.md.**
- **composer wildcard cleanup** (15 `*` constraints, the missing `php` constraint, the dead `views_combine` patch entry and orphaned pathauto patch). That can change dependency resolution, so it should not ride along with licensing metadata.

## Pre-merge checklist:

Checklist notes: no update hooks needed, nothing here touches config or data. No accessibility surface: no Twig, CSS, JS, render arrays or user-facing strings; the one code change removes a fatal. No SCSS to recompile. No new tests, but the existing suites were run against the change and no test asserted the old version strings. **Base branch is `2124-remove-dead-administrator-permissions` (#2128), not `main`** — retarget once #2126, #2127 and #2128 merge, in that order. Checks box unticked until CI reports.


michael-wynne-wsu:

Link:
- https://pr2130-4uon3qmcptmwltn7kmbqsrizn2h9pah0.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa070fa868e1053c3664831

michael-wynne-wsu:

Link:
- https://pr2130-4uon3qmcptmwltn7kmbqsrizn2h9pah0.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa070fa868e1053c3664831

michael-wynne-wsu:
Superseded by #2132, which is the same three commits flattened onto `main` (plus a composer lock content-hash fix found in the process).

Flattened because this work touches an entirely separate set of files from #2126, #2127 and #2128, so it does not need to sit behind them in a retargeting chain. `docs/stacked-prs.md` documents three occasions where a stacked PR merged into a stale base and never reached `main`; five links was more of that risk than this change warrants.

PR #2134, opened 2026-09-08
Report unconfigured server-side dependencies on the status report (#2123)

michael-wynne-wsu:
Item 1 of the remaining release-readiness work. Targets `main` directly; touches no files shared with any other open PR.

Two requirements, both surfacing server-side dependencies that currently fail silently: the private file system, and the media processing tools.

# Private file system

## The gap

Every file-based media type in Mukurtu stores its source field under `private://` — audio, document, image and video. That is what lets cultural protocols control who can download a file. Without `$settings['file_private_path']` the `private://` stream is never registered, those upload destinations resolve to an invalid scheme, and **media uploads fail**.

Nothing reported this. Core's file system requirement iterates the public, private and temporary directories but opens with:

```php
foreach ($directories as $directory) {
  if (!$directory) {
    continue;
  }
```

So an unset private path is **skipped silently** and only writability is ever checked (`SystemRequirementsHooks::runtime()`). The result is a site that looks healthy on the status report and cannot accept a media upload.

## Reproduced, not inferred

Against a real site with no private path set:

```
registered stream wrappers: assets, public, temporary, module, theme
is private:// registered?   false
audio upload target:        private://[date:custom:Y]-[date:custom:m]
valid scheme?               false
```

And after this change, on that same site:

```
title:    Private file system
value:    Not configured
severity: Error
```

## Design notes

**`hook_runtime_requirements()` rather than `hook_requirements()`.** `requirements` is on core's static deny list for attribute-based hooks and throws a `LogicException`; `runtime_requirements` is not, so this can be an OOP `#[Hook]` class per project convention.

**Runtime-only is deliberate.** The private path is configured *after* installation — the README's post-installation steps say so — and an install-phase check would fail on every new site.

**The description names the cache rebuild explicitly.** Setting the path on a running site leaves `private://` unregistered until caches are cleared, so the fix looks like it did not work. That is the step people miss.

**The stream wrapper manager is injected** so tests can cover the state a site is in between editing `settings.php` and rebuilding caches, which a kernel test cannot otherwise reach — the container will not register `private://` from a setting alone.

# Media processing tools

`mukurtu_media` shells out to **three** binaries when media is saved:

| Binary | Package | Call site | Provides |
|---|---|---|---|
| `pdftotext` | poppler-utils | `Document::preSave()` | extracts PDF text into `field_extracted_text` so documents are searchable |
| `pdftoppm` | poppler-utils | `Document::generateThumbnail()` | renders page one as a PNG thumbnail |
| `ffmpeg` | ffmpeg | `Video::generateThumbnail()` | extracts the first frame as a video thumbnail |

Each call site checks for exit code 127 and returns early if the tool is absent. **Every one of those early returns is silent** — nothing shown to the person uploading, nothing written to the log. A server without poppler-utils simply never produces document thumbnails or searchable PDFs, with no indication why.

The check probes each binary with **the same argument its own call site uses** (`pdftotext -v`, `pdftoppm --help`, `ffmpeg -version`), so it cannot disagree with the code that depends on it, and treats only exit code 127 as missing.

It also handles PHP configured without `exec()` — which is what the consuming code uses, so in that case the features cannot work at all and the honest report is that the tools cannot be checked rather than a guess either way.

Severity is **Warning**, not Error: media still uploads and the site works, only the derived features are lost. That keeps Error meaning something is actually broken, as with the private file system check above.

### A true positive on the first run

Checked against a real container, which has poppler-utils but no ffmpeg:

```
value:    ffmpeg not found
severity: Warning
```

Video thumbnails have been failing silently in that environment.

### Correcting myself

An earlier revision of this PR said nothing referenced `pdftotext` or `ffmpeg` and that the README's claims looked stale. That was wrong twice over: the binary names are assembled into `$cmd` variables so a literal grep missed them, and I never searched for `pdftoppm` at all. The README was right, and there is a third binary it does not mention by name (though installing poppler-utils covers both of its tools).

Similarly, my first pass concluded no field used the private scheme, because `FieldStorageConfig::loadByName()` returns NULL for code-defined bundle fields. The media source fields only appear through `entity_field.manager`. Worth knowing if you ever audit field settings that way.

## Verification

- **10 new kernel tests** across both requirements: unset/configured/registered states for private files, and present/missing/all-missing/exec-disabled for the tools.
- **166 `mukurtu_core` kernel tests pass.** The single deprecation is pre-existing core noise.
- `multilingual-guardrails.sh` passes in strict mode; all strings go through `$this->t()`.
- Both requirements verified rendering on a live site.

## Pre-merge checklist:

Checklist notes: no update hook needed — this only reports state, it changes no config or data. Accessibility: the one new user-facing string is plain text in core's requirements table. No SCSS. Base branch is `main`, so no retargeting. Checks box unticked until CI reports.


michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

michael-wynne-wsu:

Link:
- https://pr2134-rfqdoumyekl8ylsmmximzxdggcmnry39.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa084a1868e1053c3696dd6

brian-wsu (approved):
I see (under status report) that "Media processing tools" are missing and an error about the private files not being set on a new installation on a server without those tools.

nick-deer (approved):

PR #2135, opened 2026-09-08
Tighten composer constraints and lift the facets pin (#2123)

michael-wynne-wsu:
Item 2 of the remaining release-readiness work. Targets `main` directly.

You asked me to check the upstream bugs behind the two version pins before touching them. **One is fixed, one is not.**

## facets: pin lifted to `^3.0.7`

The pin existed for [drupal.org #3617839](https://www.drupal.org/node/3617839): a regression where `FacetsFilter::query()` fatals with `Call to a member function getSearchApiQuery() on null` whenever a facets exposed filter block renders standalone, outside the view's own execution — **exactly** what `mukurtu-browse.html.twig` does for `/digital-heritage`.

That issue is now marked **Fixed** upstream, and 3.0.7 carries the guard:

```php
// A separate exposed-filter block instance has no Views query to process.
// Empty facets should remain hidden rather than dereferencing a NULL query.
if (!$processed_facet && $this->query === NULL) {
  return $form;
}
```

> [!IMPORTANT]
> The floor is `^3.0.7`, **not** `^3.0`. I checked each release individually: 3.0.4, 3.0.5 and 3.0.6 all still lack that guard. Relaxing to `^3.0` would have let the broken versions straight back in — which is how this drifted onto 3.0.4 the first time, per #2013.

## pathauto: pin retained at `~1.14.0`

Checked and deliberately left alone, but **my original reasoning here was wrong** and is corrected below.

- **1.15.0 is still the newest release**, so there is nothing newer to move to.
- The reason for the pin is a **core** `ModuleInstaller` stale-container bug, not something pathauto can fix on its own.

### Correction: pathauto is probably not the trigger

An earlier revision of this section said pathauto 1.15's OOP hooks hit "a third unpatched path" in `ModuleInstaller`, because they implement `entity_base_field_info`, `field_info_alter`, `entity_insert/update/delete`, `token_info`/`tokens` and `help` rather than the two paths the existing core patches cover. I took that mechanism from commit `b2142dfd8`'s message. It does not hold up.

`patches/module-installer-stale-module-handler-modules-installed.patch` — written in this repo, before any of this — already names the actual trigger in its own header:

> This stayed dormant until a module implementing `hook_modules_installed()` via `#[Hook]` … was first installed alongside something that rebuilds the container - e.g. `drupal/extlink 3.0.0`'s `Drupal\extlink\Hook\ExtlinkHooks::modulesInstalled()`

extlink is always installed by this profile, and its `#[Hook('modules_installed')]` fires at `ModuleInstaller.php:257` — the exact call site that patch already fixes. So there is no evidence of a third unpatched path, and pathauto appears to have been coincidental rather than causal. Independent testing on core 11.4.6 in a separate sandbox reached the same conclusion.

### Why the pin still stays in this PR

The pin is probably liftable, but not on this evidence and not here:

- It is out of scope for a PR about wildcards, `php` and facets.
- The failure `b2142dfd8` recorded was observed **on Tugboat**, not locally, so a local sandbox result does not settle it.
- This pin has already been loosened once, in `70ad6332b`, on the assumption that a core upgrade had fixed the underlying bug. It had not, the lock drifted back to 1.15.0, and Tugboat Playwright runs failed at the login step on every push until it was re-pinned. Doing that a third time on a single local test would be careless.

Lifting it deserves its own change, verified on Tugboat — the environment where it actually broke.

### The real blocker, now being addressed

`patches/module-installer-stale-module-handler-modules-installed.patch` ends with:

> Upstream: this fix should be filed against Drupal core … Once a core issue is filed, replace this comment with the issue URL so the patch can be removed when the fix lands in a Drupal release.

That issue had never been filed, which is why the underlying bug can never be resolved by a core release. It is being filed now. **When it is, both core patch headers should be updated with the issue URL**, per their own instruction.

## 12 wildcard constraints replaced

`*` becomes `^major.minor` at the locked version:

`ctools` `^4.1` · `embed` `^1.10` · `entity_browser` `^2.17` · `entity_embed` `^1.7` · `entity_reference_revisions` `^1.14` · `redirect` `^1.13` · `restui` `^1.22` · `search_api` `^1.41` · `search_api_glossary` `^4.2` · `token` `^1.17` · `views_bulk_operations` `^4.4` · `nominatim-provider` `^5.8`

Any of these could have taken a major version on a plain `composer update`. Two are **patched by this repo** (`search_api`, `search_api_glossary`), where a major bump would have broken patch application rather than merely surprising us.

## `php: >=8.4`

There was no PHP constraint at all, so composer would install on 8.1. This matches what CI tests and what `README.md` already claims.

It is deliberately stricter than Drupal core's own `>=8.3.0`: 8.3 is never exercised by CI, and claiming an untested version seemed worse than excluding it. Reversible if you would rather widen it, ideally alongside an 8.3 CI job.

## Dead patch cleanup

- Removed the **`drupal/views_combine` patch entry**. That package is in neither `require` nor `require-dev`, and it was the only patch still fetched from a mutable `drupal.org/files/issues/` URL rather than a pinned SHA.
- Deleted **`patches/pathauto-source-module-missing-php8-attribute.patch`**, left on disk by `b2142dfd8` when its `composer.json` entry was dropped.

## Verification

The lock moves **facets 3.0.3 → 3.0.7 and changes nothing else**.

A note for anyone repeating this: `composer update --lock` is *not* sufficient when a constraint should move a version. It refreshes metadata only, so it left the lock claiming 3.0.3 against a `^3.0.7` constraint — an inconsistency `composer validate` catches but `composer install` only warns about. `composer update drupal/facets` was needed.

`composer validate` is clean apart from the missing `license` warning, which #2132 fixes on its own branch.

### Verified end to end on a purpose-built site

An earlier revision of this PR said this could not be tested locally, because the usual DDEV environment installs the profile as a published zip. I built a throwaway DDEV project wired to this branch as a **path repository** (the arrangement CI uses) and installed the site, giving the zero-content state the original bug needed.

Controlled test, with only facets changing between runs:

| facets | `/digital-heritage` | `/browse` |
| --- | --- | --- |
| **3.0.4** (known broken) | **500** | **500** |
| **3.0.7** (this PR) | **200** | **200** |

The 500 is the same bug, not an unrelated failure:

```
Error: Call to a member function getSearchApiQuery() on null in
Drupal\facets_exposed_filters\Plugin\views\filter\FacetsFilter->valueForm()
(line 242 of .../facets_exposed_filters/src/Plugin/views/filter/FacetsFilter.php)
```

Line 242 is exactly the unguarded call present in 3.0.4 and guarded in 3.0.7. Restoring `^3.0.7` returned both pages to 200 with no new errors logged, so the result round-trips rather than being order-dependent.

That confirms three things: the reproduction is real on a fresh install, 3.0.7 fixes it, and the test is sensitive enough to have caught the regression in the first place.

## Pre-merge checklist:

Checklist notes: no update hook needed — dependency metadata only, no config or data. No accessibility surface, though the facets bump is what stops `/digital-heritage` fataling, so Playwright is the relevant check. No SCSS. Base branch is `main`. Checks box unticked until CI reports, and Tugboat plus Playwright matter more than usual on this one.


michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

michael-wynne-wsu:

Link:
- https://pr2135-v8eiojygcezw2hxeqkbcv1lo5xtvmnfi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa08a6c47cc3359218556e3

brian-wsu (approved):
Everything is looking good to me.

nick-deer (approved):

PR #2154, opened 2026-09-10
Make the Tugboat wait and Playwright main job report their real failure cause

michael-wynne-wsu:
## Summary

`Tugboat Playwright Tests` has failed on **every push to main for at least its
last 100 runs**, going back to 2026-08-28 at minimum. Nothing in the last 100
runs passed.

The cause is not the tests and not the application. `TUGBOAT_MAIN_URL` points
at a Tugboat preview that no longer exists:

```
https://main-edn9hr3gbdzf34z7leqdiajdu5qt3grc.tugboatqa.com/user/login
  -> 404, <title>Tugboat - Preview Not Found</title>
```

Playwright drives that 404 page, finds no login form, and reports six failing
`default-content` tests about four minutes in, all of them
`TimeoutError: locator.fill` on `getByLabel('Username')`. That looks like a
login regression. It is a missing preview.

## What this PR does

Adds a readiness check to `playwright-tests-main` before the suite runs.

The pull request path already does exactly this, in `wait-for-tugboat`: it
polls `/user/login` and accepts only a response that is genuinely the rendered
Drupal login form, because Tugboat serves a "Preview is resuming" holding page
with a `200` while a suspended preview wakes up, so a status code alone is
misleading. That hardening was never applied to the push path. This applies the
same check there.

It also treats a `Preview Not Found` body as terminal rather than polling a
deleted preview for the full ten minutes, since that state never resolves.

## What this PR does not do

**It does not make the job pass.** The URL is still dead. Fixing that means
pointing the `TUGBOAT_MAIN_URL` repository variable at the current main base
preview, which needs Tugboat dashboard access. This PR makes the job say so, in
one line, in seconds, instead of masquerading as six login failures.

## Verification

Ran the preflight logic against four cases:

| Case | Result |
| --- | --- |
| `TUGBOAT_MAIN_URL` unset | fails immediately, names the variable and where to set it |
| The real deleted preview | fails in 0s with "Preview Not Found" and what to do |
| A live PR preview still waking up | polls once, then accepts, exit 0 |
| A local Drupal site | accepts immediately, exit 0 |

The last two matter as much as the first two: a preflight that rejects healthy
previews would be worse than none.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (CI only, none needed)
- [x] I have run an accessibility check, and resolved any issues. (CI only, no UI)
- [x] I have recompiled SCSS if required. (not required)
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges (see [docs/stacked-prs.md](../docs/stacked-prs.md)). (base is `main`)
- [x] If I added a content entity bundle... (not applicable)

---

## Review notes

Ran the `pr-review-expert` review over this change. Blast radius is low: one
file, one job, and that job only runs on push to `main`, so nothing about the
pull request path changes. No secrets, no `${{ }}` interpolated directly into a
`run:` block (the URL is passed via `env:`, which avoids script injection), no
new dependencies.

The review found one real issue, fixed in e71c8da: the readiness check
normalised the base URL for its own use, but the test step still read the raw
`TUGBOAT_MAIN_URL`, so the two could disagree. The variable is currently stored
with a trailing slash. The checked URL is now passed through as a step output.

**Caveat worth knowing when reviewing:** `playwright-tests-main` is gated on
`github.event_name == 'push'`, so it does not run on this pull request and CI
here cannot exercise the change. It shows as `skipping`. The logic was
therefore verified by running the step's script directly against four cases,
listed above, including the real dead URL and two live Drupal sites.

---

## Second change: stop failing the Tugboat wait on a single error reading

Added after this PR's own CI hit the exact problem, which is the best argument
for it.

`wait-for-tugboat` failed here in **7 seconds** while the `Tugboat` check itself
went on to **pass**. The poll loop treated the first `error` reading as final,
but the Tugboat commit status reports `error` and then recovers to `success` on
the same commit. A re-run with no code change cleared it, which is the signature
of a status race rather than a build failure.

That has now happened twice in one session, and the first time it was mistaken
for a stacked-PR problem that did not exist.

The loop already distrusted an early *success* reading, with a comment noting
the status "has been observed to flip success -> pending -> success" mid-rebuild.
It gave an early *error* reading no such benefit of the doubt. That asymmetry is
the bug. An error or failure state now has to persist across two consecutive
polls, costing one 30 second interval on a genuine failure.

Verified by simulating status sequences against the loop rather than waiting for
the race to recur:

| Sequence | Expected | Result |
| --- | --- | --- |
| `error` then `success` | pass (the real case) | pass |
| `error`, `error` | fail | fail after 2 consecutive |
| `pending`, `pending`, `success` | pass (happy path unchanged) | pass |
| `pending`, `error`, `error` | fail | fail once 2 land in a row |
| `error`/`pending` alternating forever | still fails | falls through to the MAX_WAIT timeout |

That last row is the accepted trade and is documented in the comment: such a
status never reaches two in a row, so it fails slowly rather than fast. It has
never been observed.

Both changes in this PR are the same shape of problem: this job reporting a
cause that is not the real one. The first made a dead preview look like six
login failures; this one made a transient status look like a build failure.


michael-wynne-wsu:

Link:
- https://pr2154-kvv7qjdhckd7wlknyabnpfctdvfayepq.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2152cd221bf0c539b4ca8

michael-wynne-wsu:

Link:
- https://pr2154-kvv7qjdhckd7wlknyabnpfctdvfayepq.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2152cd221bf0c539b4ca8

michael-wynne-wsu:

Link:
- https://pr2154-kvv7qjdhckd7wlknyabnpfctdvfayepq.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2152cd221bf0c539b4ca8

michael-wynne-wsu:

Link:
- https://pr2154-kvv7qjdhckd7wlknyabnpfctdvfayepq.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2152cd221bf0c539b4ca8

nick-deer (approved):

michael-wynne-wsu:

Link:
- https://pr2154-kvv7qjdhckd7wlknyabnpfctdvfayepq.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2152cd221bf0c539b4ca8

Issue #2160, opened 2026-09-10
Beta-installed sites still grant Full HTML to authenticated users


michael-wynne-wsu:
## Summary

Sites that **fresh-installed a 4.0.x beta** have `use text format full_html`
granted to the `authenticated` role, and nothing will ever remove it.
`mukurtu_core_update_40118()` was written to revoke it, but those sites were
seeded past that hook and never ran it. 4.0.1 removes the hook entirely, so the
grant is now permanent on those sites.

4.0.1 fixes this for **new** installs only.

## Why those sites were missed

`mukurtu_install()` granted the permission programmatically, at
`mukurtu.install:346` as of `4.0.0-beta37`:

```php
$roles['authenticated'] = [
  ...
  'use text format full_html',
```

`mukurtu_core_update_40118()` revoked it, and its docblock explains the
reasoning: Full HTML "is meant to be an editorial/admin format", and Basic HTML
"already covers their normal iframe/media-embed needs" without granting
"object/embed/video tags".

But that hook is numbered **40118**, and `mukurtu_core` reached **40122** by
beta37. A fresh install is seeded at `max(available updates)` without running
any of them, so any site installed from a beta at or above 40118 skipped it:

| Site history | Ran 40118? | Has full_html? |
| --- | --- | --- |
| Installed on a beta below 40118, then updated | yes | no |
| **Fresh-installed a beta at or above 40118** | **no** | **yes** |
| Fresh install of 4.0.1 | n/a | no (fixed) |

Confirmed empirically: a fresh `4.0.0-beta37` install, updated through 4.0.0 to
4.0.1, still has the permission. It showed up as the only unexplained entry in a
config parity diff between that site and a fresh 4.0.1 install.

## Impact

Bounded, and worth stating precisely so this is not over-read. This profile's
Full HTML has `filter_html` **enabled** with an allow list. It does **not**
permit `script`, `form`, `style` or `svg`. The delta over Basic HTML is:

```
object, embed, video, div, h1, small, b, i, pre, s,
table, tr, td, th, thead, tbody, tfoot, caption
```

So the real exposure is `<object>` and `<embed>`, which can load arbitrary
external content. That is hardening rather than an urgent vulnerability, which
is why it is deferred rather than hot-fixed.

## Why this is not in 4.0.1

4.0.1 is deliberately a no-op for sites updating from 4.0.0: it removes every
update hook and adds `hook_update_last_removed()` baselines. Adding a hook to it
[...]

Issue #2156, opened 2026-09-10
Taxonomy term imports cannot map the UUID column that external exports emit


michael-wynne-wsu:
## Summary

The external CSV exporters emit a `UUID` column for taxonomy terms, but none of
the 16 taxonomy term import strategies accept `UUID` as a source. So a taxonomy
export taken from one site cannot be mapped back in on another, which is the
round trip the external exporters exist to support.

Every other entity type is fine. Of the 39 bundles present in both an exporter
and an import strategy, taxonomy terms are the only ones with this gap.

## Detail

The exporters split by intent:

| Exporter | Bundles emitting `uuid` |
| --- | --- |
| `default_external_with_media`, `default_external_metadata_only` | 45 |
| `default_local_with_media`, `default_local_metadata_only` | 1 (`user__user`) |

That split is correct. UUID is the identifier that survives moving content
between sites, so the external exporters emit it and the local ones, used for
edit-and-reimport on the same site, do not.

The import side does not match. These 16 strategies have no `UUID` source:

```
taxonomy_term__category                     taxonomy_term__location
taxonomy_term__community_type               taxonomy_term__media_tag
taxonomy_term__contributor                  taxonomy_term__people
taxonomy_term__creator                      taxonomy_term__place_type
taxonomy_term__format                       taxonomy_term__publisher
taxonomy_term__interpersonal_relationship   taxonomy_term__subject
taxonomy_term__keywords                     taxonomy_term__type
taxonomy_term__language                     taxonomy_term__word_type
```

For comparison, `taxonomy_category_all_fields.yml` currently maps:

```
Term ID -> tid
Locale -> langcode
Name -> name
Description -> description
Thumbnail Image > File ID -> field_thumbnail_image/target_id
Thumbnail Image > Alternative text -> field_thumbnail_image/alt
Parent -> parent
Default translation -> -1
```

`UUID` is absent, while node, media, paragraph, community, protocol and user
strategies all include it.

## Impact

Exporting taxonomy terms from site A and importing to site B leaves the UUID
column unmapped. Terms are matched by name or created fresh instead of being
recognised as the same term, so vocabularies diverge across sites rather than
staying aligned. This is silent: both the export and the import report success.

## Suggested fix

Add a `UUID -> uuid` mapping to each of the 16 strategies in
`modules/mukurtu_import/config/install/`, matching how the other entity types
already do it.

Two things to handle when doing so:

[...]

Issue #2153, opened 2026-09-10
Add an update hook to align the submission coverage map with other geofield fields


michael-wynne-wsu:
## Summary

The submission form's coverage map does not match the coverage map on every
other content type. 4.0.1 fixes this for **new** installs by correcting the
shipped config, but deliberately ships no update hook, so **existing sites keep
the old values**. This issue tracks adding that update hook in a future
release.

## What was wrong

`core.entity_form_display.node.digital_heritage.submission` configures
`field_coverage` with the `geofield_mukurtu` widget, the same widget the six
other coverage maps use, but with different settings:

| Setting | Every other coverage map | Submission form (before 4.0.1) |
| --- | --- | --- |
| `toolbar.drawCircle` | `true` | `false` |
| `map_position.center.lat` | `20` | `0.0` |
| `map_position.center.lon` | `10` | `0.0` |
| `map_position.singlePointZoom` | `12` | absent |

The practical effect: contributors using the submission form get no circle
drawing tool, and the map opens centred on the Gulf of Guinea rather than on
the shared world view.

## Why it drifted

The values were each set by an update hook during 4.0.x. Every one of those
hooks, and every test that covered them, named the same six bundles that
existed when the hook was written. The submission form display was added later
and nothing ever looked at it.

Found while replacing those tests in #2123: the replacement
`ShippedGeofieldWidgetDefaultsTest` discovers displays by scanning for the
`geofield_mukurtu` widget instead of listing them, and it failed on the seventh
display immediately. That test now guards against the same class of drift for
any future bundle.

## What 4.0.1 does and does not do

- **Does:** corrects `modules/mukurtu_submissions/config/install/core.entity_form_display.node.digital_heritage.submission.yml`, so new installs are correct.
- **Does not:** ship an update hook. 4.0.1 is deliberately a no-op for sites updating from 4.0.0, and `docs/update-hooks.md` calls out form display widget settings as the case better recommended in release notes than forced by a hook, since sites routinely customise form displays.

Existing sites can enable the circle tool themselves at
**Structure > Content types > Digital Heritage > Manage form display** on the
submission form mode.

## What this issue asks for

Add a `hook_update_N()` in a release that is not constrained to be a no-op,
setting these values on existing sites. Two things to decide when doing it:

1. **Numbering.** `mukurtu_submissions_update_last_removed()` returns `40014`,
   and `mukurtu_core` overflowed the two-digit band as far as `40123`, so pick
[...]

PR #2151, opened 2026-09-09
Add an update hook numbering lint and rewrite the hook docs (#2123)

michael-wynne-wsu:
First of two PRs for the 4.0.1 hook strip. This one is deliberately small: it lands the *rules* so they can be reviewed before the bulk change deletes 520 hooks.

## The lint

`scripts/lint/update-hook-numbering.sh` enforces two rules, both of which have already bitten this project:

1. **No number repeats within a file.** Drupal runs only one of a duplicated pair, so the other silently never executes. `mukurtu_core.install` still carries a comment about an incident of exactly this.
2. **Every number exceeds the module's `hook_update_last_removed()`**, when it has one. Anything at or below that value never runs, because `update_get_update_list()` only lists updates above a site's installed version and a fresh install is seeded at `last_removed`. Dead code that looks live.

It follows `multilingual-guardrails.sh`: same `MUKURTU_LINT_STRICT` switch, and the same self-test that proves each pattern matches its own fixture before the script is trusted against the tree.

### Verified it can actually fail

Rule 2 **cannot fire on `main` yet**, because no module declares `last_removed`. So "OK" from a run against the real tree proves nothing about it, and testing against fixtures was the only way to know it works:

```
fixture: duplicate 40002, plus last_removed 40123 with hooks 40100 / 40123 / 40124

DEAD HOOK ... _update_40100 is not greater than last_removed (40123)
DEAD HOOK ... _update_40123 is not greater than last_removed (40123)
DUPLICATE numbers in ...: 40002
exit=1 under MUKURTU_LINT_STRICT=1
```

`_update_40124` correctly passes, and the boundary case (equal to `last_removed`) is treated as dead, matching Drupal's semantics.

On `main` today it checks 28 files containing hooks and passes.

### The job rename

The lint job is renamed from "Multilingual guardrails" to **"Lint"**, since it now runs two. I checked before doing that: `main` has no branch protection and its only ruleset ("wait for Tugboat on main") is `enforcement: disabled`, so no required-check name breaks. It will show up as a new check name in the UI.

## The docs rewrite

`docs/update-hooks.md` kept its useful guidance but had gone stale and omitted the things that have cost the most time in this work:

- **A hook never runs on a fresh install.** Drupal seeds a new module's schema version to its highest available number without invoking any of them, so a change made only in a hook reaches existing sites and never new ones. The same end state has to be in `config/install` or the two diverge permanently.
[...]

PR #2161, opened 2026-09-10
Strip all update hooks and declare the 4.0.1 baseline (#2123)

michael-wynne-wsu:
Part of #2123. **Stacked on #2151** (`2123-update-hook-numbering-lint`) - that
PR adds the numbering lint this one relies on, and must merge first.

Does **not** bump the version. That is deliberately the last step of the release.

## What this does

Removes every `hook_update_N()` the 4.0.x betas accumulated and replaces them
with `hook_update_last_removed()` baselines, so 4.0.1 is a clean starting point
for new sites and a no-op for testers coming from 4.0.0.

| | |
| --- | --- |
| `hook_update_N()` removed | 520 |
| `hook_update_dependencies()` removed | 5 |
| `hook_update_requirements_alter()` removed | 3 |
| `hook_update_last_removed()` added | 28 |
| Lines | +2,808 / -23,672 |

Functions were removed with PHP's tokenizer rather than a regex, so braces
inside strings, heredocs and closures could not be miscounted. Every baseline
value is computed from the `4.0.0` tag and re-verified against it independently.

## Verified against real sites, not just CI

CI has no functional or install test, so none of the below is caught
automatically. All four scenarios were run on a purpose-built DDEV sandbox,
using `git archive` per ref so each step tested what a release actually ships.

**1. Fresh 4.0.1 install - pass.** Installs cleanly. No pending updates. All 26
installed modules with a baseline seeded at exactly their `last_removed`, zero
mismatches. No devel modules. Guard silent.

**2. beta37 to 4.0.0 to 4.0.1 - pass.** 4.0.0 ran exactly the three hooks it
should (`mukurtu_core_update_40123`, `mukurtu_footer_update_40003`,
`mukurtu_local_contexts_update_40007`), then 4.0.1 was a clean no-op: no pending
updates, no requirement errors. This is the release's central promise.

**3. beta37 straight to 4.0.1 - the guard warns but does not block.** See below.

**4. Config parity - 7 differences, all explained.** Filtering per-site `uuid`
and `default_config_hash` (1,121 of 1,124 raw diffs) leaves devel's config and
`core.extension` (expected: 4.0.0 dropped devel from dependencies but
deliberately does not uninstall it), the submission geofield display (expected,
#2153), the test sites' own names, and one real finding (#2160).

## An important correction

I built the guard on the premise that **core does not enforce
`hook_update_last_removed()`**. That is wrong, and I had written it into 30
places plus `docs/update-hooks.md`.

`SystemRequirementsHooks::checkRequirements()` compares every module's installed
schema version against its `hook_update_last_removed()` and reports an
error-severity requirement per module that is behind. I had checked
`update_get_update_list()` and `_update_fix_missing_schema()`, concluded core
read the value nowhere else, and missed this. It only surfaced when core's error
appeared beside mine during scenario 3.

Two consequences, both now reflected in the code and docs:

- **The update-phase hook is not load-bearing for detection.** What it adds is a
  single message naming 4.0.0 as the release to install first, where core says
  only "update to an intermediate version first".
- **The runtime hook is the half that matters.** Core gates its check on
  `$phase == 'update'`, so nothing appears on the status report. Verified on a
  behind site: of 36 runtime requirements across 16 modules, ours is the only
  one that flags it.

Separately: **an error-severity requirement does not stop `drush updb -y`.**
Drush raises it as a confirmable prompt, so it prints and proceeds, exiting 0
with "No pending updates". Only an explicit `--no` aborts. Our own
`.tugboat/config.yml` uses `-y`. Treat it as a loud warning, not a gate.

## Four bugs found while writing the replacement tests

None were in scope. All surfaced because the new tests read shipped config
instead of running hooks.

1. **Full HTML was granted to every logged-in user on fresh installs.** Hook
   40118 revoked it in 4.0.0, its docblock stating "the shipped authenticated
   role has never granted it" - true of `user.role.authenticated.yml`, untrue of
   `mukurtu_install()`, which granted it programmatically. **Fixed here**, no
   hook needed: existing sites were already corrected by 40118, so this
   converges fresh installs onto where they already are. To be precise about
   scope, this profile's Full HTML has `filter_html` enabled and permits no
   `script`, `form`, `style` or `svg`; the delta over Basic HTML is `object`,
   `embed`, `video` plus formatting tags.
2. **The submission coverage map had drifted** from the other six (no circle
   tool, centred on 0,0, no `singlePointZoom`). Config fixed here; #2153 tracks
   the update hook for a release that is not constrained to be a no-op.
3. **Taxonomy imports cannot map the UUID column external exports emit**, across
   all 16 taxonomy bundles. Filed as #2156.
4. **A test of mine was giving false assurance.** An earlier commit in this
   branch added `testAuthenticatedUsersDoNotGetFullHtml`, which read only the
   YAML and was green while the install code granted the permission. Deleted and
   replaced with one asserting the union of both sources.

Plus #2160, from the parity check: sites that **fresh-installed a late beta**
were seeded past hook 40118 and never ran it, so they still hold Full HTML.
Deferred to 4.0.2 to keep 4.0.1 a no-op.

## Test rework

67 files invoked update hooks. Only 7 were purely transition tests; 50 carried
invariants still true of a fresh install, and 10 were mixed. Deleting all 67, as
originally planned, would have destroyed real coverage: the
`mukurtu_browse_auto_index` 64-key MySQL budget, the export-header to
import-mapper contract, the "protocol_affiliate must not have `view media`"
negative, and the security-relevant comment defaults among them.

They are replaced by 10 test classes executing **206 tests / 4,651 assertions**,
asserting shipped config directly. That is a stronger check than what it
replaces: the old tests stripped keys out of the shipped file before asserting,
so they could pass while the file itself was wrong.

Every replacement was **mutation-proven** rather than accepted on a green run -
each was shown to fail when the invariant is broken, with the intended message.

Net: 248 to 201 test files, 1,019 to 860 test methods, but the replacements are
data-provided so executed tests went up in the consolidated areas.

## Suites

- Unit: 255 tests, 4,752 assertions
- Kernel: 814 tests, 26,135 assertions (10 pre-existing deprecations, none in
  the files touched here; CI does not fail on deprecations)
- `scripts/lint/update-hook-numbering.sh`: 28 baselines checked, clean
- `scripts/lint/multilingual-guardrails.sh`: clean

## Accessibility

Ran the WCAG check. No new WCAG 2.1/2.2 AA or ATAG 2.0 failures. The status
report entry inherits core's accessible pattern (severity is conveyed by a
visually-hidden label and group heading, not colour alone) and is plain text, so
it also renders correctly under `drush core:requirements`. The circle tool
inherits the existing keyboard scaffolding in `mukurtu-leaflet-widget.js`.

Two notes worth carrying forward, neither introduced here: the submission form
swaps geofield to the accessible `geofield_mukurtu_latlon` widget via
`SUBMISSION_WIDGET_OVERRIDES`, so `drawCircle` is inert on the public form in
practice and this change is config bookkeeping; and a circle serialises to a
128-vertex polygon that the accessible editor renders as 128 rows, which is
pre-existing and deserves its own issue.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (This PR removes them; the guard and baselines are the mechanism. No new hooks, deliberately, so 4.0.1 stays a no-op.)
- [x] I have recompiled SCSS if required. (No SCSS, CSS, JS or Twig in the diff.)
- [x] I have updated from `main` and resolved any merge conflicts. (Via the base branch, which carries main.)
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges (see [docs/stacked-prs.md](../docs/stacked-prs.md)). **Base is `2123-update-hook-numbering-lint` (#2151).**
- [x] If I added a content entity bundle... (Not applicable. No new bundles or views; the one config change is a form widget setting.)


michael-wynne-wsu:

Link:
- https://pr2161-mm8xkggzjy19rnrtyt72l32v4ivcqjhd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2db6cf5ee2672d69a4c97

michael-wynne-wsu:

Link:
- https://pr2161-mm8xkggzjy19rnrtyt72l32v4ivcqjhd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2db6cf5ee2672d69a4c97

nick-deer (approved):

michael-wynne-wsu:

Link:
- https://pr2161-mm8xkggzjy19rnrtyt72l32v4ivcqjhd.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa2db6cf5ee2672d69a4c97

PR #2167, opened 2026-09-10
Fix the Dictionary Word upload locator so the test actually runs (#2155)

### LINKED ISSUE #2155: Dictionary Word Playwright test fails on every first attempt and is hidden by retries
## Summary

`Default Content: Dictionary Word` fails on its first attempt every time, at the
audio upload step. CI hides this because Playwright is configured with
`retries: 2` and reports the test as **flaky** rather than failed, which does
not fail the job.

The most recent green `playwright-tests` run
([34412110003](https://github.com/MukurtuCMS/Mukurtu-CMS/actions/runs/34412110003))
ended with:

```
1 flaky
  [default-content] › tests/default-content.spec.ts:317:5 › Default Content: Dictionary Word
8 passed (1.6m)
```

So this has been failing in every run that people have been reading as green.

## The failure

[`tests/playwright/tests/default-content.spec.ts:356`](../blob/main/tests/playwright/tests/default-content.spec.ts#L356)

```js
await modal.getByRole('textbox', { name: 'Add file' }).setInputFiles(uploadFilePath);
```

```
TimeoutError: locator.setInputFiles: Timeout 5000ms exceeded.
Call log:
  - waiting for getByRole('dialog', { name: 'Add or select media' })
                .getByRole('textbox', { name: 'Add file' })
```

The dialog opens correctly. The problem is the role. Playwright's own accessibility
snapshot of the open modal shows the file input exposed as a **button**, not a
textbox:

```yaml
- dialog "Add or select media":
    - generic: Add file
    - button "Add file" [active]
    - generic:
      - text: One file only.
      - text: 100 MB limit.
      - text: "Allowed types: mp3 wav aac m4a ogg."
```

`getByRole('textbox', ...)` therefore has nothing to match, and waits out the
5000ms `expect` timeout.

## Why a retry ever passes is unclear

This is the part worth investigating rather than assuming. If the role mapping
were simply wrong the test should fail identically on every attempt, yet the
retry passes often enough that CI has stayed green. Possibilities, none
confirmed:

- The retry takes a different path because media created by the first attempt
  already exists, so the upload branch behaves differently.
- The accessible role of the input differs depending on whether Claro has
  finished processing the widget, making it a race rather than a static
  mismatch.

Reproducing locally with `retries: 0` makes it fail deterministically, which is
the quickest way in:

```bash
cd tests/playwright
PLAYWRIGHT_BASE_URL=<a preview url> npx playwright test \
  tests/default-content.spec.ts --grep "Dictionary Word" --reporter=line
```

Note that `CI=1` turns retries back on, so leave it unset to see the real
first-attempt behaviour.

## Suggested direction

Target the input directly rather than by inferred role, which is both what
Playwright recommends for file inputs and immune to role mapping changes:

```js
await modal.locator('input[type="file"]').setInputFiles(uploadFilePath);
```

This is the only `setInputFiles` call site in the suite, so the change is
contained to that one line. Confirm against the accessibility snapshot before
and after rather than trusting a green run, since a green run here is exactly
what has been misleading.

## Worth deciding separately

Playwright can be configured to fail the run on flaky tests. Right now a test
that fails every single first attempt reports as success, which is how this went
unnoticed. That is a policy call for the team rather than part of this fix.

## How this surfaced

Found while fixing the unrelated `TUGBOAT_MAIN_URL` breakage in #2154. That job
had been failing on 100+ consecutive runs because the base preview URL 404d.
Once pointed at a live preview, 5 of the 6 `default-content` tests passed
immediately and this was the only genuine failure left.

michael-wynne-wsu:
Fixes #2155.

Test-only change, no update hooks and no shipped config, so it is safe to merge
either side of the 4.0.1 tag.

## The bug

The upload step in `Default Content: Dictionary Word` located the file input by
role:

```js
await modal.getByRole('textbox', { name: 'Add file' }).setInputFiles(uploadFilePath);
```

An `<input type="file">` has no textbox role. Claro renders it behind a styled
"Add file" button and the accessibility tree exposes it as a **button**, which
Playwright's own snapshot of the open modal confirms:

```yaml
- dialog "Add or select media":
    - generic: Add file
    - button "Add file" [active]
```

So the locator never matched and the step timed out on every single run.
`setInputFiles`' documentation recommends locating file inputs directly, which
is what this changes it to.

## Why nobody noticed

CI was hiding it rather than reporting it, in two layers.

`retries: 2` means the run reports **"1 flaky, 8 passed"** and exits 0, so the
job is green.

And the retry never rescues it. `testContentExists` is cached per worker; a
retry gets a **fresh** worker, re-checks `/communities`, finds the content the
earlier passing tests just created, and `test.skip()`s. So the upload assertion
has **never once executed in CI**.

## Verification

Run against a real site with retries off, so nothing could be masked:

| | Result |
| --- | --- |
| Before | 5 passed, 1 failed at the `setInputFiles` locator |
| After | **6 passed** |

## Two things worth knowing for anyone touching these tests

**They are order dependent and share state.** `Dictionary Word` needs the
protocols that `Community` creates. Running `--grep "Dictionary Word"` alone
fails on an unrelated Cultural Protocols checkbox rather than reproducing
anything, and the project has `fullyParallel: false` for that reason. Run the
whole project, against a site with no content, or `beforeEach` skips everything.

**A missing private file system looks like a test bug.** An earlier version of
this fix also raised the `waitForAjax` budget after the upload, because the AJAX
never settled. That turned out to be my sandbox missing `file_private_path`:
audio media uses the private scheme, so the upload failed with Drupal's generic
"Oops, something went wrong" and the request never completed. `mukurtu_core`'s
own `PrivateFileSystemRequirements` diagnosed it correctly ("Not configured",
error severity). Once configured, the default 5 second wait was sufficient, so
that change was dropped and this PR is the one-line locator fix only.

## Worth deciding separately

Playwright can be configured to fail a run on flaky tests. A test that fails
every first attempt currently reports success, which is exactly how this
survived. That is a team policy call, not part of this fix.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Test-only, none needed.)
- [x] I have run an accessibility check, and resolved any issues. (No UI change; the diff is a test locator.)


michael-wynne-wsu:

Link:
- https://pr2167-uguhw8agiittzcpqk1xnnpokahrged7u.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa311b20ed7640340762532

michael-wynne-wsu:

Link:
- https://pr2167-uguhw8agiittzcpqk1xnnpokahrged7u.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa311b20ed7640340762532

nick-deer (approved):

michael-wynne-wsu:

Link:
- https://pr2167-uguhw8agiittzcpqk1xnnpokahrged7u.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa311b20ed7640340762532

michael-wynne-wsu:

Link:
- https://pr2167-uguhw8agiittzcpqk1xnnpokahrged7u.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa311b20ed7640340762532

PR #2169, opened 2026-09-10
Let taxonomy imports map the UUID column exports produce (#2156)

### LINKED ISSUE #2156: Taxonomy term imports cannot map the UUID column that external exports emit

michael-wynne-wsu:
Fixes #2156. Safe to merge now that 4.0.1 is tagged.

## The bug

The external CSV exporters emit a `UUID` column for 45 bundles, because UUID is
the identifier that survives moving content between sites, which is what those
exporters are for. Every import strategy accepted a `UUID` source **except the
16 taxonomy term ones**.

So a taxonomy export taken from one Mukurtu site could not be mapped back in on
another. Terms were matched by name or created fresh rather than recognised as
the same term, and vocabularies silently diverged across sites. Both the export
and the import reported success.

## The fix

Adds the mapping to all 16 shipped strategies in `config/install`, which fixes
new sites, plus `mukurtu_import_update_40201()` for sites that already exist.

The hook only touches the 16 shipped ids. Discovering every taxonomy strategy
on the site would also reach ones a site authored itself, and editing somebody
else's config is a bigger liberty than fixing our own. It appends rather than
rewrites, skips any strategy that already has a `UUID` row, and matches on the
**source** string, since the source is the exporter's column header and so is
the half that has to line up.

## The exception list is gone

`ShippedExportImportContractTest` loses `KNOWN_UNMAPPABLE` entirely.

That list was added alongside the test that found this gap, deliberately shaped
so that closing the gap would force it to shrink: a second test failed if any
entry became mappable. It has now shrunk to nothing, so the contract is
enforced for **all 39 bundles with no taxonomy exemption at all**.

## Verification

Run as a real update: stripped the `UUID` row from all 16 strategies on a 4.0.1
site, ran `drush updb -y`, confirmed the hook restored exactly one row on each,
and that a second run is a clean no-op.

```
>  [notice] Update started: mukurtu_import_update_40201
>  [notice] Added the UUID column to 16 taxonomy import template(s), ...
>  [notice] Update completed: mukurtu_import_update_40201
```

3 kernel tests / 139 assertions, plus the contract test at 19 / 3971.

## A pre-existing bug this uncovered, not fixed here

The kernel test builds its strategy from the shipped YAML rather than calling
`installConfig('mukurtu_import')`, because that fails on an unrelated schema
gap:

```
Schema errors for views.view.mukurtu_import_results_content:
  display.default.display_options.fields.timestamp.link_to_node missing schema,
  display.default.display_options.fields.timestamp.comments missing schema
```

Setting `strictConfigSchema = FALSE` would have hidden that rather than fixed
it. It deserves its own issue.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (`mukurtu_import_update_40201`, above `last_removed` 40046, lint enforced.)
- [x] I have run an accessibility check, and resolved any issues. (No UI; config plus operator strings.)


michael-wynne-wsu:

Link:
- https://pr2169-6tmk6yklyvd6uvhmtg8ebm57tif1bdwu.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331688c321becccd890f2

michael-wynne-wsu:

Link:
- https://pr2169-6tmk6yklyvd6uvhmtg8ebm57tif1bdwu.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331688c321becccd890f2

michael-wynne-wsu:

Link:
- https://pr2169-6tmk6yklyvd6uvhmtg8ebm57tif1bdwu.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331688c321becccd890f2

nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

michael-wynne-wsu:

Link:
- https://pr2169-6tmk6yklyvd6uvhmtg8ebm57tif1bdwu.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331688c321becccd890f2

Issue #2164, opened 2026-09-10
Circle coverage shapes are unusable in the accessible lat/long editor, and edits are silently discarded


michael-wynne-wsu:
## Summary

A coverage shape drawn as a **circle** is unusable in the accessible lat/long
editor, in two compounding ways:

1. It presents as **128 vertex rows** rather than a centre and radius.
2. Any edit a keyboard user makes to those rows is **silently discarded**, because
   the shape is rebuilt from stale `circle_center`/`circle_radius` properties the
   next time the map widget loads it.

The second is the serious half. It is silent data loss for keyboard users only,
and it undercuts an ATAG 2.0 Part B claim: the accessible alternative editor
accepts input and then throws it away.

Pre-existing across all shipped coverage maps, unchanged by #2161 (which only
brought the submission form's config in line with the others).

## Verified mechanism

**Circles are stored as 128-gons carrying their circle metadata.**
[`modules/mukurtu_core/js/mukurtu-leaflet-widget.js:19-27`](../blob/main/modules/mukurtu_core/js/mukurtu-leaflet-widget.js#L19-L27)

```js
const polygon = L.PM.Utils.circleToPolygon(this, 128);
...
circle_center: [center.lat, center.lng],
circle_radius: this.getRadius()
```

**The accessible editor renders one row per vertex.**
[`GeofieldMukurtuLatLonWidget.php:241`](../blob/main/modules/mukurtu_core/src/Plugin/Field/FieldWidget/GeofieldMukurtuLatLonWidget.php#L241)
loops `for ($v = 0; $v < $vertex_count; $v++)` with `vertex_count` taken from the
stored shape, so a circle yields 128 latitude/longitude field pairs to tab
through, labelled as a generic polygon.

**Edits to those rows do not survive.** `massageFormValues()` rewrites the
geometry from the edited vertices but carries the original properties back
verbatim
([`:547-563`](../blob/main/modules/mukurtu_core/src/Plugin/Field/FieldWidget/GeofieldMukurtuLatLonWidget.php#L547-L563)):

```php
$feature = is_array($shape_row['source_feature'] ?? NULL) ? $shape_row['source_feature'] : [...];
$feature['geometry'] = ['type' => $shape_type, 'coordinates' => $coordinates];
$properties = is_array($feature['properties'] ?? NULL) ? $feature['properties'] : [];
...
$feature['properties'] = $properties ?: new \stdClass();
```

[...]

PR #2170, opened 2026-09-10
Match the submission coverage map to the other content types (#2153)

### LINKED ISSUE #2153: Add an update hook to align the submission coverage map with other geofield fields

michael-wynne-wsu:
Fixes #2153. Safe to merge now that 4.0.1 is tagged.

## The bug

The submission form's coverage map had drifted from the six other
`geofield_mukurtu` displays: no circle drawing tool, centred on 0,0 in the Gulf
of Guinea rather than the shared 20,10 world view, and missing
`singlePointZoom`.

The update hooks that set those values only ever touched the bundles that
existed when each was written, and every test named the same six, so nothing
ever looked at the seventh. It surfaced in #2123 only because the replacement
test discovers displays by scanning for the widget instead of listing them.

4.0.1 corrected the shipped config, which fixed new installs. This is the other
half.

## What the guard actually protects, and what it does not

Each value is only changed where it still holds the old default. I want to be
straight about the limit of that rather than oversell it.

| Case | Behaviour |
| --- | --- |
| Centre moved to something like 45,-100 | **left alone** - obviously chosen, covered by a test |
| Centre still 0,0 | set to 20,10 - indistinguishable from the drift |
| `drawCircle: false` | set to `true` - `false` is both the old default *and* the only way to say "off" |
| `singlePointZoom` absent | added |
| Widget swapped away from `geofield_mukurtu` | **skipped entirely** - a deliberate choice about the field |

For the two ambiguous cases it converges on the documented default and tells
the operator what changed, the same reasoning as #2160.

## Verification

Run as a real update: reverted a 4.0.1 site's display to the drifted state, ran
`drush updb -y`, confirmed all four values landed, and that a second run is a
no-op.

```
before: lat=0.0 lon=0.0 spz=NULL circle=false
after:  lat=20.0 lon=10.0 spz=12 circle=true
```

5 kernel tests / 52 assertions.

## Two things the test needed that were not obvious

The submission **form mode** has to exist as an entity or saving the display
fails with `getConfigDependencyName() on null`.

And the negative case compares against what was actually **stored** rather than
what was passed in, because the widget plugin normalises settings to its own
schema on save and the lat/long widget has no `toolbar` key at all. Asserting
on the input would have failed for the wrong reason.

## Worth knowing when reviewing

`SubmissionFormDisplayManager::SUBMISSION_WIDGET_OVERRIDES` swaps geofield to
the accessible `geofield_mukurtu_latlon` widget when a submission form is
enabled through the UI, so on many sites this display never renders the map at
all and the change is config consistency rather than a visible one. See #2164
for the accessibility side of circles.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (`mukurtu_submissions_update_40201`, above `last_removed` 40014, lint enforced.)
- [x] I have run an accessibility check, and resolved any issues. (Done in #2161 for this same widget change; no new UI here.)


michael-wynne-wsu:

Link:
- https://pr2170-fyxukjxc4khtvt7vlzxtbz4p2y2wczki.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331800ed7640340791ad4

michael-wynne-wsu:

Link:
- https://pr2170-fyxukjxc4khtvt7vlzxtbz4p2y2wczki.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331800ed7640340791ad4

michael-wynne-wsu:

Link:
- https://pr2170-fyxukjxc4khtvt7vlzxtbz4p2y2wczki.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331800ed7640340791ad4

michael-wynne-wsu:

Link:
- https://pr2170-fyxukjxc4khtvt7vlzxtbz4p2y2wczki.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa331800ed7640340791ad4

nick-deer (approved):
Code review checklist:
- [x] Clean rebuild

Issue #2187, opened 2026-09-11
[WCAG 2.4.7] Primary navigation links have no visible focus indicator


michael-wynne-wsu:
Found while implementing #1038. Pre-existing on `main`, unrelated to that work.

## Problem

Tabbing to a link in the main navigation produces no visible focus indicator.

Measured in Chromium on a stock install with no header image, tabbing with real key presses so `:focus-visible` applies:

```
text         : "Communities"
outline      : none 0px rgb(70, 70, 70)
border-bottom: 4px rgba(0, 0, 0, 0)
```

`themes/mukurtu_v4/components/02-molecules/menus/_nav-primary.scss:67` declares `outline: auto 2px var(--brand-secondary-dark)`, but the computed outline is `none` at `0px`, so that rule is not reaching the focused element. The 4px bottom border that serves as the hover underline stays fully transparent on focus.

The result is that a keyboard user tabbing through the header has no indication of where they are.

## Steps to reproduce

1. Load any page as an anonymous user.
2. Press Tab until focus reaches "Communities" in the main navigation.
3. Nothing changes visually.

## Relevant

- WCAG 2.1 AA 2.4.7 Focus Visible
- WCAG 2.2 AA 2.4.11 Focus Not Obscured and 2.4.13 Focus Appearance are worth checking in the same pass
- Related in kind to #1999, though that covers dictionary word page elements rather than the main nav
- The colour the outline *would* use comes from `--brand-secondary-dark`, which #1038 now retunes inside the header when a background image is set, so a fix will resolve correctly there too

michael-wynne-wsu:
Re-tested this on a live 4.0 sandbox with real Tab keypresses (so `:focus-visible` applies) at both viewports, on both built-in palettes. **A visible focus indicator is present in both cases, and both pass 1.4.11.** Closing as not reproducible — details below in case the methodology is useful elsewhere.

## Wide (>= 960px)

The computed `outline` really is `none 0px`, as reported. But that's deliberate, not a missing rule: `themes/mukurtu_v4/components/02-molecules/menus/_nav-primary-wide.scss:25-47` zeroes the native outline inside `body:not(.is-always-mobile-nav)` at `>=lg` and replaces it with a `::before` pill:

```scss
& .primary-nav__menu-link {
  &:focus {
    position: relative;
    outline: 0;

    &::before {
      /* ...absolutely positioned, centred on the link... */
      border: solid 2px var(--brand-primary);
[...]

Issue #2186, opened 2026-09-11
--color-text-neutral-loud is read but never defined


michael-wynne-wsu:
Found while implementing #1038.

`themes/mukurtu_v4/components/02-molecules/menus/_nav-primary.scss:38` sets:

```scss
color: var(--color-text-neutral-loud);
```

That custom property is **not defined anywhere in the theme**. `grep -rn -- "--color-text-neutral-loud:" themes/mukurtu_v4/components/` returns nothing.

A `var()` with no fallback and no definition makes the declaration invalid at computed-value time, so the property falls back to the inherited value. The primary nav links therefore take their colour from `body` (`--brand-text-color`, `#464646`) rather than from the rule that appears to set it.

## Why it matters

It looks right today purely by coincidence: the inherited colour happens to be the one intended. But the declaration is inert, so:

- Anyone editing that line to change nav link colour will find it has no effect.
- Any ancestor that sets `color` silently repaints the nav. This is exactly how #1038 broke: a colour set on `.site-header` reached straight through to the menu links and, on the light treatment, rendered them white on the nav's white bar. Confirmed live before the fix.

## Suggested fix

Point it at the real token:

```scss
color: var(--brand-text-color);
```

Worth a quick grep for other undefined custom properties in the theme at the same time, since nothing currently catches this class of error. Note stylelint cannot run against the theme at present, see the stylelint peer-dependency issue.

Issue #2181, opened 2026-09-11
Custom palette CSS is not regenerated on config import


michael-wynne-wsu:
Found while auditing #1009.

## Problem

`public://mukurtu-design/custom-palette.css` is written in exactly two places:

1. `MukurtuDesignSettingsForm::submitForm()`, when the palette is `custom`
2. `DesignPalette::alterThemeLibraries()`, but only when the file is **missing**

`modules/mukurtu_design` has no config event subscriber. So if `mukurtu_design.settings:colors` changes by any route other than that form - `drush config:import`, a deployment, `drush config:set`, a recipe - the file already exists and is never rewritten. The site keeps serving the previous palette's colours, with nothing to indicate the config and the generated CSS disagree.

## Suggested fix

Subscribe to `ConfigEvents::SAVE` (or config import completion) for `mukurtu_design.settings` and regenerate. A cheaper alternative is to stop caching to disk and attach the six custom properties inline, which removes the whole class of staleness.

Issue #2175, opened 2026-09-11
[WCAG 1.4.3] Hero blocks: a light/dark text switch cannot meet contrast over a background image


michael-wynne-wsu:
Found while auditing #1009. Pre-existing, and it affects the hero block that ships on every site's homepage, not only the ones newly creatable via #2174.

## Problem

`full_image_with_description` renders its title and body directly over the uploaded image, with no scrim. The only contrast control is `field_text_color`, a binary Light/Dark choice. `themes/mukurtu_v4/components/02-molecules/image-w-description/_image-w-description.scss` adds `text-shadow: 1px 1px 0 rgb(... / 80%)`, which is not a recognised contrast technique.

Measured against a realistic mid-tone photographic gradient, neither setting passes across the whole block:

| Region | Light (white) | Dark (`#9a1134`) |
|---|---|---|
| lighter third | 2.29 - 2.57:1 | 3.27 - 3.66:1 |
| middle | 2.98 - 3.38:1 | 2.48 - 2.82:1 |
| darker third | 4.76 - 7.98:1 | 1.05 - 1.76:1 |

The Light variant also sets body copy to `var(--font-size-l)` = 20px at weight 400. That is **not** WCAG "large text" (which needs 24px, or 18.66px bold), so it needs the full 4.5:1, which is unreachable over most photographs.

The field's own help text tells authors to verify 4.5:1, but gives them no control capable of achieving it. That makes it an ATAG 2.0 B.2.2 problem as well: the tool asks the author to guarantee something it does not let them guarantee.

## Suggested fix

Add a semi-opaque scrim between the image and the text, keyed off the existing choice, e.g. `linear-gradient(rgb(0 0 0 / 55%), rgb(0 0 0 / 55%))` over the image when `text-light` and the inverse for `text-dark`. That makes the author's choice deterministic rather than a guess, and would also resolve the full variant's half of #2178.

Needs a design decision, since it changes the look of existing sites' heroes.

PR #2189, opened 2026-09-11
Add home page and header background images

### LINKED ISSUE #1034: Full-bleed homepage background
Based on the [PPWP](https://plateauportal.libraries.wsu.edu/), we want the option to enable a full background image on the homepage/landing page.

This would most likely be in place of the regular hero image options.

Ideally something users can upload to the site, instead of manually editing code/files.

Image

michael-wynne-wsu:
Closes #1034 and #1038.

## What changed and why the branch was reworked

This started as a header background for #1038. Reviewing #1034 showed that was the wrong shape.

The live Plateau Peoples' Web Portal puts **one image on a page-level wrapper**, not on a block and not on the header:

| | |
|---|---|
| Element | `div.main-container-wrapper`, `background-size: cover` |
| Size | 1440×1320 on a 3511px page |
| Contains | the header, the title and the browse grid |
| Interior pages | `background-image: none` |

So #1034 was closed on the strength of the `full_image_with_description` block, which is a block background, not a page one. And #1038's header-only scoping could not reach the hero or the grid below it.

The branch now puts the image on `.layout-container`, **on the home page only**.

## Two images, because PPWP uses two

I twice claimed PPWP's interior pages carry no background image, having only checked `.main-container-wrapper`. A full scan finds it on a different element. The site uses two:

| | Home page | Interior pages |
|---|---|---|
| Element | `div.main-container-wrapper` | `div#top-wrapper` |
| Image | `bg.jpg` | `bg_interior.jpg` |
| Box | 1440×1320 | 1440×236 |

That is exactly what #1038 meant by "will likely need to be a separate image from homepage, because of dimensioning". An earlier revision of this branch was home-page-only, partly on my incorrect reading.

So there are two settings:

- **Page background** covers the home page.
- **Header image** covers the logo and menu on every page — except where the home page has its own background, since that already covers the header and a second image would stack on top of it.

A site can use one tall image on the home page and a short one everywhere else, like PPWP, or a single header image everywhere by leaving the page background empty. Each has its own light/dark treatment, since the two images differ in tone and crop.

## How it works

`mukurtu_design` gains three settings: an image, a light/dark text treatment, and nothing else — the front page opt-out was removed once the background became home-page only.

The URL is set as a custom property inline on the page container, not written to a generated stylesheet. The existing generated file is attached only when the palette is Custom, so it could not have carried this on the other palettes, and a second generated file would inherit the staleness bug filed as #2181.

## Accessibility

Contrast is handled with a scrim, deliberately the approach proposed for #2175 rather than the bare light/dark switch that issue documents as unable to meet 1.4.3 over a photograph. Measured against a full white-to-black gradient: **5.74:1 light, 7.37:1 dark**. Light has the least room — below about 54% the white text drops under 4.5:1.

Putting an image behind everything invalidates text colours designed for a white ground, so the treatment now reaches headings and body copy via `inherit`. That is self-correcting rather than a flat override: inside the footer it picks up `--footer-text-color`, inside a block with its own ground it picks up that block's colour, and only text actually on the image takes the treatment. A flat override is what made the main menu render white-on-white twice during development.

**Front page: 10 text items, 0 failing, worst case 5.65:1. Interior header: 5.74:1 light.** Before the recolouring, block titles measured 1.65:1.

Two defects only visible on screen, not in the numbers:

- The horizontal image-with-description block paints a tan ground but never set a text colour, so `inherit` walked past it and rendered white on tan at **1.48:1**. Fixed by giving that component its own colour — a no-op without a background.
- `.layout--onecol .block-wrapper` carries a pale drop shadow to separate blocks on white. Over a photograph it read as a glowing bar across the full width. The override needs the layout class in its selector: the original is also two classes and `04-templates` is imported after `00-base`, so an equally specific rule silently lost.

## Filed rather than fixed here

- #2187 — primary nav links have no visible focus indicator, with or without a background. The retuned `--brand-secondary-dark` means it will resolve correctly on the scrim once fixed.
- #2186 — `--color-text-neutral-loud` is read but never defined, which is why a colour set on an ancestor reached the menu links at all.

## Update hook

`mukurtu_design_update_40005()` seeds the keys with no image set, so nothing changes visually until an admin uploads one. It fills in only absent keys: the shipped default for `image` is NULL, so a `get()`-based check would read a configured site as unconfigured and reset it. `update_last_removed` is 40004 and nothing else in the module uses 40005.

## Testing

23 kernel tests. Verified in a DDEV site: the property renders on the home page, interior pages carry neither the class nor an image, the mobile drawer keeps its own white ground with dark links, and the hook is idempotent.

Guards confirmed by mutation: removing the precedence rule, the front-page check or the dangling-file check each fails exactly the intended test and nothing else.

One test exists because of a bug shipped mid-branch: `create()` type-hinted a `FileUsageInterface` that does not exist, so `/admin/config/color-settings` fataled for everyone. Every test drove the service directly and nothing instantiated the form, so the route had no coverage at all.

## Two things to know

**The image help text** now reads "behind your home page" rather than "behind the whole page" — a factual correction the scope change forced, not separately approved wording.

**The compiled CSS** is built with sass 1.86.0 to match what is committed on `main`, so the diff stays reviewable. #2190 fixes that mismatch properly; this will want a recompile once it lands.

## Pre-merge checklist:

- [x] If I added a content entity bundle... (None.)


michael-wynne-wsu:

Link:
- https://pr2189-yzakhegg4ajvqbmgur13vivz0fvmt0ly.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46d4e8c321becccf51408

michael-wynne-wsu:

Link:
- https://pr2189-yzakhegg4ajvqbmgur13vivz0fvmt0ly.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46d4e8c321becccf51408

michael-wynne-wsu:

Link:
- https://pr2189-yzakhegg4ajvqbmgur13vivz0fvmt0ly.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46d4e8c321becccf51408

michael-wynne-wsu:

Link:
- https://pr2189-yzakhegg4ajvqbmgur13vivz0fvmt0ly.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46d4e8c321becccf51408

michael-wynne-wsu:

Link:
- https://pr2189-yzakhegg4ajvqbmgur13vivz0fvmt0ly.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46d4e8c321becccf51408

michael-wynne-wsu:

Link:
- https://pr2189-yzakhegg4ajvqbmgur13vivz0fvmt0ly.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46d4e8c321becccf51408

PR #2188, opened 2026-09-11
Fix the theme's npm install, broken by the stylelint config bump

michael-wynne-wsu:
`npm install` in `themes/mukurtu_v4` fails outright on a clean checkout:

```
npm error ERESOLVE could not resolve
npm error While resolving: stylelint-config-standard-scss@17.0.0
npm error Found: stylelint@16.16.0
npm error Could not resolve dependency:
npm error peer stylelint@"^17.0.0" from stylelint-config-standard-scss@17.0.0
```

#2144 bumped `stylelint-config-standard-scss` to `^17`, which requires `stylelint ^17`, while `stylelint` stayed at `^16.16.0`.

Found while doing theme work for #1038, where it blocked recompiling SCSS at all without `--legacy-peer-deps`.

## Why the fix goes down, not up

Bumping stylelint to 17 looks like the obvious move, but the rest of the toolchain is not ready:

| Package | Requires |
|---|---|
| `stylelint-use-logical-spec` | `>=11 < 17` — **blocks 17 outright** |
| `stylelint-config-standard@37` | `^16.13.0` |
| `stylelint-scss` | `^16.8.2 \|\| ^17.0.0` |
| `stylelint-config-standard-scss@17` | `^17.0.0` |

So `stylelint-config-standard-scss` goes back to `^16`, and stylelint moves within 16.x to 16.26.1, which is what `stylelint-config-standard-scss@16` itself requires. Everything resolves without `--legacy-peer-deps`.

Moving the whole toolchain to stylelint 17 is worth doing later, but it needs `stylelint-use-logical-spec` to support it first, and that is a bigger change than unbreaking the install.

## The linter has not been running

This is not only an install-time annoyance. With the mismatched config in place, `npx stylelint` failed with eight `Unknown rule` errors — `layer-name-pattern`, `media-type-no-deprecated`, `nesting-selector-no-missing-scoping-root`, `no-invalid-position-declaration`, `property-no-deprecated` and others — because the v17 config references rules the installed stylelint 16 does not have. The theme has effectively had no working linter since #2144 landed.

With this fix, stylelint runs and reports **242 pre-existing problems** across the theme, 82 of them auto-fixable. **None are touched here** — that deserves its own pass.

## Scope

`package.json` and `package-lock.json` only. No SCSS, no compiled CSS.

[...]

PR #2190, opened 2026-09-11
Rebuild style.css so it matches the declared sass version

michael-wynne-wsu:
Housekeeping. No SCSS, no behaviour change, no new feature.

## The problem

`themes/mukurtu_v4/css/style.css` was compiled with sass **1.86.0**. #2147 then bumped the dependency to `^1.104.0` without recompiling, and #2148 bumped `gulp-sourcemaps`. So the committed artifact and the toolchain declared to produce it have disagreed ever since.

That cost is paid by whoever touches the theme next. Running `gulp buildSass` emits ~165 lines of diff that belong to neither their change nor the previous one. On #2189 it buried roughly 12 lines of real work in 261 lines of churn, and I ended up installing sass 1.86.0 by hand to keep that PR reviewable. That workaround shouldn't be the price of doing theme work.

## What changed

Nothing hand-edited. This is exactly what `npx gulp buildSass` emits under the declared dependencies.

Two formatting differences from 1.104:

- strings are quoted with single quotes rather than double
- nested at-rules are emitted in source order instead of hoisted after the parent rule's remaining declarations

## Verification

Because the second one reorders output, and CSS is order-dependent, "it's only formatting" needed proving rather than asserting. Three checks:

**1. Declaration inventory.** Parsed both files with postcss and compared every `(selector chain, property, value)`:

```
declarations before: 4747   after: 4747
REMOVED (0):
ADDED (0):
```

At-rule nesting is part of the chain, so a declaration moving into or out of an `@media` / `@supports` block would have shown here.

**2. Cascade winners.** The inventory being equal doesn't prove the cascade is, since order decides ties. So for every `(selector, property, condition-set)` I compared the last-wins value, and separately checked whether any conditional block flipped position relative to a plain rule setting the same property:

```
No declaration winner changed in any selector/property/condition.
No conditional/plain ordering flipped for a shared property.
```

**3. Computed styles in a browser.** Static analysis still can't rule out equal-specificity cross-selector flips, so the end-to-end check: load the site with each stylesheet and compare computed styles for every element.

| | |
|---|---|
| Pages | `/`, `/user/login`, `/communities`, `/collections`, `/digital-heritage` |
| Viewports | 1440px and 390px |
| Elements compared | 1040 |
| Properties compared | 33,280 |
| **Differences** | **0** |

(The first pass reported 10 differences; all were Views' random `js-view-dom-id-*` hashes regenerating per request, not styles.)

## Related

- #2188 fixes `npm install` in this theme, which currently fails outright. Independent of this PR, but you'll need it to rebuild the CSS yourself.
- #2189 carries the hand-built workaround this PR makes unnecessary for future branches.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (None needed.)
- [x] I have run an accessibility check, and resolved any issues. (Zero computed-style changes, so no rendering to audit.)
- [x] I have recompiled SCSS if required. (That is the entire PR.)
- [x] I have updated or created CI/CD tests, if required. (None applicable.)
- [x] If I added a content entity bundle... (None.)


michael-wynne-wsu:

Link:
- https://pr2190-ca7isaxvodzjdxjcncstqsvq5avqjvoz.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46efb8c321becccf53748

michael-wynne-wsu:

Link:
- https://pr2190-ca7isaxvodzjdxjcncstqsvq5avqjvoz.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa46efb8c321becccf53748

Issue #2179, opened 2026-09-11
Custom colour palette only controls 6 of the 16 palette variables


michael-wynne-wsu:
Found while auditing #1009. Related to #1054.

## Problem

`DesignPalette::CSS_VAR_MAPPING` maps six config keys to CSS custom properties, and `generateCustomCss()` writes only those six into `public://mukurtu-design/custom-palette.css`.

A shipped palette defines 16. The 10 a custom palette cannot reach:

```
--mukurtu-breadcrumb-gradient-color
--mukurtu-collection-pane-bg-color
--mukurtu-collection-title-font-color
--mukurtu-dictionary-word-label-background-color
--mukurtu-mejs-background-color
--mukurtu-mejs-time-rail-current-color
--mukurtu-mejs-time-rail-loaded-color
--mukurtu-navbar-collapse-link-hover-color
--mukurtu-panel-pane-content-background-color
--mukurtu-panel-pane-text-color
```

The `custom` library loads **only** the generated file, with no base palette underneath, so those 10 fall through to `style.css`. And `themes/mukurtu_v4/components/style.scss` imports `red-bone` then `yellow-blue`, so yellow-blue's values win on equal specificity.

Net effect: an admin picks six brand colours and still gets yellow-blue's hardcoded golds for the breadcrumb gradient, navbar hover, media player rails, dictionary word labels, collection panes and panel panes.

Worth noting `red-bone.css` already defines several of these as `var(--brand-secondary)` - i.e. brand-relative rather than hardcoded. If the custom palette emitted the same brand-relative definitions, most of the 10 would track the admin's choice for free.

## Reproduce

1. `/admin/config/color-settings`, choose Custom, set six clearly distinct colours, save.
2. Inspect `:root` - only the six brand properties come from `custom-palette.css`.
3. `--mukurtu-navbar-collapse-link-hover-color` is still `#e5ab48`.

Issue #2177, opened 2026-09-11
[ATAG B.2.2] "Display title" is an inert control on image-with-description blocks


michael-wynne-wsu:
Found while auditing #1009. Distinct from #1826, which is about the toggle's greyscale appearance; this is about the control doing nothing at all.

## Problem

The Configure block dialog for the three image-with-description bundles exposes a required **Title** field and a **Display title** checkbox. The templates never print `label` or `label_display` - they always render `block_entity.info.value` as a heading regardless.

Verified live: unchecking "Display title" still emits `<h2>A11y Audit Second Hero</h2>` in the Layout Builder preview.

So an author is handed a control that has no effect, and there is no way to place one of these blocks without a heading.

## Why it matters

ATAG 2.0 B.2.2 expects authoring controls to actually do what they say. A no-op control is worse than an absent one: it tells the author the page is in a state it is not in.

## Suggested fix

Either honour `label_display` in the three templates, or remove the checkbox from the form display for these bundles. Probably wants doing alongside the hardcoded heading levels (#2176).

Issue #2176, opened 2026-09-11
[WCAG 1.3.1] Image-with-description blocks hardcode h1 on the front page


michael-wynne-wsu:
Found while auditing #1009.

## Problem

All three templates hardcode the heading level from the front-page flag:

```twig
{% set title_tag = is_front ? 'h1' : 'h2' %}
```

- `block--block-content--type--image-with-description.html.twig`
- `block--block-content--type--vertical-image-with-description.html.twig`
- `block--block-content--type--full-image-with-description.html.twig`

Two consequences:

1. **Multiple `<h1>`.** #2174 makes these bundles creatable as inline blocks, so an author can now add several to the homepage, each emitting its own `<h1>`.
2. **No `<h1>` at all.** The landing page does not render the node title, so one of these blocks is the homepage's only `<h1>`. Remove it and the page has none.

Neither state is under the author's control, and there is no way to choose a heading level.

## Suggested fix

Tied to the inert "Display title" control (#2177); they are probably one piece of work. At minimum the heading level should not be derived from `is_front`, and the landing page needs a reliable `<h1>` that does not depend on which blocks an author happens to place.

PR #2174, opened 2026-09-11
Fix which blocks landing pages offer in Layout Builder

michael-wynne-wsu:
Refs #1009.

The `landing_page` default view display restricts Layout Builder through `layout_builder_restrictions`, and two of those restrictions were wrong. Found while auditing #1009 to decide which Plateau Peoples' Web Portal theme features belong in stock Mukurtu.

## What was broken

**The hero blocks could not be created.** The `Inline blocks` allowlist omitted all three image-with-description bundles. `DefaultLandingPage` pre-creates them as reusable content blocks, so a site builder could place *those*, which is why nobody noticed, but could never build a new hero inline.

**Browse by Community could not be re-added.** `EntityViewModeRestriction::alterBlockDefinitions()` enforces `allowed_block_categories` as a fallback: when that list is non-empty, any category absent from it is stripped from the Layout UI. `Mukurtu` was absent, so the block vanished from the chooser. `DefaultLandingPage` writes it onto the default homepage directly, bypassing restrictions, so it shipped on every site but could not be put back once removed.

## What this changes

- Adds the three image-with-description bundles to the `Inline blocks` allowlist.
- Allowlists `mukurtu_browse_by_community` under a new `Mukurtu` key, rather than opening the whole category. The other Mukurtu-category blocks are a node-context sidebar list, a search-page toggle and the dashboard setup checklist, none of which belong on a landing page.
- Drops `views_block:browse_by_community-community_browse_block` from the allowlist. Allowlisting the block plugin put two identically labelled "Browse by Community" entries in the Add block list, and the two are **not** equivalent: the view filters on published status and langcode only, with no visibility filter, so it lists communities whose sharing setting is "Community only" to everyone. That is the bug fixed for the block plugin in #1005.
- Defaults `field_text_color` on `full_image_with_description` to `dark`. See below.

## Accessibility

Making the hero bundles creatable inline opened an authoring path that produced inaccessible output by default, which is an ATAG 2.0 Part B problem rather than a theme bug: `field_text_color` shipped a default of `light` (white text) while `field_image` is optional, so the minimal authoring action (add block, type a title, save) rendered white text on the white page background at 1:1 contrast. The default is now `dark`.

`mukurtu_update_40030()` carries that to existing sites, but only rewrites the value this profile shipped, and never touches existing blocks: their stored value is the author's choice.

### Deliberately not fixed here

These came out of the accessibility audit, are pre-existing, and already affect the hero block that was placeable before this PR. Filing separately rather than growing this PR:

[...]

PR #2191, opened 2026-09-11
Make the text-only homepage hero work properly

### LINKED ISSUE #1035: Text-only homepage title

michael-wynne-wsu:
Closes #1035.

## The capability already existed

`field_image` on the full-background hero is `required: false`, and the template renders the block title then the body regardless. So leaving the image empty already produced a centred title with welcome text and a link — PPWP's homepage treatment, with no code.

Nobody had noticed, plausibly because on `main` that configuration renders **white text on the white page**: `field_text_color` defaults to `light`. #2174 changes that default to `dark`, which is what made this visible at all.

So this PR is not a new feature. It is the three things that were wrong with a case the block already supported.

## 1. The colour control does not fit a text-only hero

`field_text_color` asks the author to pick text colour against a background that is not there. "Light" is white on white, 1:1. "Dark" paints *everything* brand red, body copy included, through a `* { color }` rule.

`field_hero_background` picks a ground instead, and the text colour is set alongside it rather than chosen separately — the same principle as the header scrim in #2189, and deliberately not a second light/dark toggle.

| Preset | Ground | Text |
|---|---|---|
| None | page background | body `#464646`, heading `--brand-primary-dark` |
| Brand | `--brand-primary-dark` | white |
| Soft | `--brand-secondary` | `--brand-text-color` |

Measured against all three shipped palettes, worst case per preset:

| Palette | None | Brand | Soft |
|---|---|---|---|
| Red and bone | 8.39:1 | 8.39:1 | 6.36:1 |
| Blue and gold | 5.00:1 | 5.00:1 | 4.62:1 |
| Yellow and blue | 5.00:1 | 5.00:1 | 4.62:1 |

All clear 4.5:1. Two findings from measuring rather than assuming:

- **Brand uses `--brand-primary-dark`, not `--brand-primary`.** The lighter primary with white text fails on Blue and gold at 4.00:1.
- **Soft cannot use a brand text colour.** `--brand-secondary` with `--brand-primary-dark` is 2.45:1 on Blue and gold. It needs the neutral reading colour.

A **custom** palette is six author-chosen colours, so none of this is guaranteed there — that is #2179 and #1054.

Verified live on the rendered page, nine measurements across the three presets, all passing.

## 2. It reserved 640px for 150px of content

The 600px `min-block-size` gives text room over a full-bleed photograph. With no image it is dead space, so it is dropped for that case. The block goes from 640px to 260px.

## 3. The heading and body sat side by side

The layout is a wrapping row that relies on the image's height to push the heading and body onto separate lines. Without an image they shared one. The no-image case now states the stacking explicitly.

**The with-image hero is untouched** — verified: still 640px, 600px min-height, row/wrap, `place-items: center`. All the new rules are scoped to a `--no-image` class the template only adds when the rendered image field is empty.

## Out of scope, already filed

- #2176 — the `h1` is hardcoded via `is_front`, so several heroes on the homepage means several `<h1>`
- #2177 — "Display title" cannot turn the heading off on these bundles
- #2175 — the with-image contrast handling, which this deliberately does not change

## A bug worth reading if you write update hooks

`mukurtu_update_40031()` initially fataled with:

```
The configuration property settings.allowed_values.0.label.0 doesn't exist.
```

`config/install` states `allowed_values` in the documented `[{value, label}]` form, but `FieldStorageConfig` stores the older value-keyed map, and saving an entity casts against the stored shape. Installing config never hits this because the installer writes trusted data and skips the cast — **which is exactly why a fresh install passed while the update hook failed**.

I first attributed this to a kernel-test quirk. It was not; it failed identically on a real site. The hook now converts just that key, leaving the shipped YAML canonical, and the kernel test that reproduces the original failure passes.

## Testing

20 kernel tests in the profile suite. Beyond the field definition and the update hook, one test asserts every allowed value has a matching `.hero-ground-*` rule in the compiled CSS, since the field's values and the theme are edited in different files and nothing else connects them.

Verified on a real site: the hook runs, reports correctly, is idempotent, and leaves the site identical to a fresh install.

The one `phpcs` finding is a deliberate `strictConfigSchema = FALSE`, justified in the docblock: the display-wiring test writes entity displays without the field instances they normally reference, and installing the full bundle would mean asserting against a fixture rather than against the hook.

## Note on the compiled CSS

Built with sass 1.86.0 to match what is committed on `main`, so the diff stays reviewable. See #2190, which fixes that mismatch properly.

## Pre-merge checklist:

- [x] If I added a content entity bundle... (No new bundles or views; a field added to an existing bundle.)


michael-wynne-wsu:

Link:
- https://pr2191-gpf7yphyxheearaxl4jmpfl737zjw7qx.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa4797cf5ee2672d68aecc0

michael-wynne-wsu:

Link:
- https://pr2191-gpf7yphyxheearaxl4jmpfl737zjw7qx.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa4797cf5ee2672d68aecc0

michael-wynne-wsu:

Link:
- https://pr2191-gpf7yphyxheearaxl4jmpfl737zjw7qx.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa4797cf5ee2672d68aecc0

Issue #2199, opened 2026-09-13
[WCAG 1.4.11] Focus rings are too faint to see on light backgrounds


michael-wynne-wsu:
## Summary

`--focus-color` is `#66afe9`, a light blue. It is the outline colour for every link, button, input and select in `mukurtu_v4`, and it fails WCAG 2.1 AA **1.4.11 Non-text Contrast** (3:1) against every light surface in the theme.

The indicator is present and correctly sized. It is just too faint to see on the backgrounds the theme actually uses. That makes it distinct from #2187 and #1999, which are about focus indicators that are missing entirely.

## Measured

Chromium at 1280px against a 4.0 site, walking up from each focusable element to the first opaque background, then computing the ratio against the computed `outline-color`.

| Page | Element | Background | Ratio | 3:1 |
|---|---|---|---|---|
| `/digital-heritage` | facet link | bone `#d9d4c2` | **1.59:1** | fail |
| `/digital-heritage` | `input.form-text` | form grey `#e5e9e9` | **1.93:1** | fail |
| `/digital-heritage` | `select.form-select` | form grey `#e5e9e9` | **1.93:1** | fail |
| `/digital-heritage` | `.search-collapse-toggle__input` | `#f1efe9` | **2.06:1** | fail |
| `/digital-heritage` | button | white | **2.37:1** | fail |
| `/` | Klaro toggle | brand red `#9a1134` | 3.54:1 | pass |
| `/` | footer link | grey `#464646` | 3.99:1 | pass |

It only passes where it sits on a dark surface.

## Why this is not "pick a better blue"

A focus ring in this theme has to work on white, on the bone and cream panel fills, **and** on brand red. No single flat colour does that:

| Candidate | white | `#f1efe9` | `#e5e9e9` | `#d9d4c2` | `#464646` | `#9a1134` | worst |
|---|---|---|---|---|---|---|---|
| `#66afe9` (current) | 2.37 | 2.06 | 1.93 | 1.59 | 3.99 | 3.54 | **1.59** |
| `#1a5fb4` | 6.29 | 5.47 | 5.14 | 4.24 | 1.50 | 1.33 | **1.33** |
| `#005a9c` | 7.14 | 6.21 | 5.83 | 4.81 | 1.32 | 1.18 | **1.18** |
| `#000000` | 21.00 | 18.26 | 17.16 | 14.15 | 2.22 | 2.50 | **2.22** |

[...]

Issue #1009, opened 2025-07-24
Plateau Peoples' Web Portal theme


michael-wynne-wsu:
- We need to add theming support that will bring us closer to re-implement the theme present on https://plateauportal.libraries.wsu.edu/home (v3 site)
- In v3 it's a lot of work to maintain, we hope that v4 will be easier to adapt.
- We should determine which of this we incorporate into v4, so all users can apply it, and what needs to be done as a custom theme.
- The site has no custom development beyond that noted below (I think).

- [ ] Landing page/home
   - [x] #1034
   - [ ] #1035
   - [ ] Browse by tribal path 
      - This is the "browse by community" block. 
      - It has no special functionality, we just need to rename it.
      - I don't believe we need the custom styles.
   - [ ] #1036 
- [ ] #133
   - Site has a custom palette. 
   - If a general solution (above) is implemented, we don't need custom development. If not, we will.
- [ ] Custom footer
   - I believe this is just custom HTML and bootstrap - custom code is not required.
- [ ] Internal pages
   - [ ] #1037
   - [ ] #1038
- [ ] /communities is lightly styled
   - I don't think we need the custom styles.
- [ ] /categories-browse is lightly styled
   - I don't think we need the custom styles.
- [ ] /curriculum is a custom styled page
   - It is a sub-set of collections.
   - This may need to be replicated. 
- [x] #124

michael-wynne-wsu:
@Javi-er I added sub-issues and detail. I realize there is likely too much here, so let me know if you want to discuss it before deciding where to start (after the other open issues).

Issue #1008, opened 2025-07-24
Landing page and blocks


michael-wynne-wsu:
These blocks/views/etc need to be created and/or updated.

### General landing page

- [x] Include spacing, a divider, or something else between blocks
- [x] Need a little spacing between block titles and block content
- [x] Do we NEED to display page title?
- [ ] See for reference: https://demo.mukurtu.org

### Browse by Category

- [x] Already created
- [x] Enable on install
- [x] Center, not full width
- [x] Already using the appearance at /categories, perfect
- [ ] IF possible, could rows center previews? Currently it's a strict 4-column grid, but if someone has categories that aren't a multiple of 4, centering the short rows would be nice.
   - [ ] Dynamic 3-4 column display based on count would be great, yes!

### Browse by Community

- [x] Already created
- [x] Enable on install
- [x] Center, not full width
- [x] Copy the appearance at /communities
- [ ] IF possible, could rows center previews? Currently it's a strict 4-column grid, but if someone has communities that aren't a multiple of 4, centering the short rows would be nice.
   - [ ] Dynamic 3-4 column display based on count would be great, yes!
- [x] Note this should only display communities that are open, or that a user is in. If this is addressed it should take care of it: https://github.com/MukurtuCMS/Mukurtu-CMS/issues/1005

### Featured Content

- [x] Already created
- [x] Enable on install
- [ ] See #1031 
- [ ] Center, not full width
- [x] Rename to "Featured Content"
- [ ] All Mukurtu content types should be allowed (dh item, collection, word, word list, person record)
- [x] How do we feature content? Either "promote to front page" or select-in-block (I think that is better).
- [ ] Would like to have BOTH a 3-column grid and a carousel, Mukurtu admin can select which to display.
- [ ] Copy appearance from v3 (I don't believe we have implemented this teaser view in v4 yet?)

### Browse by Map

- [x] Already created
- [x] Enable on install
- [x] Center, not full width
- [x] Teasers should use Format > rendered entity > map browse view
- [x] Add "Browse by Map" title

### Hero image: One column

- [x] Already created
- [x] Enable on install
- [x] Full width
- [x] Do not display block description above the image. Would like the option of displaying it over top of the image, like in v3. Or, that could be the page title.
- [x] Remove "image" label
- [x] Display body text BELOW the image. Maybe need a styled background (and not full width) so it stands out.

### Hero image: Two columns

- [x] Not created
[...]

PR #2201, opened 2026-09-14
Add a carousel display option to Featured Content blocks

### LINKED ISSUE #1036: Featured content carousel

michael-wynne-wsu:
Closes #1036. Also completes the outstanding carousel item from #1008, and the last open sub-issue of #1009.

Branched from `main`, not stacked. It touches the `featured_content` bundle and the theme, so it is independent of #2174 / #2189 / #2191.

## What it does

Featured Content rendered as a grid and nothing else. This adds a **Display style** field with two options.

**Grid is the default**, and it is what every existing block keeps. A site that updates and changes nothing sees no difference at all: no carousel markup, and the carousel JavaScript is never attached.

The carousel reuses **Splide**, already shipped as `mukurtu_core/splide` for the media carousel. No new library.

## Accessibility

This is most of the work, and it is why the carousel is opt-in.

| Concern | Handling |
|---|---|
| **2.2.2** auto-advance | Visible Pause/Play button, plus Splide's `pauseOnHover` / `pauseOnFocus`. The label follows the real state, including when hover or focus stopped the timer rather than a click. |
| Reduced motion | `prefers-reduced-motion` suppresses auto-advance entirely, and hides the pause button, because there is nothing left to pause. Arrows still work. |
| **2.4.3** focus order | Off-screen slides are `inert`. Splide's own `aria-hidden` covers screen readers but leaves the links tabbable, which sends focus to something invisible. |
| Duplicate content | `type: 'slide'` with rewind, never `'loop'`. |
| Name and role | `role="group"` + `aria-roledescription="carousel"`, labelled by the block's own heading, per the ARIA Authoring Practices Guide. |
| **1.4.11** / **2.5.5** | Controls are 44px targets, 8.39:1 text and 7.20:1 border. |
| Keyboard | Splide's `keyboard: 'focused'`, verified with ArrowRight. |

Two of these were bugs I introduced and then caught by measuring rather than assuming:

- **`type: 'loop'` clones slides.** The DOM had 8 slides for 4 items, so every card link appeared **twice** in the tab order and screen readers read the set twice. Rewind gives the same wrap-around with no clones.
- **`inert` was bound to `'move'`, which fires before `is-active` has shifted.** On load that left *every* slide inert, including the visible one, so the card's link could not be reached by keyboard at all until the first navigation. `'moved'` fixes it.

Focus rings use `currentcolor` rather than `--focus-color`, which is too light to meet 1.4.11 on white. That is a site-wide problem, filed separately as #2199.

## Fewer than two items falls back to the grid

Controls that cannot move anywhere are noise, and a pause button for something that cannot move is worse than none.

## Rendering

The preprocess renders each referenced entity itself, because Splide needs a `.splide__track > .splide__list > .splide__slide` structure and will not mount without it. Same approach as `mukurtu_v4_preprocess_block__inline_block__media_carousel_block()`.

That means doing two things the field formatter would otherwise have done:

- the **access check**, so an unpublished item is not a slide;
- resolving each entity through **`getTranslationFromContext()`**, so a slide shows the language of the page rather than whichever translation the reference resolves to.

## Card styling is now shared

A card looks the same in either display, so the styling is shared rather than duplicated; only the container differs. Splitting grid layout out of that block took `_featured-items.scss` from **7 stylelint problems to 3** (the scoped `no-descending-specificity` disable also covers 7 inner ones that became redundant).

Slides cap the card at `24rem`, the same cap the grid uses. Without it a single slide stretches the card's 4:3 media well to about **1280x960** on a desktop.

## Update hook

`mukurtu_update_40032()` adds the field to existing sites and backfills every block to Grid. The same change ships in `config/install` so fresh installs get it too. 40030 and 40031 are claimed by #2174 and #2191.

## Testing

10 new kernel tests across `FeaturedContentCarouselTest` and `FeaturedContentDisplayStyleUpdateTest`. All 20 profile kernel tests pass.

**Every guard was mutation-tested.** Removing the single-item fallback, removing the access check, removing the library attach, or flipping the shipped default to carousel each fails its test, and the restored suite passes.

Verified on a real site as well as by test: grid renders 4 cards with no carousel markup and no carousel JS; carousel renders 4 slides with working controls; a one-item carousel falls back to the grid. Auto-advance, pause, reduced motion, keyboard arrows and control contrast were each checked in a browser at 1280px and 480px.

## Strings

All user-facing text was approved before it was written: **Display style** / *Choose how featured items are arranged.* / **Grid**, **Carousel**; and **Previous item**, **Next item**, **Pause**, **Play**, **Featured content**, **Item 2 of 5**. Every one goes through `|t` or `Drupal.t()`.

## One open question for review

The carousel shows **one card at a time**, centred. Showing 2-3 per page would use the width better on desktop, but it changes what "Item 2 of 5" should say and needs the `inert` logic to key off visibility rather than the active index. Happy to do it if you prefer that shape.

## Pre-merge checklist:

- [x] If this PR's base branch is not `main`... (N/A, branched from `main`.)
- [x] If I added a content entity bundle... (None. A field on an existing bundle; `translatable: false` because the value is a display switch, not content.)


michael-wynne-wsu:

Link:
- https://pr2201-ahi0ua4hnku8ppat9puwmprcvl9kq141.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aa7f08c8c321beccc36d6dc

PR #2219, opened 2026-09-15
Scope the CSV user-import destination to the import wizard

michael-wynne-wsu:
Fixes #2217.

## Summary

Since #1910, `mukurtu_import_migrate_destination_info_alter()` replaced the class of core's `entity:user` migrate destination site-wide with `ProtocolAwareUserContent`, the CSV user-import destination. Its rules are written for the wizard, but every migration targeting `entity:user` ran through it, including the Mukurtu 3 to 4 user migrations:

- `mukurtu_cms_v3_users_uid1` failed with "The site superuser account (uid 1) cannot be created or updated via import", so uid 1 lost its v3 data and the id map had no row for source uid 1 for later lookups.
- `pass` was stripped from every row. Confirmed on a real migration: every migrated account had an empty password.
- `applyAccountStatus()` forced `status = 1` on every new account, so blocked v3 users came in active.

## Fix

- `ProtocolAwareUserContent` is registered as its own destination plugin, `mukurtu_import_user`, with `getEntityTypeId()` fixed to `user`.
- `MukurtuImportStrategy::toDefinition()` selects that plugin for user imports; every other entity type keeps `entity:*`.
- The `entity:user` class swap is removed from `mukurtu_import.module`, so `entity:user` is core's `EntityUser` again.

No config or update hook is involved: the wizard builds its definitions at run time.

## Tests

`ImportUserDestinationScopeTest` asserts the wizard picks `mukurtu_import_user` for users and `entity:node` for nodes, that `entity:user` resolves to core's `EntityUser` and `mukurtu_import_user` to `ProtocolAwareUserContent` bound to the user entity type, and that a plain `entity:user` migration (the shape of the v3 migrations) updates uid 1 and keeps a blocked user blocked. The existing `ImportUser*` tests (33 tests) cover the wizard path unchanged and pass.

A v3-migration-side regression test on the shared users test base will follow once #2214 is on `main`.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (None needed.)
- [x] I have run an accessibility check, and resolved any issues. (No UI changes.)
- [x] I have recompiled SCSS if required. (No SCSS changes.)
- [x] I have updated or created CI/CD tests, if required.
- [x] I have updated from `main` and resolved any merge conflicts.
- [x] If this PR's base branch is not `main`, I understand it needs to be retargeted once that base branch's own PR merges (see [docs/stacked-prs.md](../docs/stacked-prs.md)).
- [x] I have verified that all expected checks pass.
[...]

PR #2234, opened 2026-09-16
Fix two Mukurtu 3 user migration defects: stripped admin roles and dropped duplicate usernames

### LINKED ISSUE #2233: Migrating a Mukurtu 3 site strips this site's administrator roles (and can block the account)
## Summary

`mukurtu_cms_v3_users_uid1` updates this site's existing user 1 in place. The migrate destination overwrites every field the migration maps, so mapping `roles` replaced the roles of an account that already exists.

A Mukurtu 3 user 1 carrying only the plain authenticated rid (2) maps to `authenticated`, which core strips as implicit, so this site's administrator is left with **no roles at all** and loses the `administrator` role.

The same applies to `status`: a Mukurtu 3 site whose user 1 was blocked blocks this site's administrator, locking the operator out of the site they are migrating into.

This became reachable in 4.0.3. Before #2219, the whole uid 1 row was refused by the CSV import destination, so neither mapping ever applied.

Reproduced in a kernel test on 4.0.3: this site's administrator goes from `['administrator']` to `[]`. It does not depend on the two sites' user 1 sharing a name; that just makes the row look like a no-op, which is why it went unnoticed.

## Fix

Drop `roles` and `status` from the uid1 migration, alongside the name, password and email it already leaves alone: activity metadata (created, access, login, timezone, language, init, picture) is what that migration is for, and access belongs to the destination site. An update hook removes both mappings from existing sites' active config.

michael-wynne-wsu:
Fixes #2233. Also fixes a second, long-standing migration defect found while
answering "what happens when the two sites' user 1 names match?" (see the
second section); both ride into the same release.

## Summary

`mukurtu_cms_v3_users_uid1` updates this site's existing user 1 in place, and the migrate destination overwrites every field the migration maps. So mapping `roles` replaced the roles of an account that already exists: a Mukurtu 3 user 1 carrying only the plain authenticated rid (2) maps to `authenticated`, which core strips as implicit, leaving this site's administrator with **no roles at all**.

`status` has the same shape of problem: a Mukurtu 3 site whose user 1 was blocked blocks the account the operator is migrating with, locking them out of the destination site.

This became reachable in 4.0.3. Before #2219 the whole uid 1 row was refused by the CSV import destination, so neither mapping ever applied.

## Fix

Drop `roles` and `status` from the uid1 migration, alongside the name, password and email it already leaves alone. Activity metadata (created, access, login, timezone, language, init, picture) is what that migration is for; access belongs to the destination site. `mukurtu_migrate_update_40204()` removes both mappings from existing sites' active config, and skips a site that has customized either one.

`mukurtu_migrate_update_40202()` (shipped in 4.0.3) still lists the uid1 migration, which is correct: on a site updating from 4.0.2 it runs before 40204, when that mapping is still present. Its `!is_array($roles)` guard means it simply skips once the mapping is gone.

## Tests

- `MukurtuCmsV3UsersAccountFieldsTest` gains `testAdminKeepsItsOwnRoles()` and `testAdminIsNotBlockedByTheSourceSite()`, both run over a data provider covering **both shapes of source site**: user 1's name matching this site's administrator and differing from it. A matching name is what made the overwrite look like a no-op row and kept it hidden.
- `MukurtuCmsV3UsersUid1ProcessUpdateTest` covers the update hook: removes both mappings, idempotent, leaves customized or missing config alone.
- `MukurtuCmsV3UsersRolesUpdateTest` is scoped to the migration that still maps roles, with a note explaining why.
- `MukurtuCmsV3UsersRolesTest`'s uid 1 case now asserts this site's administrator roles are untouched.

Verified red then green: with the old mappings restored, all four new data-provider cases fail; with the fix they pass. Full `mukurtu_migrate` kernel suite: 57 tests, 875 assertions. phpcs and both repo lint scripts clean.

## Also: a Mukurtu 3 user whose username is already taken was dropped

Drupal keeps a unique index on the username, so a Mukurtu 3 user whose name matches an account that already exists here failed the row with:

```
SQLSTATE[23000]: Integrity constraint violation: 1062
Duplicate entry 'site_admin-en' for key 'user__name'
```

The account was never created and their Mukurtu 3 content was left with no owner to attribute it to. This is easy to hit: a Mukurtu 3 site often has its own "admin" account that isn't user 1, and this site's administrator may be named after someone who also has a Mukurtu 3 account. It is long-standing behaviour, not something 4.0.3 introduced.

Note the safe case, which is what prompted the question: the two sites' **user 1** sharing a name is fine, because the uid1 migration never maps the name and the non-admin migration excludes uid 1.

`avoid_username_collision` (mirroring the existing `avoid_uid_collision`) leaves the account already here untouched and gives the incoming user the first free numeric suffix, truncated to Drupal's 60-character limit, logging a notice naming both. A row the migration has already imported keeps the name it was given, so `--update` re-runs don't rename anyone twice. `mukurtu_migrate_update_40205()` swaps the mapping on existing sites.

`MukurtuCmsV3UsersNameCollisionTest` covers the rename, the next free number when `_1` is also taken, an `--update` re-run, and the length limit. Verified red then green: all four fail with the plain `name: name` mapping.

## Pre-merge checklist:

- [x] I have run an accessibility check, and resolved any issues. (No UI changes.)
- [x] I have recompiled SCSS if required. (No SCSS changes.)
- [x] If I added a content entity bundle, I added its `language.content_settings` and base field overrides to `mukurtu_multilingual` (see `docs/content-language-policy.md`). If I added or changed a view, I applied the content language policy in that doc. (Not applicable.)


michael-wynne-wsu:

Link:
- https://pr2234-ppe8wmzfbhj4m3fwhrbqobvntymyfiej.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aaab029f3e91ad527c9fab5

michael-wynne-wsu:

Link:
- https://pr2234-ppe8wmzfbhj4m3fwhrbqobvntymyfiej.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aaab029f3e91ad527c9fab5

Issue #1976, opened 2026-08-06
[WCAG 4.1.2, 4.1.3] Import mapping table: unlabeled Remove buttons, no focus/live-region on add/remove


michael-wynne-wsu:
**Component:** Import strategy configuration form
**File:** `modules/mukurtu_import/src/Form/MukurtuImportStrategyForm.php` (`buildMappingTable()`, lines 207-253)

### Problem

Each row in the source-column to target-field mapping table has a "Remove" button (`#value => $this->t('Remove')`) with no distinguishing accessible name - every row announces identically to a screen reader user, so there's no way to tell which row a given button removes without reading the surrounding table cells.

`addMappingCallback`/`removeMappingSubmit` trigger an AJAX replace of the whole table (`mappingTableAjaxCallback`, which just returns `$form['mapping']` verbatim) with no focus management and no `aria-live` region - a keyboard/screen-reader user gets no confirmation that "Add mapping" or "Remove" did anything.

This is the same class of defect as #1975's other two filed items (`CollectionOrganizationForm`, `MukurtuContentWarningsSettingsForm`) - three forms share this AJAX remove-row anti-pattern. `CommunityAddForm`/`ProtocolAddForm` (mukurtu_protocol) and `ManageSupportedProjectsBase` (mukurtu_local_contexts) already solve this correctly and could serve as the reference pattern.

### WCAG criteria

4.1.2 Name, Role, Value; 4.1.3 Status Messages

### Suggested fix

Give each Remove button a distinguishing accessible name (e.g. an `aria-label` referencing the row number or column name). Move focus to a sensible element after the AJAX rebuild and add an `aria-live="polite"` region reporting the change - mirrors the pattern already used in `modules/mukurtu_media/js/bulk-upload-dropzone.js`.

Filed as part of the Phase 2 (admin/authoring) accessibility inventory - see #1975.

Issue #1975, opened 2026-08-06
Accessibility Phase 2: admin/authoring inventory (complementary to #1817)


michael-wynne-wsu:
Complementary to Alex's accessibility program (#1817, `docs/accessibility/`). That program's charter marks **Phase 2 - admin/authoring UI + ATAG 2.0 - as "Not started."** This issue is the Phase 2 kickoff inventory (all 135 custom form classes individually checked, not just listed), plus coverage gaps found in Phase 1's own scope.

**Correction to the charter's status table:** Phase 2 isn't a blank slate. Reading the actual code turned up deliberate, WCAG-cited fixes already on `main` this year - `modules/mukurtu_media/js/bulk-upload-dropzone.js`, `modules/mukurtu_core/js/mukurtu-tagify-override.js`, most of `modules/mukurtu_gin_custom/js/*`, and correct labeling on the large majority of admin forms (131 of 135 checked clean). Worth knowing before scoping a baseline scan so effort isn't spent re-solving solved problems. Only a handful of concrete gaps exist, all filed below.

**Update 2026-08-06:** all 135 forms are now individually checked (the 8 that were outstanding are done - see the inventory below). `ManageCommunityMembershipForm`, flagged below as a possible duplicate of this same defect pattern, turned out to be dead code entirely - see #1973, which found it unreachable and calls a nonexistent service; it's being deleted outright rather than fixed.

**Living document:** this reflects `main` as of 2026-08-06. There are several open feature PRs in flight (multilingual stack #1960-62/1970-71, submission-forms stack #1924, map-draw #1915, lightbox #1923, etc.) that will add or change surface area - rather than wait for all of them, treat this the same way Alex's own program cycle works: scan, triage, update, repeat each cycle, folding in new pages/forms as they land.

---

## Confirmed defects (filed separately, per the program's `[WCAG X.X.X]` + `accessibility`-label convention)

- [x] #1976 - `MukurtuImportStrategyForm`: unlabeled per-row Remove buttons, no AJAX focus/live-region
- [x] #1977 - `entity_browser_view_decoration.js`: no role/aria-checked on generic browser rows (content browser, multipage-item browser)
- [x] #1978 - `CollectionOrganizationForm`: same unlabeled-Remove-button + no-live-region pattern
- [x] #1979 - `MukurtuContentWarningsSettingsForm`: same pattern

[...]

Issue #1924, opened 2026-08-01
Submission Forms Part 2: configuration, generalization to all content types, and notifications


michael-wynne-wsu:
## Context

Part 1 (#272, implemented in #1890) built the initial visitor submission workflow for Digital Heritage items: a form usable by anonymous visitors and authenticated users, built from existing DH metadata fields, with media upload, contact info collection, and admin review/publishing.

This issue tracks the follow-up work identified while building and reviewing #1890, to take on once that PR is approved and merged.

## Work items

**Field groups**
- Review the field-group configuration added during form building (collapsible sections, field assignment, ordering) for usability now that it's had real field-by-field use building the Digital Heritage form.

**Thank-you message**
- Allow admins to configure the thank-you/confirmation message shown after a successful submission.

**Multiple submission forms**
- Allow creation of more than one submission form (currently scoped to a single Digital Heritage form).

**Generalize to all content types**
- Create a default "all fields" submission form for every content type, following the same pattern built for Digital Heritage items in #1890 - so other content types (dictionary words, collections, person records, etc.) get equivalent visitor-submission support out of the box.

**Notification integration**
- Specify and implement notifications for new submissions and review outcomes.
- Depends on #1760 (notifications phase 2).
- Will likely need configurable notification rules per submission form (who gets notified on new submission, on publish/return-to-draft, etc.), similar in spirit to the per-form field configuration already built.

**Naming**
- Rename "Public Submissions" to "Submission Forms" throughout the user-facing UI - better reflects the feature once it supports multiple forms across content types, not just one public-facing form.

**Publishing workflow integration**
- Verify the submission workflow integrates correctly with both the default and enhanced (protocol-aware) publishing workflows.

## Related

- Parent issue: #272
- Implementation: #1890
- Notification dependency: #1760

PR #2245, opened 2026-09-17
Announce and label the import wizard's AJAX forms

michael-wynne-wsu:
Closes the two AJAX-announcement observations recorded in the body of #1975 but never filed, plus two defects found in the same rows while fixing them.

## The two silent AJAX rewrites

Both forms rewrite content in place and tell the user nothing happened.

**`ImportFileSummaryForm`** — choosing a different import template swaps that file's summary (`Importing Digital Heritage: Item. 12 of 15 importable fields mapped`) with no cue of any kind.

**`CustomStrategyFromFileForm`** — changing **Type** re-populates the Sub-type radios *and* every target dropdown in the mapping table, then auto-remaps them. On a wide CSV that is dozens of controls rewritten by one radio click, silently. Changing **Sub-type** does the same minus the reset.

Both now emit a core `AnnounceCommand` carrying the same information the sighted user reads.

`AnnounceCommand` rather than `AjaxRowTableAccessibilityTrait`: that trait exists to build a live region that survives its own wrapper being destroyed, and to restore focus dropped by a removed button. Neither applies here. Nothing is added or destroyed and the control that triggered the change stays in the DOM, so the trait would have been machinery with nothing to do. No new JS, no new library.

## Two more defects in the same rows

Both in `ImportFileSummaryForm`, both found while fixing the above:

**The per-row mapping select had no `#title` at all.** The visible "Import Settings" column header is not programmatically associated with the cell's input, so the control had no accessible name. With three files uploaded, a screen reader user got three identical anonymous comboboxes. It now carries a visually-hidden per-file label — which is exactly what the `weight` element two lines below it was already doing correctly.

**Every row's "Customize Settings" button has identical visible text** by design. Each now carries a per-file `aria-label`, the same remedy #1976 applied to the import mapping table's Remove buttons. The visible text is unchanged.

The summary text sitting beside the select is now associated with it via `aria-describedby`. The referenced wrapper keeps its id across the AJAX replace, so the association survives it.

## Why a 135-form audit missed these

Worth recording, because it is a reusable lesson rather than an oversight:

- **#1975 was a code read.** It marked this form clean apart from the AJAX note, and a missing `#title` on one element in a 460-line form is exactly the kind of thing reading catches unreliably.
- **The automated scans never touched it.** Only `/admin/import`, the upload step, is in `page-inventory-admin.md`. Both of these forms live behind wizard state and have never been scanned, so axe never got the chance to flag `select-name`.

That is the same class of blind spot as the submission forms, which merged with #1924 and reached neither inventory. Multi-step wizards are structurally under-covered by a scan keyed on fixed URLs.

## Criteria

- **WCAG 4.1.2 Name, Role, Value** (A) — the unlabeled select, the indistinguishable buttons
- **WCAG 4.1.3 Status Messages** (AA) — the two silent rewrites
- **WCAG 3.3.2 Labels or Instructions** (A) — the unlabeled select
- **ATAG 2.0 A.2.1 / A.3.1** — authoring interface perceivability and operability

## Verification

- New `ImportFormAccessibilityTest`: 6 cases covering the two per-row accessible names, the `aria-describedby` association and all three announcements. `OK (6 tests, 251 assertions)`.
- Full `mukurtu_import` kernel suite: **`OK (145 tests, 7368 assertions)`**, exit 0.

One seam was needed. An unreferenced private file is invisible to the access-checked entity query behind `getMetadataFiles()` outside a real request, so `buildForm()` would render zero rows and there would be nothing to assert against. `TestImportFileSummaryForm` overrides that one method; the weight filtering, row building and AJAX callbacks all run as shipped.

Note for reviewers: `phpunit.xml` sets `failOnWarning="true"`, so the test deliberately supplies the three keys (`#id`, `#title_display`, `#description_display`) that `FormBuilder` adds during element processing and a bare `buildForm()` call does not. Without them the ReplaceCommand's render emits warnings and the whole suite exits nonzero despite passing.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required: no config, schema or stored-data changes.)


michael-wynne-wsu:

Link:
- https://pr2245-8qoa9nzzizdipcszfpydhgirwhejdhek.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aac7e3f1d74270f6b653468

michael-wynne-wsu:

Link:
- https://pr2245-8qoa9nzzizdipcszfpydhgirwhejdhek.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aac7e3f1d74270f6b653468

nick-deer (approved):

PR #2246, opened 2026-09-18
Add the public submission form to the accessibility page inventory

michael-wynne-wsu:
The submission forms feature merged with #1924 and reached **neither** page inventory, so nothing has ever scanned it. #1975 flagged this as a to-do ("add the latter to `page-inventory.md` once it merges") and it was missed.

## What is covered now

Phase 1, in **both** spec layers (axe, and the reflow/zoom/focus/link-text/keyboard-trap layer):

| Page | Path |
|---|---|
| Submission form | `/submit/node/digital_heritage` |
| Submission thank-you | `/submit/node/digital_heritage/thank-you` |

Phase 2 (admin):

| Page | Path |
|---|---|
| Submission forms (list) | `/admin/config/mukurtu/submissions` |
| Submission form settings | `/admin/config/mukurtu/submissions/digital_heritage` |
| Pending submissions | `/admin/content/pending-submissions` |

## Why this needed more than a table row

**The form ships disabled.** The settings entity is `status: false` with `access_level: authenticated`, and `submit node digital_heritage content` is granted to the anonymous role only once both change. So `/submit/...` is a 403 on a stock install.

I should correct something I said while scoping this: it is **not** a live anonymous hole on stock installs. It is opt-in. It still needs coverage, because it is a long public-facing form with media upload that many sites will enable, but the framing was wrong.

A plain inventory entry would have skipped forever and looked like coverage while providing none. That is precisely the failure mode that let an unlabeled `select` survive in the import wizard for months (#2245). So the suite **enables the form, scans it, and restores the previous setting** (`tests/playwright/src/helpers/submissions.ts`).

Driven through the admin UI rather than drush, for two reasons. CI runs Playwright from a GitHub runner against a remote Tugboat preview, where there is no local site. And more fundamentally: **setting the config with drush does not grant the permission**, because the module syncs role permissions on entity save. I confirmed this directly, `drush cset status true` + `access_level anonymous` still leaves the form 403 for anonymous. The UI path is the only one that produces a reachable form.

## Two bugs found in my own setup code, both fixed

- **The describe block had to be made serial.** `beforeAll`/`afterAll` run once *per worker*, and `fullyParallel` is on. Split across two workers, one worker's teardown can disable the form while the other is still scanning it, or its setup can re-enable after the other already restored, leaving the site enabled when the run ends. I hit this: a run finished with `status: true` still set.
- **Reachability now checks for the form's title field, not just a 2xx.** Mukurtu serves a themed "Incorrect Permissions" page for denied access, so a status check alone can pass while the thing being scanned is an error page, reporting a clean result for a form nobody scanned.

## Verification

Run against a local site from a stock (disabled) starting state:

- Both scans **run rather than skip**.
- The form reports **60 focusable elements**, so it is demonstrably the real form and not an error page.
- Afterwards: `status: false`, `access_level: authenticated`, and **no stray anonymous permission** left behind. Checked directly against the config and the role.
- `npx playwright test --list` parses all 8 spec files, 107 tests.

## First findings from the new coverage

- **0 axe violations** on both pages.
- **0 findings** for reflow, text zoom, link text and keyboard trap.
- **WCAG 2.4.7 focus visible: 2 `.tagify__input` elements** with no detected outline or box-shadow on focus.

Two caveats on that last one, both worth stating rather than filing a possibly-bogus issue:

1. The check detects `outline`/`box-shadow` only. An indicator drawn as a `::before` pseudo-element reads as missing, which is exactly what produced the false positive in #2187. Tagify already carries deliberate focus work (`mukurtu-tagify-focus.js`), so this needs a human to look before it is called a defect.
2. The same run also reports 4 `.primary-nav__menu-link` elements, but those are **pre-existing and sitewide**, the home page reports the same 4. They are not a finding of this form, and I have not filed them here.

Also documents in `page-inventory-admin.md` that a wizard's presence in the table covers its **first step only**, since that is the gap that produced #2245.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required: test and documentation changes only, no config, schema or stored data.)
- [x] I have updated or created CI/CD tests, if required. (This PR is the test coverage.)


michael-wynne-wsu:

Link:
- https://pr2246-iovmb1piwbnajcbs4sxegl0sirlwlisi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aac822d1d74270f6b65e383

michael-wynne-wsu:

Link:
- https://pr2246-iovmb1piwbnajcbs4sxegl0sirlwlisi.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aac822d1d74270f6b65e383

nick-deer (approved):

Issue #2241, opened 2026-09-17
[Accessibility] Harden the CI accessibility scans so their output is representative


michael-wynne-wsu:
## Problem

All four accessibility specs already run on every PR and every push to `main` (`.github/workflows/playwright.yml` runs an unfiltered `npx playwright test`). But the output they produce is not currently trustworthy, and nothing is kept.

1. **No `A11Y_*` credentials are set in CI**, so the suite falls back to `admin`/`admin`. Every member, manage-adjacent and admin scan silently runs as UID 1. This is the exact failure mode called out in bold in `docs/accessibility/findings/2026-07-22-post-merge-verification.md`, where it produced findings that looked new but were pre-existing admin-toolbar noise. It means **CI axe output is not representative for Phase 1** and does not distinguish the member surface from the admin one.
2. **No `actions/upload-artifact` step**, so `test-results/a11y/` and `test-results/a11y-extra/` are discarded at job end. The scans are report-only by design, which makes the report the entire point.
3. **The 10-minute `globalTimeout`** may be truncating the slower admin specs.
4. **Phase 1 and Phase 2 results share `test-results/a11y/` with no phase prefix**, so they co-mingle by slug and a Phase 2 result can overwrite a Phase 1 one.

## Proposed fix

- Add repository secrets for `A11Y_USERNAME`/`A11Y_PASSWORD` (a real community/protocol member, `authenticated` only) and `A11Y_MANAGER_USERNAME`/`A11Y_MANAGER_PASSWORD` (a real Community Manager), and pass them through to the Playwright step.
- Upload `test-results/a11y/` and `test-results/a11y-extra/` as a build artifact.
- Give the accessibility specs timeout headroom.
- Prefix the Phase 2 output slugs so the two phases stop sharing a namespace.

## Why this is filed

This is step 1 of the path to a publishable conformance report in `docs/accessibility/README.md`, and it had no GitHub representation. It gates step 2 (the manual audit) only loosely, but it does gate trusting any automated evidence the ACR cites.

Scope note: workflow-only change. Keep it out of the docs PRs.

michael-wynne-wsu:
All four problems here are now resolved, and the result has been verified end to end rather than assumed.

| Problem | Resolved by | Evidence |
|---|---|---|
[...]

PR #2225, opened 2026-09-15
Fix all 18 accessibility failures on the collapsed Gin admin toolbar

michael-wynne-wsu:
## Summary

Fixes #2003 — all three findings, 18 → 0 axe violations on `/admin/content` with the toolbar collapsed.

This matters more here than in a stock Gin install: `gin-collapsed-defaults.js` makes the collapsed toolbar the **default for every new Mukurtu admin**, so every admin hit this out of the box.

### Bugs 2+3 — `button-name` (7) + `link-name` (8): CSS override

In the collapsed state Gin hides each level-1 toolbar link/button label with `display: none`, which removes the text from the accessibility tree, not just the screen — so every level-1 control loses its accessible name (WCAG 4.1.2 / 2.4.4).

One rule in `mukurtu_gin_custom.css` overrides that with the standard visually-hidden technique (absolute, 1×1, clipped). It takes no layout space, so the icon-only rail is pixel-identical, but the text stays available to assistive technology. Scoped to `html:not(.admin-toolbar-expanded)` so the expanded state is untouched.

Upstream Gin #3539420's merge request [!655](https://git.drupalcode.org/project/gin/-/merge_requests/655) exists but has sat at **"needs work" since 2025-08** and takes a heavier approach (`aria-label` on every link in the Twig templates). A CSS override is smaller and lives in the module that already exists to override Gin; the code comment links the upstream issue so it can be removed once a release restores the names.

### Bug 1 — `aria-valid-attr` (3): composer patch

Gin's `menu-region--middle.html.twig` and `menu-region--top.html.twig` write `aria-toolbar-link__labelledby="…"` on the three toolbar `<ul>`s — a typo for `aria-labelledby`, so the menus aren't actually labelled by the visually-hidden headings meant to name them.

That's a Twig attribute, not reachable from CSS. But the upstream fix — Gin [!809](https://git.drupalcode.org/project/gin/-/merge_requests/809) (#3598613) — is **RTBC**, just not committed or released as of Gin 5.0.15 (the latest stable). So it's carried as a composer patch, the same way this repo carries core/paragraphs/og fixes: `patches/gin-aria-labelledby-toolbar-menus-3598613.patch` is a verbatim copy of !809, pinned by commit SHA per the repo convention. When Gin ships it, the patch stops applying and gets removed.

## Test plan

[...]

PR #2247, opened 2026-09-18
Harden the CI accessibility scans so their output is representative

michael-wynne-wsu:
Addresses #2241. Covers the parts that do not depend on repository secrets existing yet.

**Safe to merge as-is.** With no secrets set, behaviour is unchanged from today except that results are now uploaded, the timeout has headroom, and the report says out loud that the scans ran as admin.

> [!IMPORTANT]
> **Do not set the `A11Y_*` secrets yet.** I originally wrote that setting them was all that remained. That was wrong, and following it would turn the suite red.
>
> The Tugboat preview runs `drush site-install mukurtu` on every build (`.tugboat/config.yml`), producing a fresh site whose only account is `admin`. No member or Community Manager account exists. `Login.login()` waits for a redirect away from `/user/login` and throws after 30s, so every member and manage-adjacent test would fail, each burning 30s+ — which would also eat the new 30-minute budget.
>
> The accounts have to be provisioned on the preview *before* the secrets mean anything. That is a separate piece of work, described below.

## 1. Results are uploaded

`test-results/a11y/` and `a11y-extra/` were discarded with the runner. The scans are report-only by design, so the report was the entire point of running them. Uploaded with `always()`, because a run that fails partway still produced results for the pages it reached.

## 2. The `A11Y_*` credentials are wired through both jobs

Unset, every member, manage-adjacent and admin scan silently runs as `admin`/`admin` and describes the administrator's view of the site rather than the role's. That is the failure mode flagged in bold in `findings/2026-07-22-post-merge-verification.md`, where it made a verification pass report admin-toolbar findings as if they were new member-facing ones.

**Wiring them exposed a trap worth naming.** An unset GitHub Actions secret interpolates to an **empty string**, not to nothing. So the existing

```ts
process.env.A11Y_USERNAME ?? 'admin'
```

would have yielded `''` the moment these were added to the workflow, and every login would have failed with an empty username. `??` only catches `null`/`undefined`. The twelve scattered fallbacks are now one shared helper that treats empty as unset, verified by running the suite with `A11Y_USERNAME=""`.

The suite also **annotates its report when it falls back**:

> member scan ran as the admin/admin fallback, so these results are not representative of the member role. Set the A11Y\_\* credentials to fix.

Previously the fallback was completely silent: the run went green and the results looked like member results with nothing anywhere saying otherwise.

## 3. `globalTimeout` raised from 10 to 30 minutes

Not precautionary. The run on 2026-09-18 reported:

```
89 passed (9.3m)
```

against a 600,000ms budget, so **93% of the limit**, with 18 tests skipped for want of seeded content. Those skips are nearly free, which means the real margin is thinner than 9.3/10 suggests. Either a preview with fuller content, or the credentials above being set so the member and manage-adjacent scans stop short-circuiting, pushes it over.

Past the limit Playwright stops the run partway, which surfaces as an opaque failure rather than as "this needs more time" — the same class of confusion the `TUGBOAT_MAIN_URL` readiness check was added to prevent.

## 4. Phase 2 results are written under a `phase2-` prefix

Both phases wrote into the same `test-results/a11y[-extra]` directory keyed on slug alone, so a Phase 2 page sharing a slug with a Phase 1 page would have silently overwritten it with the run still green. No collision exists today; this removes the possibility rather than relying on nobody ever choosing a colliding name.

Test titles are unchanged, so this does not churn the report; only the output filenames are prefixed.

## Verification

Run locally against a DDEV site:

- Member scans still log in after the credentials refactor.
- `A11Y_USERNAME=""` falls back to admin correctly, which the previous `??` would not have.
- `phase2-admin-overview.json` lands under the new prefix.
- The fallback annotation reaches the report (checked via the JSON reporter).
- `npx playwright test --list` still enumerates 97 tests across 8 files.
- The workflow YAML parses.

One thing I checked and ruled out while testing: a local run reported `aria-valid-attr` on `member-account`, which is the defect PR #2225 patched. That is a stale local `vendor` tree, not a regression. The patch is referenced in `composer.json` on `main` and `patches/gin-aria-labelledby-toolbar-menus-3598613.patch` is present; my DDEV site simply predates it.

## What provisioning the accounts actually involves

Worth writing down, because it is more than `drush user:create`:

- **The member account is easy.** A user with no roles beyond `authenticated`. One `drush user:create` in the Tugboat build phase, with the password read from a Tugboat environment variable.
- **The Community Manager account is not.** "Community Manager" is not a site role. The shipped site roles are `administrator`, `mukurtu_manager`, `mukurtu_roundtrip_manager` and `mukurtu_submission_reviewer`. The manage-adjacent routes resolve through `MukurtuRoleAccessCheck`, which unions site roles with **OG membership roles**. So that account needs a membership on an actual community carrying the right OG role, and communities only exist on the preview after `default-content.spec.ts` has seeded them — which runs inside Playwright, after the build.

So the values would need to live in two places: Tugboat environment variables (to create the accounts) and GitHub secrets (for Playwright to log in as them), kept in sync.

That work is deliberately **not** in this PR. This one is scoped to making the existing scans keep and report their output honestly.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required: CI and test configuration only.)
- [x] I have updated or created CI/CD tests, if required. (This PR is CI configuration.)


michael-wynne-wsu:

Link:
- https://pr2247-dvph5badgswtwjrypklcmr46npbu0g74.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aac8711a8f65b36c1a6e7df

michael-wynne-wsu:

Link:
- https://pr2247-dvph5badgswtwjrypklcmr46npbu0g74.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aac8711a8f65b36c1a6e7df

nick-deer (approved):

PR #2249, opened 2026-09-18
Provision the accounts the accessibility scans log in as

michael-wynne-wsu:
The other half of #2241. [PR #2247](https://github.com/MukurtuCMS/Mukurtu-CMS/pull/2247) makes the workflow consume the `A11Y_*` secrets; this makes the accounts those secrets name actually exist.

The two are independent and can merge in either order, but **both** are needed before the secrets should be set.

## Why this is needed

Setting the secrets without this does not improve the scans, it breaks them. The Tugboat preview runs `drush site-install mukurtu` on every build, so its only account is `admin`. `Login::login()` waits for a redirect away from `/user/login` and throws after 30s, so every member and manage-adjacent test would fail, each burning 30s+ of the run budget.

## What it creates

`scripts/tugboat/provision-a11y-accounts.php`, run from the build phase:

| Account | Site roles | Group memberships |
|---|---|---|
| Member | `authenticated` only | plain member of a community and a protocol |
| Manage-adjacent | `authenticated` only | `community_manager` on a community, `protocol_steward` on a protocol |

**It no-ops entirely unless the matching `A11Y_*` environment variables are set on the Tugboat project**, so a preview without them behaves exactly as it does today. It is also idempotent, since rebuilds and refreshes re-run it.

## Why a script rather than two `drush user:create` lines

The manage-adjacent account is not a user with a site role. **"Community Manager" is not a site role at all** — the shipped site roles are `administrator`, `mukurtu_manager`, `mukurtu_roundtrip_manager` and `mukurtu_submission_reviewer`. Those routes resolve through OG group memberships:

- `/admin/people/list` and `/admin/communities/create-user` require the OG permission `community:manage members`. Of the community OG roles, only `community_manager` carries it, and that role is created at runtime rather than shipped in `config/install`.
- `/admin/content` requires one of the roles in `mukurtu_core`'s `RouteSubscriber` list, of which `protocol_steward` is the natural fit.
- The Local Contexts projects pages accept either.

One subtlety the script has to handle: `Community::addMember()` and `Protocol::addMember()` **return early when a membership already exists**, so they never repair roles on a rebuild. The script sets roles on the membership directly instead.

## The member's memberships are not incidental

A member account with no community or protocol membership cannot reach protocol-gated item pages, so those scans skip and the run looks clean while covering less. `page-inventory.md` already asked for "a regular community/protocol member account" for exactly this reason.

Measured against a local site:

| Member account | Result |
|---|---|
| No memberships | 6 passed, 4 skipped |
| Community + protocol membership | **7 passed, 3 skipped** |
| `admin` (for comparison) | 7 reachable, same 3 skipped |

So the memberships recover `member-community-page`, and the account now covers exactly what `admin` covers while actually being a member. The remaining 3 skips are content my local site lacks, and they skip for `admin` too.

## Verification

Run against a local DDEV site:

- With no environment variables set, the script reports what it is skipping and changes nothing.
- With them set, both accounts are created; a second run creates nothing further.
- It **reuses** an existing community and protocol rather than adding more, so it does not clutter the `/communities` listing the scans discover pages from. On my site it joined the existing "Tribal community" and "Tribal members only".
- Resulting roles confirmed directly: manager has `community-community-community_manager` and `protocol-protocol-protocol_steward`; member has `authenticated` and the plain member roles.
- All four manage-adjacent pages then scan at their **real URLs** rather than redirecting to access-denied: `/admin/content`, `/admin/people/list`, `/admin/communities/create-user` and the community Local Contexts projects page. The fifth skips, as it also does for `admin`, so that skip is pre-existing and not caused by this change.

## What you do after both PRs merge

1. Set `A11Y_USERNAME`, `A11Y_PASSWORD`, `A11Y_MANAGER_USERNAME`, `A11Y_MANAGER_PASSWORD` as **Tugboat project environment variables**, so the build creates the accounts.
2. Set the same four as **GitHub Actions secrets** (Settings > Secrets and variables > Actions), so Playwright logs in as them.

They have to match, since one creates the account and the other authenticates against it. Until both are set the suite keeps falling back to `admin`, which #2247 now says out loud in the report rather than leaving silent.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required: a build-time provisioning script, no config, schema or stored data in the distribution.)
- [x] I have updated or created CI/CD tests, if required. (This is CI provisioning.)


michael-wynne-wsu:

Link:
- https://pr2249-lnnqxhqva9gsri3mujxsxprwibhzc9tb.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad49c543b28b258de4cced

nick-deer (approved):

Issue #2178, opened 2026-09-11
[WCAG 1.4.3] Horizontal image-with-description fails contrast on the Blue and gold palette


michael-wynne-wsu:
Found while auditing #1009.

## Problem

`.block--image-with-description--horizontal .block-wrapper__second` uses `background: var(--brand-secondary)`, with headings and links at `var(--brand-primary-dark)`.

| Palette | heading/link on secondary |
|---|---|
| Red and bone (shipped default) | 5.65:1 - passes |
| Blue and gold | `#107996` on `#e6ab49` = **2.45:1** - fails even the 3:1 large-text floor |
| Custom | unvalidated, author-chosen |

Palettes are one click away at `/admin/config/color-settings`, so a site can drop into the failing state without touching any content. #2174 makes the horizontal variant creatable inline, which multiplies the exposure.

## Suggested fix

Either pick a text colour per palette that clears 4.5:1 against `--brand-secondary`, or derive it. Related to #1054, which proposes automatic contrast checking in the palette chooser - that would have caught this.

Issue #1054, opened 2025-08-15
Colour palette chooser improvements


michael-wynne-wsu:
From Javier: "maybe more colors could be included in the future (or not) and a color contrast automatic testing could be implemented as well to inform editors if the colors they choose will work or not, this isn't too complicate to do actually."

michael-wynne-wsu:
Scoping the accessibility half of this ("a color contrast automatic testing could be implemented as well to inform editors if the colors they choose will work or not"), from the WCAG/ATAG program side. The "more colours" half is separate and not covered here.

## This is more urgent than the issue reads

The palette chooser is not only a choice between two curated palettes. `MukurtuDesignSettingsForm` also offers a **Custom** palette with six `#type => color` inputs, and `DesignPalette::generateCustomCss()` writes whatever the author picks straight into `:root` as custom properties:

```
brand_primary, brand_primary_dark, brand_primary_accent,
brand_secondary, brand_secondary_dark, brand_secondary_accent
```

Nothing validates them. A site admin can pick six colours and ship a site where body text fails WCAG 1.4.3 everywhere, with no indication anything is wrong. Under ATAG 2.0 Part B (B.2.2, the tool should help authors produce accessible content) that is a gap: Mukurtu asks the author to guarantee something it gives them no way to check.

## It is the systemic cause of two shipped defects

- **#2178** - the horizontal image-with-description panel used `--brand-primary-dark` text on `--brand-secondary`. That is 5.65:1 on Red and bone and **2.45:1** on Blue and gold. It shipped and had to be fixed per-palette.
- **#2175** - hero text over an author-uploaded photograph, the same shape of problem: the author is asked to guarantee a contrast the tool gives them no way to guarantee.

Both were caught by hand, months after shipping. A checker here would have caught #2178 at the point the palette was defined.

## What to check

The six editable tokens, in the combinations the theme actually renders. A quick pass over the compiled CSS shows `--brand-primary-dark` is the dominant pairing on both sides: 33 rules use it as a background and 16 as a text colour, against `--brand-secondary` (9 backgrounds), `--brand-primary` (12), the accents, and white.

The pair list should be **derived from the compiled `style.css`** rather than hand-written, the same method used for the focus-ring audit in #2199: walk the rules, collect every (text token, background token) pair that actually co-occurs, and check those. A hand-maintained list silently goes stale the first time someone adds a component.

[...]

PR #2251, opened 2026-09-18
Warn authors when a custom palette fails contrast

michael-wynne-wsu:
Implements the accessibility half of #1054, from the scoping posted there, in three commits. The "more colours" half is explicitly out of scope and parked as possible future work.

> **Note**
> This PR grew after it was opened, at your request. It started as the save-time check; it now also carries the accuracy fixes and the live readout. Nothing has been reviewed yet, so nothing was invalidated, but the diff is larger than when the link was first shared.

## Why

The palette chooser is not only a choice between two curated palettes. It also offers a **Custom** palette with six colour inputs, and `DesignPalette::generateCustomCss()` writes whatever the author picks straight into `:root`. Nothing validated them, so a site could ship text failing WCAG 1.4.3 everywhere with no indication anything was wrong. Under ATAG 2.0 B.2.2 that is a gap: the tool asked authors to guarantee a contrast it gave them no way to check.

It is also the systemic cause of two defects fixed this cycle. **#2178** put brand primary dark on brand secondary at 2.45:1 on one palette and 5.65:1 on the other. **#2175** was the same shape of problem over a photograph. Both were caught by hand, months after shipping.

## How

| Piece | Where | Does |
|---|---|---|
| `ContrastCalculator` | `mukurtu_core` | WCAG maths: sRGB linearisation, relative luminance, ratio. No CSS, no forms, so it can be tested against published values and reused. |
| `PaletteContrastAnalyzer` | `mukurtu_design` | Derives colour pairs from the **compiled stylesheet**, resolves `var()` chains against theme defaults plus the author's six values, reports pairs below threshold. |
| `validateForm()` | `MukurtuDesignSettingsForm` | Warns per failing pair, naming colours by their form labels and quoting the measured ratio. |

Pairs come from the stylesheet rather than a hand-written list because a hand-written list goes stale the first time someone adds a component, with nothing to say so.

**Warnings, not errors, and only for the custom palette.** A community or institution may have a mandated brand palette, and refusing to let them use their own colours is the wrong trade for a tool serving cultural heritage organisations. ATAG B.2.2 asks the tool to guide the author and explicitly allows them to proceed.

## Two things this had to get right

Both found by testing against the real stylesheet rather than a fixture.

**Backgrounds usually sit on an ancestor, not in the same rule.** The #2178 case sets `background-color` on `.block-wrapper__second` and the heading colour on a descendant in a separate rule. Matching only within a declaration block misses *precisely* the defect this exists to prevent. The analyzer resolves a selector's background from the longest background-setting selector it begins with. That cannot see siblings or JS-applied classes, and those pairs are skipped rather than guessed at.

**Selector lists must not be split at commas inside parentheses.** The theme uses `:is(#extra-specificity-hack, .horizontal-tabs)` widely. A plain `explode(',')` tears that into two fragments which then prefix-match the wrong rules and invent pairings that exist on no page. This was producing visibly bogus results until fixed, and is pinned by a test.

## It found two real defects on its first run

Neither is fixed here — this PR is the checker, and these want their own change and their own review. Reported so they are not lost:

1. **`.collections__content__container h2 a`** — `--brand-primary` on `--brand-secondary` at **1.96:1** on Blue and gold. Verified by hand in the compiled CSS: the container really does set `background-color: var(--brand-secondary)` and the heading link really does set `color: var(--brand-primary)`.
2. **`:is(#extra-specificity-hack, .horizontal-tabs) ul.horizontal-tabs-list li a`** — `--brand-primary-dark` on `--cr-default-bg` at **4.31:1**, just under the 4.5 threshold.

Both are now filed: **#2252** and **#2253**. Both fail only on Blue and gold, the same shape as #2178.

## Follow-up commit: resolving values the way the cascade does

Writing up #2252 exposed three related inaccuracies in what the warnings told authors. All fixed in the second commit; none change a verdict, but they change the numbers shown.

1. **`font-size` is now resolved through the same nearest-ancestor lookup as backgrounds.** It is inherited and routinely set on a different element from the colour: `.collections__content__container h2` sets the size, `h2 a` sets the colour. Reading only the colour's own block reported a 4.5:1 requirement for text that legitimately needs 3:1, which would push an author to change a colour that was already fine.
2. **Rules setting only a `font-size` were being dropped** before reaching the lookup table, because the parser skipped any block with neither a colour nor a background. That is exactly the shape of the ancestor rule establishing large text, so fixing (1) alone changed nothing at all.
3. **All seven `:root` blocks are now read, not just the first.** Colours happen to live in the first, so the reported pairs were always right, but font sizes live in later blocks and were simply unresolvable.

Also stopped treating `em` as resolvable: it is relative to the parent's computed size, which is not knowable from the stylesheet alone, and guessing produced a confident wrong threshold.

Pinned by `testLargeTextUsesTheLowerThreshold()`, which asserts the **threshold** rather than just the failure, because two separate bugs produced the wrong threshold here and the existing tests passed through both.

## Third commit: the live readout

Until now the author only learned a palette failed *after* saving it. The Custom palette fieldset now carries a readout that refreshes as each colour changes:

```
Contrast check
1 pairing needs attention.
  #fff on Brand Primary Accent: 1.78:1, needs 4.5:1
```

The callback re-runs the **same server-side analyzer** the save-time validation uses, so there is one implementation of the contrast maths and the readout cannot drift from the warnings. Nothing is reimplemented in JavaScript. Without JavaScript nothing is lost, since the save-time validation still warns.

Three decisions worth flagging, given this is itself an accessibility feature:

- **Announced with `AnnounceCommand`, not an `aria-live` region inside the readout.** The readout is replaced wholesale, so a live region inside it is destroyed and recreated by the very command meant to announce it, and never fires. Same reasoning as #2245.
- **The readout is a sibling of the colour fieldset, not a child.** `$form['colors']` is `#tree`'d and `submitForm()` saves `$values['colors']` wholesale, so a display-only element nested inside it risks being written into config as though it were a colour.
- **Colours the author does not edit are named by value, not by token.** `#fff on Brand Primary Accent` tells an author something; `--light-text-color on Brand Primary Accent` does not. This applied to the save-time warnings too, which had the same problem.

## Tests

- **`ContrastCalculatorTest`** — 16 unit tests. Pins the maths against values measured independently for the September remediation cycle: the 2.45:1 that caused #2178, the 4.78:1 that fixed it, and the 5.00/8.39:1 hero stroke figures. Also pins that `#fff` expands by repeating digits (read as `#0f0f0f` every ratio would be wrong in the safe-looking direction) and that unparseable values return NULL rather than a confident wrong number.
- **`PaletteContrastAnalyzerTest`** — 7 kernel tests against the real compiled stylesheet.
- **`PaletteContrastWarningTest`** — 7 kernel tests on the form wiring and the live readout.

That last file exists because of a bug worth recording: `colorLabel()` declared a `string` return type and handed back a `TranslatableMarkup`, fataling on the first failing palette — while **every analyzer test still passed**, because none of them went through the form. Caught by exercising the form end to end, then covered so it cannot come back.

`OK (16 tests, 49 assertions)` for the kernel suites, `OK (74 tests, 779 assertions)` for the full `mukurtu_core` unit suite.

## Pre-merge checklist:

- [x] I have checked for and if required, created update hooks. (Not required: no config, schema or stored-data changes. The check reads config, it does not alter it.)


michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:
Here is what it looks like.

Screenshot 2026-09-19 at 9 19 34 AM

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

michael-wynne-wsu:

Link:
- https://pr2251-dj33n3ofbn9c4nmsjxqcjh0jqaw9xpsh.tugboatqa.com


Dashboard:
- https://dashboard.tugboatqa.com/6aad507443b28b258deee160

